PR #3, opened 2017-03-16
Address some Py3 bytes/string issues

MatthewTurk:
I was running into problems using `v.image()` calls on Python 3.  These changes make the server use `json_decode` from `tornado.escape`, and they also manually manipulate the encoding/decoding of the b64-encoded image data.  A "hello world" of images now displays properly on Python 3 for me.


ajabri:
Thanks @MatthewTurk !

PR #5, opened 2017-03-17
Update demo.py

mehdi-shiba:
Fix spell mistake to `legend` on bar plot example code

PR #14, opened 2017-03-17
Fix JSON decode problems for Python 3

ludc:
See Issues #5

Add ```to_basestring``` into ```json_decode``` to fix string to bytes conversion problems in Python 3


ajabri:
Thanks @ludc ! Sorry for the delay, there was a misconfiguration with facebook-github-bot.

PR #23, opened 2017-03-19
Add capability of plotting mesh

wkentaro:
newplot


lvdmaaten:
Awesome!

Is there any chance you could do the same additions to the Lua-wrapper? We would like to keep the Python and Lua versions in sync (we are looking into a more sustainable solution in which the Lua-wrapper calls the Python wrapper, making development and maintenance easier).

Also, could you please update the `README.md` with this new feature?

wkentaro:
Maybe it takes a little time to write the lua code for me, because I've never tried that.

PR #33, opened 2017-03-21
Fix `plot.image` input shape

zuoxingdong:


lvdmaaten:
I think the correct size is `CxHxW`?

zuoxingdong:
@lvdmaaten In the [example](https://github.com/facebookresearch/visdom/blob/master/example/demo.py) it has

```
viz.image(
    np.random.rand(512, 256, 3),
    opts=dict(title='Random!', caption='How random.'),
)
``` 

which I guess it should be `HxWxC`.

lvdmaaten:
Hmmm so then the input format may not be consistent between the Python and the Lua wrapper (as far as I can tell, Lua uses `CxHxW`).

ajabri:
I just committed a pull request to change the python API to use CxHxW as well.

PR #36, opened 2017-03-22
Replace slashes in an environment name to underscores

### LINKED ISSUE #32: Cannot save the environment if it contains slashes
Hi, thank you for the great software first!

I am really enjoying using the software, however I found some trivial bugs that occur in corner cases.
When I try to save an environment which contains slashes, for example "model_1/dim100", it cannot save the environment and makes some unhappy side effects.
If I save it programmatically, for example `vis.save([env_name])`, the server makes error logs of `FileNotFoundError`, since the directory `$HOME/.visdom/model_1` does not exist.
And if I try to save it on the webpage, the webpage goes offline and I have to click the "offline" button or refresh the page to revive functionalities.

I think two solutions are possible:
1) making the directory (using something like `mkdir -p` or `os.makedirs`) before making the JSON file of an environment, or
2) replacing slash characters with other safe characters, like `_`.

jihunchoi:
It fixes #32.

### DIFF
--- a/py/server.py
+++ b/py/server.py
@@ -60,6 +60,17 @@ def get_path(filename):
     return os.path.join(cwd, filename)
 
 
+def extract_eid(args):
+    """Extract eid from args. If eid does not exist in args,
+    it returns 'main'."""
+
+    eid = 'main' if args.get('eid') is None else args.get('eid')
+    # Replace slashes with underscores, to avoid recognizing them
+    # as directories.
+    eid = eid.replace('/', '_')
+    return eid
+
+
 tornado_settings = {
     "autoescape": None,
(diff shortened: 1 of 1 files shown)


lvdmaaten:
Can you please make the same change in the Lua wrapper to keep the two in sync?

Character replacement in Lua can be done via `eid = eid:gsub('/', '_')`.

ajabri:
This is fine since it's server-side. Thanks @jihunchoi !

ajabri:
Looks good, could you please rebase your commit on master?

jihunchoi:
Could you explain what I should do more in detail please? There is no conflict with master.

PR #41, opened 2017-03-24
Fix default env

DmitryUlyanov:
`eid` is present in the dict, but it is `None` if default value should be used. Fixed the if statement.


ajabri:
Embarassing... :) Thanks @DmitryUlyanov !

Issue #34, opened 2017-03-21
Feature Request: Grid of multiple images


zuoxingdong:
It is often useful to plot multiple images in a grid, for comparison need. It shall be very good if Visdom adds this functionality, e.g. PyTorch has this [make_grid](https://github.com/pytorch/vision/blob/master/torchvision/utils.py) function which is nice. Hopefully it can be improved a bit and integrated into Visdom.

ajabri:
Hi @zuoxingdong ,

Yes, definitely. In torch-image, you have `image.toDisplayTensor`. We originally thought that people make such grids in their own code before calling `vis.image`, but we can add it if you think it's useful. Want to submit a pull request?

zuoxingdong:
Hi @ajabri , I am happy to work on it during next few days. Do you suggest to make it as a new API, e.g. `vis.image_grid(tensor, nrow, padding)`, 

or put it as a flag with same API , `vis.image(img, )` to make everything internally, to check 
1. if it is 3D tensor, then use original source code. 
2. If it is 4D tensor, then make a grid that mimics the functionality in torchvision `make_grid`

ajabri:
Both sound reasonable. You can also make `vis.images(tensor, ...)` that makes the grid and then calls `vis.image(img, )`.

soumith:
if you are interested, you can simply port this function over: https://github.com/pytorch/vision/blob/master/torchvision/utils.py#L6-L78

zuoxingdong:
@soumith @ajabri Thanks for the info. I've adapted this code to Numpy version, so that dependencies of torchvision is not required. The change made in PR #51

ajabri:
Fixed in #51

isalirezag:
if you guys can also add the option to it so the user can specify if it is for gray images or rgb images that would be great

JackUrb:
Hi @isalirezag - please open that as a new feature request issue so that people can see it as a higher visibility request they can work on. In the meantime you can expand your 1 channel image to 3 channels with the same content and that should plot.

PR #51, opened 2017-03-29
Fix issue #34

zuoxingdong:


ajabri:
Thanks @zuoxingdong ! Could you add the same to the Lua API? It is pretty straight forward if you use: https://github.com/torch/image/blob/master/doc/gui.md#res-imagetodisplaytensorinput-

Otherwise, I can add it after your PR...

zuoxingdong:
@ajabri I've tried to make a Lua version. Could you help if it is correct, since I've never used Lua before.

zuoxingdong:
@ajabri Does it pass the test check ?

ajabri:
Hi @zuoxingdong , sorry I've been swamped in the past few days but I will pull this in and test later today! Apologies!

Issue #72, opened 2017-04-26
`opts` vs. `options`


spro:
The true name of the options dictionary is `opts` but it is referred to as `options` in the Readme.

lvdmaaten:
Yeah, this is due to a discrepancy in the naming between the Lua and the Python wrapper. It may be hard to fix now without breaking things in people's code?

spro:
Would it be worth overloading with `options` for the sake of consistency (and documentation), and gradually deprecating the opts version? It might be a heavy change looking at the code, but better now than later.

ajabri:
IMO, we should actually change `options` to `opts` :)

spro:
Either sound good as long as they're the same

spro:
I'd be down to change that on the Lua side - it doesn't look as heavily used by Github search. Or even just the docs.

JackUrb:
moved from options to opts in 135525a224fe4b7bf859768517ef7e9344c0b570

PR #73, opened 2017-04-26
Change `opts` to `options`

spro:
This addresses #72, changing `opts` to `options` throughout. Solves potential confusion for new users by making the Python interface consistent with the Lua interface and documentation.

To avoid breaking old code, the methods that take `options` are also wrapped to accept `opts` in the kwargs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ import visdom
 import numpy as np
 vis = visdom.Visdom()
 vis.text('Hello, world!')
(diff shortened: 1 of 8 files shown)


JackUrb:
Hi @spro, thanks for flagging this discrepancy and for the PR. In the end we decided to go with the transition from options to opts rather than the other way around in 135525a224fe4b7bf859768517ef7e9344c0b570, so I'm closing this.

PR #83, opened 2017-05-15
Added scrollbars to text windows

kaniblu:
Modified css to display overflown text using scrollbars instead of just cutting the text.


ajabri:
Thanks @kaniblu !

PR #108, opened 2017-06-25
Update Python video demo

j-min:
I found Python video demo doesn't work properly on my Mac OSx since
- It doesn't automatically download the trailer video.
- '/home/%s/trailer.ogv' % getpass.getuser() is not a correct path for OSx.


example/demo.py
```diff
@@ -27,12 +29,22 @@
...
-    videofile = '/home/%s/trailer.ogv' % getpass.getuser()
+    video_url = 'http://media.w3.org/2010/05/sintel/trailer.ogv'
+    # linux
+    if _platform == "linux" or _platform == "linux2":
+        videofile = '/home/%s/trailer.ogv' % getpass.getuser()
+    # MAC OS X
+    elif _platform == "darwin":
+        videofile = '/Users/%s/trailer.ogv' % getpass.getuser()
+    # download video
+    urllib.request.urlretrieve(video_url, videofile)
```
lvdmaaten, line 40: This will work in Python 3 but break in Python 2 (there is no `urllib.request` in Python 2).

lvdmaaten:
Thanks for this! We just assumed people would manually download the video, but I agree this is cleaner.

My only concern is that I think the current code will not work on Python 2.

lvdmaaten:
Ah sorry, I missed that. Awesome, thanks!

Issue #55, opened 2017-03-29
Feature Request: Image zooming


cpdurham:
Heatmap has a nice way to zoom around and scale, was wondering about getting this into the image plot

lvdmaaten:
Good suggestion! (We actually had this in an earlier implementation of visdom.)

As a quick work-around, you can use the zoom functionality of your browser (press Command and + in Chrome / OS X) to zoom in on an image.

cpdurham:
Cool! Really liking this tool a lot!

I noticed that the zoom persists between tabs, so I have chrome zoomed out to fit the pictures and firefox for charts lol

shubhamjain0594:
@lvdmaaten is anyone working on this or is it up for grabs?

lvdmaaten:
It's up for grabs!

lvdmaaten:
Solved by https://github.com/facebookresearch/visdom/commit/ff379c05c00ca3f91e3787880b731bd07aa6c620

PR #113, opened 2017-07-08
Modified css to be able to zoom in image.

shubhamjain0594:
Solves issue #55 .

Actual Image
screen shot 2017-07-08 at 11 49 24 am

Before this commit

screen shot 2017-07-08 at 11 49 34 am

After this commit

screen shot 2017-07-08 at 11 54 02 am

### DIFF
--- a/py/static/css/style.css
+++ b/py/static/css/style.css
@@ -138,6 +138,7 @@ button {
   background-color: white;
   margin: auto auto;
   display: block;
+  width: 100%;
 }
 
 .content-plot {


ajabri:
Hi @shubhamjain0594 

Thanks for your pull request! I agree that this is an easy solution.

One hesitation I have is that you can imagine that someone might use the same pane to show different images, e.g. variable sized images as training progresses. In that case,  might it be undesirable for the image to fill the pane? It is sometimes helpful to have a sense of the image's true size.

shubhamjain0594:
Hi @ajabri 

In cases when images are of different sizes, when the panes are created on the visdom window, the size of panes will be different at that time, i.e. each pane will be having the actual size of image. Though on changing the pane size the image size will be change and then it is difficult to get original size. 

So what do yo think might be a better solution? 

Few I have in mind
- Maybe buttons for zooming in and zooming out
- Set the min-width/min-height as the original size of the image.

mlajtos:
How about [zooming on hover](https://www.youtube.com/watch?v=TJ06oyFQjuM)?

ajabri:
@mlajtos looks great to me!

lvdmaaten:
We implemented an alternative to this PR that implements wheel / trackpad zooming and resetting to original size via double-click.

PR #115, opened 2017-07-12
Zoom images on hover

mlajtos:
Solves #55 
Alternative to #113 
[Demo](https://www.youtube.com/watch?v=TJ06oyFQjuM)


lvdmaaten:
We implemented an alternative to this PR that implements wheel / trackpad zooming and resetting to original size via double-click.

PR #122, opened 2017-07-28
Fix #120 and modified README for `images` API.

### LINKED ISSUE #120: Text Scrolling
screen shot 2017-07-24 at 5 27 37 pm 1
Currently content in vis.text if larger than window, then window needs to be dragged and resized. 

But instead we can have scrollable (in both x and y) window for text. So whenever content is larger than window, scroll is much better option than resizing windows. 

If this sounds reasonable then I can send a PR for this. 

On sidelines, pip package is not updated with latest additions to this package. Filter and zoom in additions are tremendously helpful. So can you let us know how is the release cycle for visdom.

shubhamjain0594:
- Scroll ability for text
- Modified README for `images` API.


shubhamjain0594:
There was some issue with PR sorry for messing up.

ajabri:
No problem.

PR #123, opened 2017-07-28
Modified README to add  API and added scroll ability to text.

### LINKED ISSUE #120: Text Scrolling

shubhamjain0594:
- README.md modified for `images` API inclusion
- Fix #120


js/TextPane.js
```diff
@@ -23,7 +23,9 @@ class TextPane extends React.Component {
   render() {
     return (
       <Pane {...this.props} handleDownload={this.handleDownload}>
-        <div dangerouslySetInnerHTML={{__html: this.props.content}}></div>
+        <div className="text-pane"><div
```
ajabri: Can we make the className "content-text"? To match with "content-image" :)

ajabri:
Other than the above comment, looks good :)

shubhamjain0594:
@ajabri Comments have been addressed.

ajabri:
Thanks @shubhamjain0594 !

PR #127, opened 2017-08-10
Add PyTorch tensor support to `images`

t-ae:
`images` was not supporting PyTorch tensors, while [Torch version looks supporting it](https://github.com/facebookresearch/visdom/blob/master/th/init.lua#L1242).
Added it in `wrap_tensor_methods`.
And also fixed indentations.


ajabri:
Thanks @t-ae !

PR #129, opened 2017-08-10
Print URL when the server is started

t-ae:
- Print URL to access(Like TensorBoard)
- Moved [line that was never executed](https://github.com/facebookresearch/visdom/blob/9599210932760fa63647709f5d8eb361fdcba9d5/py/server.py#L434)

Now it looks 
```
$ python py/server.py 
It's Alive!
INFO:root:Application Started
(You can navigate to http://localhost:8097)
```

### DIFF
--- a/py/server.py
+++ b/py/server.py
@@ -431,7 +431,11 @@ def main():
     app = Application()
     app.listen(FLAGS.port, max_buffer_size=1024 ** 3)
     logging.info("Application Started")
-    print("(You can navigate to http://localhost:%s)" % FLAGS.port)
+    if "HOSTNAME" in os.environ:
+        hostname = os.environ["HOSTNAME"]
+    else:
+        hostname = "localhost"
+    print("(You can navigate to http://%s:%s)" % (hostname, FLAGS.port))
     ioloop.IOLoop.instance().start()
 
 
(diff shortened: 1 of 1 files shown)


lvdmaaten:
I don't think this is correct: in many use cases, the visdom server will run on a different machine than the machine on which your accessing the browser (and there will be no SSH tunnel). In such use cases, the right URL to access would be something like `http://' + os.environ['HOSTNAME'] + ':' + FLAGS.port`.

t-ae:
@lvdmaaten 
You are right.
I made some changes.
Now it becomes
```
$ python py/server.py 
It's Alive!
INFO:root:Application Started
(You can navigate to http://localhost:8097)
```
or
```
$ HOSTNAME=example.com python py/server.py 
It's Alive!
INFO:root:Application Started
(You can navigate to http://example.com:8097)
```
I'm not confident whether it is enough or not...

lvdmaaten:
I think this will work fine on OS X and Linux, but it may still give wrong results on Windows? I believe the corresponding environment variable is called `%COMPUTERNAME%` there, but I don't have a Windows machine so I cannot check it.

Merging this nevertheless, since it is better than nothing. Thanks!

PR #131, opened 2017-08-14
Fix width/height dimension order in torch code

gheaeckkseqrz:
There was a mistake in the image code, swapping the wight and height of a picture and messing up the ration when displayed.


ajabri:
Thanks @gheaeckkseqrz ! Not sure how that snuck in there.

Issue #133, opened 2017-08-16
Blank blue screen


oleggrinch:
Hi, I'm working on remote server.
I start visdom with

```
vis = visdom.Visdom(server='http://10.0.4.165',port = 8097)
vis.text('Hello, world!')
vis.image(np.ones((3, 10, 10)))
```

Server responses smth like

> INFO:tornado.access:200 POST /events (10.0.4.165) 0.69ms
> INFO:tornado.access:304 GET / (10.0.6.242) 4.36ms
> INFO:tornado.access:304 GET /favicon.png (10.0.6.242) 2.64ms

but http://10.0.4.165:8097 remains blank

Any ideas on what's going wrong? Thanks!

xqms:
Also experiencing this issue here. It seems that the CDN that visdom is using (http://unpkg.com) is down or has some problems. For example, see what is delivered instead of the react.js library: https://unpkg.com/react@15.6.1/dist/react.min.js

I think we should remove the dependency on external CDNs - what exactly do we gain?

lvdmaaten:
This seems to be working again now (it does for me)?

The reason for having the dependency like this is that we cannot distribute code by others as part of the repository. The right way to fix it would presumably be to have the install scripts download the dependencies and copy them somewhere in the install directory.

MartinKolarik:
Lots of people have been complaining about *random* 404 and 503 unpkg responses recently. You can always use jsDelivr instead, which works in a similar way but is more reliable: https://cdn.jsdelivr.net/npm/react@15.6.1/dist/react.min.js

SelvamArul:
Not solved yet. still facing the same issue.

xqms:
> The reason for having the dependency like this is that we cannot distribute code by others as part of the repository.

@lvdmaaten I'm not a lawyer, but I think all open source licenses allow redistribution as long as you also include the license file and copyright notice. I think that would be the simplest solution.

lvdmaaten:
Even if the license may allow it, Facebook does not want to re-distribute code.

SelvamArul:
@lvdmaaten Shall I work on a patch to download the required files to the py/static/css or js directories as a part of setup.py file and modify  index.html file to use the local files?

lvdmaaten:
@SelvamArul That would be great!

heurainbow:
same problem, any solution now?

lvdmaaten:
[...]

PR #135, opened 2017-08-18
Remove cd ns

SelvamArul:
This PR removes the CDN links by downloading the external files as part of post-installation hooks in setup.py. This avoids #133

### DIFF
--- a/py/server.py
+++ b/py/server.py
@@ -479,10 +479,11 @@ def download_scripts():
         if 'css' in k:
             sub_dir = 'css'
 
-        full_path = path / 'visdom' / 'static' / sub_dir / v
+        full_path = path / 'static' / sub_dir / v
         if not full_path.exists():
             req = request.Request(k, headers={'User-Agent': 'Chrome/30.0.0.0'})
             try:
+                full_path.parents[0].mkdir(exist_ok=True)
                 data = request.urlopen(req).read()
                 with open(str(full_path), 'wb') as data_file:
                     data_file.write(data)
--- a/py/server.py
+++ b/py/server.py
@@ -464,13 +464,13 @@ def download_scripts():
                 'https://unpkg.com/layout-bin-packer@1.2.2' : 'layout_bin_packer',
                 'https://cdn.rawgit.com/STRML/react-grid-layout/0.14.0/dist/react-grid-layout.min.js' : 'react-grid-layout.min.js',
                 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG': 'mathjax-MathJax.js',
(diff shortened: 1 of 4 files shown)


lvdmaaten:
Thanks, this is a great start!

I think there are still dependencies on the MathJax CDN left. In particular, [this Javascript](https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_SVG) is pulling in some other JavaScripts. This is going to be a problem as it prints the following warning in the console: `WARNING: cdn.mathjax.org has been retired. Check https://www.mathjax.org/cdn-shutting-down/ for migration tips.`

Also, the MathJax appears to be expecting some fonts in `static/fonts/`?

lvdmaaten:
@ajabri just pointed out another potential issue: `setup.py` is not run with `pip install` if the package is distributed as a wheel. Which means we probably have to run it at server launch time (which, admittedly, is not a very nice solution).

SelvamArul:
@lvdmaaten I moved the downloading of scripts to server.py and executed before launching the server. This means we don't have access rights to copy the files to visdom installation directory. So I am using `FLAGS.env_path` defined in `server.py` as the download target. To enabling the web server to serve files in this new directory I created a new `StaticFileHandler`. I am not sure this is the best way of doing it.

Now, I have issues with `mathjax-MathJax.js` downloading `extensions/MathMenu.js` which intern downloads more files from the CDN. Apparently, we need to download the whole `MathJax` package. So I am thinking of downloading the whole `MathJax` package from `https://github.com/mathjax/MathJax/archive/2.7.1.zip`. Do you have a suggestion regarding this?

lvdmaaten:
@SelvamArul Thanks, this looks great!

I'm going to need some time to test this internally (right now, it will break our internal visdom setup but I know how to fix it). Why is the `StaticFileHandler` needed? Can't you change the path from which static files are served in https://github.com/facebookresearch/visdom/blob/master/py/server.py#L82?

lvdmaaten:
I just committed an adapted version of your PR. It seems to be working correctly for me. Let me know if you have any problems!

SelvamArul:
@lvdmaaten The installation works. Just a small note regarding MathJax.  I still have this error:
`Uncaught SyntaxError: Unexpected number  MathMenu.js:2`
`Uncaught SyntaxError: Unexpected number  MathZoom.js:2` 

`mathjax-MathJax.js` looks for these files (`"[MathJax]/extensions/MathMenu.js"`). Should we download the whole MathJax package?

lvdmaaten:
@SelvamArul I haven't seen that issue myself (yet), so it would be great if you could send a PR adding downloads of the files we're still missing. I'm okay with downloading the full package. It is probably not exceptionally large, and it would only happens once at server launch anyway.

PR #164, opened 2017-10-08
API for checking server connection

jkarimi91:


lvdmaaten:
Hmmm on second thought, this does feel very hacky (in particular, the `sleep(1)` part). Since we have added the `win_exists()` function, can't we use that? If it returns `None`, I believe there is no connection; see https://github.com/facebookresearch/visdom/blob/master/py/__init__.py#L309

What do you think?

Btw... the `check_connection()` function that is added here still needs to be exposed in Lua, akin to https://github.com/facebookresearch/visdom/blob/master/th/init.lua#L219

JackUrb:
I second @lvdmaaten's idea - using `win_exists` is definitely a cleaner implementation, especially on the python end. We'd be happy to bring that version in. As far as the Lua version goes, if you're having trouble getting torch running I can add the wrapper before I merge the code in.

jkarimi91:
@JackUrb and @lvdmaaten I completely agree, using `win_exists` is the way to go; I will update this PR in a moment. @JackUrb that be great since I am having trouble getting torch running.

PR #166, opened 2017-10-10
Use os.path for saving in env_path

varunagrawal:
We use the `os.path` module to join paths rather than string formatting. This alleviates the need to end the `env_path` with a /.


JackUrb:
This is a nice little fix - I needed to check to make sure it wouldn't break any existing workflows but it seems good to me.

PR #167, opened 2017-10-10
logging_level string argument

varunagrawal:
The `logging_level` command line parameter now takes string directly and checks amongst the available choices in the python `logging` library.


JackUrb:
Hi @varunagrawal, while I definitely agree that having logging level take a string, I know that some already utilize or otherwise prefer using the ints. It's possible to have a happy middle ground that relies on custom argparser validation (such as in [this](https://stackoverflow.com/questions/14117415/in-python-using-argparse-allow-only-positive-integers)) where you take in a string, see if it converts to an int and return it if so, otherwise check it against the list of choices and return the value for that choice on a match, and raise an error for everything else.

varunagrawal:
@JackUrb thank you! Let me update the PR.

varunagrawal:
Updated!

Issue #20, opened 2017-03-17
Feature Request: update text pane


elanmart:
Hi, 
sometimes it's useful to send some textual data from the model in regular intervals. As far as I understand, right now we can only replace text in the window with something new, but we cannot append.

I can easily monkey-patch it by adding the following to `UpdateHandler` and modifying `css` to display newlines, but I don't know how to do this properly. 
```python 
def update(self, p, args):
    if isstr(p['content']):
        p['content'] += "\r\n" + args['data'][0]['content']
        return p
```

lvdmaaten:
The most logical way would be to add an `update` option to the Python and Lua implements of `text()`. If `update` is set to `True`, the `UpdateHandler` should be called instead. Here's an example of how to send a request to the `UpdateHandler`: https://github.com/facebookresearch/visdom/blob/master/py/__init__.py#L306-L315

In the `UpdateHandler`, you would have to check that the `type` of the request is `'text'` before executing the string concatenation code above.

lvdmaaten:
Fixed by https://github.com/facebookresearch/visdom/pull/168

PR #168, opened 2017-10-10
Fix for #20

varunagrawal:
This PR adds the `update` option to the `text` function to allow the text in the window to be updated with new text appended rather than replaced.

I'm not sure if `update` is the right arg name here. Should it be something like `append`?


JackUrb:
Hi @varunagrawal thanks this looks great - I'll pull it in soon

PR #193, opened 2017-11-03
Updatetrace fix

### LINKED ISSUE #175: Error while running Demo file
While running [scatter](https://github.com/facebookresearch/visdom/blob/master/example/demo.py#L92) in demo.py, raises an error:

```bash
Exception in user code:
------------------------------------------------------------
File "/home/users/shubham/anaconda2/envs/cxr_dl/lib/python3.6/site-packages/visdom-0.1.6-py3.6.egg/visdom/__init__.py", line 262, in _send
   data=json.dumps(msg),
File "/home/users/shubham/anaconda2/envs/cxr_dl/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
File "/home/users/shubham/anaconda2/envs/cxr_dl/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
File "/home/users/shubham/anaconda2/envs/cxr_dl/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
TypeError: keys must be a string
```

This is because in opts['markercolor'] has keys as class numbers not strings. Which is being set [here](https://github.com/facebookresearch/visdom/blob/master/py/__init__.py#L612).

One way to solve is by deleting markercolor from opts if it exists since we are already passing it while preparing _data. If this sounds good, I will submit a PR for this.

shubhamjain0594:
Fixes #175


py/__init__.py
```diff
@@ -563,7 +563,8 @@ def updateTrace(self, X, Y, win, env=None, name=None,
             'opts': opts,
```
JackUrb, line 569: Can you add a print up here in updateTrace that notes using updateTrace will be deprecated in the future and that users should switch to using `func(..., update='append', name=<trace_name>)`?

py/server.py
```diff
@@ -365,7 +367,7 @@ def update(p, args):
         idxs = list(range(len(pdata)))
 
         if name is not None:
-            assert len(new_data['x']) == 1
+            assert len(new_data) == 1
```
JackUrb: Does this change break the existing ability to use updateTrace? If so, can you use a case to maintain the old functionality of updateTrace and stick a `#TODO remove when updateTrace is deprecated` there (can be done by just copying the old function to a new def, putting a case here, and calling the old one for the old case)?

py/__init__.py
shubhamjain0594: Addressed.

shubhamjain0594:
@JackUrb thanks for reviewing. The issues have been addressed.

JackUrb:
Thanks for the contribution!

PR #212, opened 2017-11-29
Use pathname when crafting addresses in the js

albanD:
Fix addresses crafted in the js when visdom is not at the "base address".
This allows to use visdom behind a proxy at addresses like "ip_address:port/foo".

I just rebuilt the static js for python, let me know if anything else should be changed.


js/main.js
```diff
@@ -142,7 +142,7 @@ class App extends React.Component {
     }
 
     var url = window.location;
-    var socket = new WebSocket('ws://' + url.host + '/socket');
+    var socket = new WebSocket('ws://' + url.host + url.pathname + 'socket');
```
JackUrb: window.location.pathname doesn't always end in a `/`, will that be a problem here?

albanD: Hi,
I am not familiar with js, from my tests in this code, `url.pathname` contains either a single `/` or `/some_things/`.
I am not sure where to look online to find exactly what this should contain, are there cases I'm missing? Would you have a pointer to the documentation of exactly what this contains so that I can make a more resilient code?

JackUrb: Ah that would make some sense based on our current codebase, however url.pathname doesn't always append that final slash (it might return `/pathname`) so this might break in the future or if somebody is trying to make their own changes to visdom. Rather than have them need to fix it then I'd rather fix it in advance.

Can you instead declare a variable `var pathname = url.pathname` and then use a conditional `if pathname.slice(-1) != '/' { pathname = pathname + '/' }`? This could also be a separate function `correctPathname` so you don't need to put it in both locations

Sorry for being somewhat picky on an otherwise great change to have!

albanD: No problem !

One last question: can `windown.location.pathname` be an empty string? Or will it at least be `'/'` ?
EDIT: this is not relevant as `.slice` will work on an empty string.

Sorry for all the questions, but I can't find the answer on google (I never expected to say that one day ! )

JackUrb:
Looking good - thanks for this!

PR #236, opened 2017-12-28
Delete environments

JackUrb:
Exposes both a frontend and backend interface for deleting environments. 

The frontend one is used in a followup that also supports multiple layout views.
Tested using the upcoming frontend implementation. 

Also exposes a method to send Environment updates to the frontend, meaning new environments are accessible from the frontend without a refresh.


lvdmaaten:
Nice!

PR #237, opened 2017-12-28
Saving multiple views per environment

JackUrb:
A bunch happens here:
- A feature is added that saves multiple views for a given environment. It's not perfect just yet but it works and adds a bunch of flexibility. Issues: 
  - As the saved views aren't pushed to the server, they're lost on refresh. 
  - Something is still strange about the packing function, so sometimes after switching views the pack button needs to be pressed 2-3 times before it settles. They do settle into the original saved position as long as that position was already packed when it was saved)
- Top navbar is split into dialogs that are rendered on button press
- Delete env functionality is added into the frontend

Screens:
**New nav bar**
screen shot 2017-12-28 at 5 29 11 pm
**Environment Manager**
screen shot 2017-12-28 at 5 28 59 pm
**View manager**
screen shot 2017-12-28 at 5 50 11 pm

TODO: 
- Once this is approved and finalized, add new photos and gifs of the functionality to the docs.
- Add little X to the dialog as it's unclear that unfocusing makes them go away
- Bump the version number with this one

Open Q's:
- Should we label dropdowns for env and layout?
- What is wrong with packing? 🤔


JackUrb:
Some updated views now. Dumping images here to use for the README
*Toolbar is cleaner:*
screen shot 2018-01-05 at 11 28 40 am
*Filter has a merged input+button* (Still want to just make this field auto-update w/o button)
screen shot 2018-01-05 at 10 32 32 am
*View options are grouped, Pack is now a view option*
screen shot 2018-01-05 at 10 31 29 am
*Environment options are grouped, clear was returned to the bar*
screen shot 2018-01-05 at 10 31 21 am
*Tooltips were added for icon based buttons*
screen shot 2018-01-05 at 11 28 48 am
*Dropdowns have been updated to match toolbar format, selected indicators still exist*
screen shot 2018-01-05 at 10 30 57 am
*Manage windows are now essentially just save/delete for both envs and views*
screen shot 2018-01-05 at 10 32 05 am

JackUrb:
I want to add a video demo of the view changing process too:
visdomvid2

PR #243, opened 2018-01-06
Fix string formating in assert

gheaeckkseqrz:
Was giving the following error when passing bad arguments : 
```
  File "/Users/wilmot_p/PROG/PONG/.env/lib/python2.7/site-packages/visdom/__init__.py", line 130, in _markerColorCheck
    (X.shape[0], X.shape[1], L, L, 'x'.join(mc.shape))
TypeError: sequence item 0: expected string or Unicode, int found
```


lvdmaaten:
Great, thanks for fixing this!

PR #246, opened 2018-01-08
Use PNG instead of JPG for temp image saving

ajabri:
Use PNG for temp image saving format to mitigate artifacts from JPEG. Since most people using this tool care more about image accuracy than compression, IMO this should be the default.

 test JPEG (quality: 100)
test PNG

### DIFF
--- a/py/__init__.py
+++ b/py/__init__.py
@@ -403,7 +403,7 @@ class Visdom(object):
 
         data = [{
             'content': {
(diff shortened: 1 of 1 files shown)


ajabri:
Done. Weirdly enough it didn't seem to make a difference when I tested but better done correctly!

lvdmaaten:
Thanks!

PR #249, opened 2018-01-12
Adds a way to save views to the server to load later

JackUrb:
Previously to this commit, views were lost on refresh. This allows for views to be synced with the server so that they can be used later for any number of reasons. Also moves the views feature out of *beta* now that it has at least the minimum planned functionality implemented.


README.md
```diff
@@ -91,18 +91,16 @@ You can use the `filter` to dynamically sift through windows present in an env -
...
 
-### [BETA] Views
+### Views
 <p align="center"><img align="center" src="https://user-images.githubusercontent.com/1276867/34618173-e2546f40-f20b-11e7-9969-16267891fb53.png" width="300" /></p>
 
 It is possible to manage the views simply by dragging the tops of windows around, however additional features exist to keep views organized and save common views. View management can be useful for saving and switching between multiple common organizations of your windows.
 
 #### Saving/Deleting Views
-Using the folder icon, a dialog window opens where views can be forked in the same way that envs can be. Saving a view will retain the position and sizes of all of the panes in a given environment.
+Using the folder icon, a dialog window opens where views can be forked in the same way that envs can be. Saving a view will retain the position and sizes of all of the panes in a given environment. Views are saved in `view/layouts.json` in the visdom filepath.
```
lvdmaaten: Perhaps include `$HOME/.visdom/` to clarify which path is used here?

py/server.py
```diff
@@ -145,21 +147,50 @@ def __init__(self):
...
+        if target_subs is None:
+            target_subs = self.subs.values()
+        for sub in target_subs:
+            sub.write_message(json.dumps(
+                {'command': 'layout_update', 'data': self.layouts}
+            ))
+
+    def save_layouts(self):
+        layout_filepath = os.path.join(FLAGS.env_path, 'view', LAYOUT_FILE)
+        fn = open(layout_filepath, 'w')
```
lvdmaaten: Check that `fn` is actually open? (You could use the `with open(layout_filepath, 'w') as fn:` syntax here.)

py/server.py
```diff
@@ -145,21 +147,50 @@ def __init__(self):
...
+        fn.write(self.layouts)
+        fn.close()
+
+    def load_layouts(self):
+        layout_filepath = os.path.join(FLAGS.env_path, 'view', LAYOUT_FILE)
+        ensure_dir_exists(layout_filepath)
+        if os.path.isfile(layout_filepath):
+            fn = open(layout_filepath, 'r')
```
lvdmaaten: Idem.

py/server.py
```diff
@@ -197,6 +230,12 @@ def on_message(self, message):
                 os.remove(p)
                 broadcast_envs(self)
 
+        if cmd == 'save_layouts':
+            if 'data' in msg:
+                self.layouts = msg.get('data')
+                print(self.layouts)
```
lvdmaaten: This print can be removed. (I presume you were using it for debugging?)

JackUrb: Agh thought I got all of the debugs - yeah it made sure the layout was coming through as expected.

PR #250, opened 2018-01-15
Logging argument

varunagrawal:
Command line option for logger level now accepts more human friendly strings instead of obscure integers.


varunagrawal:
@JackUrb ah I see. The argument option for the logging level has already been updated. Terribly sorry about that. Will reset and only include the README changes.

JackUrb:
No worries! I'm sure if we had it in the docs it would've been clearer. Thanks for the contributions!

PR #253, opened 2018-01-17
convert loaded svg from bytes to string

### LINKED ISSUE #252: Error when loading svg image TypeError: cannot use a string pattern on a bytes-like object
This happens when i try loading an svg image using Visdom.svg().
I think i fixed it in my case replacing the following line from [https://github.com/facebookresearch/visdom/blob/master/py/__init__.py#L370](url):
`svgstr = loadfile(svgfile)`
with
`svgstr = str(loadfile(svgfile))`
but i don't know if it works in every case.

jj0mst:
Fixes #252.

### DIFF
--- a/py/__init__.py
+++ b/py/__init__.py
@@ -367,7 +367,7 @@ class Visdom(object):
         _assert_opts(opts)
 
         if svgfile is not None:
-            svgstr = loadfile(svgfile)
+            svgstr = str(loadfile(svgfile))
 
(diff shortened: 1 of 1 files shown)


JackUrb:
Thanks for the fix!

JackUrb:
reimplemented and closed in https://github.com/facebookresearch/visdom/commit/641c64fe6b4e63d31d975b7b6502ba6dd826d67b

PR #254, opened 2018-01-22
Fixed chart axis display options ([xy]tickstep, [xy]tick.

TOertel:
Removed dead "[xy]tickfont" reference that wasn't being used.
Fixed support for [xy]tickstep (assuming README description is intended functionality).
Fixed '[xy]tick' parameter to actually work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,6 +479,7 @@ The following `opts` are generic in the sense that they are the same for all vis
 - `opts.xtickvals`   : locations of ticks on x-axis (`table` of `number`s)
 - `opts.xticklabels` : ticks labels on x-axis (`table` of `string`s)
 - `opts.xtickstep`   : distances between ticks on x-axis (`number`)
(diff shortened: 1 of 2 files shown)


JackUrb:
Thanks for the fixes to the buggy tickstep and tick params! 

Re: tickfont - ack that was supposed to be something that was considered but I suppose I dropped the line. Can you add a line for `'tickfont': opts.get(x + 'tickfont')` into the returned dict instead of removing the option?

TOertel:
I looked at the font stuff prior to removing it from the list, and it required it's own structured data, which was more complicated than I wanted to do (and test).  Font isn't currently supported anywhere else that I can see...

But apparently a naive implementation works, so I've added it.  I wasn't sure how to test README.md, but it's updated as well.

JackUrb:
Thanks a lot for your help here! Yeah we're planning on adding more of the font support moving forward (hopefully in a way that lets visdom be a more direct interface to the plotly library) but tickfont was a special request.

PR #260, opened 2018-02-06
Back socket

JackUrb:
Whew this is a doozy - This establishes the backchannel via websockets in the first commit, and then shows a working implementation in the form of the keyboard functionality in the second commit.

This includes a demo in the demo file of using these callbacks to set up a simple notepad-like pane. Seems to work:
screen shot 2018-02-06 at 3 26 19 pm

Callback implementation format is described in the README, essentially allows registering one callback per pane, and then that function can case on the different event types if desired.

If desire appears, we'll also move to support handling events by event type rather than just by pane, but IDK if there's a strong enough case for that.


README.md
```diff
@@ -45,7 +45,25 @@ The UI begins as a blank slate -- you can populate it with plots, images, and te
 
 > **Tip**: You can use the zoom of your browser to adjust the scale of the UI.
 
+##### Callbacks
 
+The python Visdom implementation supports callbacks on a pane. The demo shows an example of this in the form of an editable text pad. The functionality of these callbacks allows the Visdom object to recieve and react to events that happen in the frontend.
```
lvdmaaten: `receive`

README.md
```diff
@@ -45,7 +45,25 @@ The UI begins as a blank slate -- you can populate it with plots, images, and te
...
+ - `target`: the pane id the event is called on
+
+Additional parameters are defined below.
+
+Right now two callback events are supported:
+
+1. `close` - Triggers when a window is closed. Returns a dict with only the aforementioned fields.
+2. `keyPress` - Triggers when a key is pressed. Contains additional parameters:
+    - `key` - A string representation of the key pressed (applying state modifiers such as SHIFT)
+    - `keyCode` - The javascript event keycode for the pressed key (no modifiers)
```
lvdmaaten: To be consistent with the other input arguments in visdom, I think it is better to not use camel-case argument names here.

example/demo.py
```diff
@@ -27,26 +27,26 @@
...
+        elif event['key'] == 'Backspace':
+            curr_txt = curr_txt[:-1]
+        elif event['key'] == 'Delete':
+            curr_txt = txt
+        elif len(event['key']) == 1:
+            curr_txt += event['key']
+        viz.text(curr_txt, win=callback_text_window)
+
+
+viz.event_handlers[callback_text_window] = type_callback
```
lvdmaaten: An alternative API would be to have a function `register_event_handler(handler, win)` that updates the `event_handlers` dict. The advantage of this would be that the function could check that the specified `win` exists (and raise an exception if it doesn't), that `iscallable(callback)`, and that it could compose callbacks of multiple callbacks are added to the same window.

py/__init__.py
```diff
@@ -228,20 +232,70 @@ def __init__(
...
+        def on_message(ws, message):
+            message = json.loads(message)
+            if 'command' in message:
+                # Handle server commands
+                if message['command'] == 'alive':
+                    if 'data' in message and message['data'] == 'vis_alive':
+                        logger.info('Visdom successfully connected to server')
+                    else:
+                        logger.warn('Visdom server failed handshake, may not '
+                                    'be properly connected')
```
lvdmaaten: I think a similar change may be needed in `check_connection()`? If `self.send == True`, I think we should check that the socket is functioning in that function.

py/server.py
```diff
@@ -208,8 +249,14 @@ def on_message(self, message):
...
-                self.state[msg['eid']]['jsons'].pop(msg['data'], None)
-
+                p_data = self.state[msg['eid']]['jsons'].pop(msg['data'], None)
+                event = {
+                    'eventType': 'close',
+                    'target': msg['data'],
+                    'eid': msg['eid'],
+                    'paneData': p_data,
+                }
+                send_to_sources(self, event)
```
lvdmaaten: Nice! :)

README.md
JackUrb: Makes sense - I wasn't really sure where to draw the line between our JS components (all camel case) and the python stuff

example/demo.py
JackUrb: Yeah having it be an API call would be rather nice. I'm keeping away from function composition for now though b/c I don't think there's a good clean way to deal with decomposition. Could be dealt with later.

JackUrb: A user may want to create a callback for a window that doesn't yet exist though, I feel this behavior is not degenerate as users are allowed to declare window names on creation.

lvdmaaten: Wouldn't callback composition simply amount to running each of the callbacks sequentially in the order that they were added? Under the hood, `event_handlers[callback_text_window]` would maintain a list of callbacks rather than a single callback. `register_event_handler()` would simply append to the correct list (creating a new empty list if it doesn't exist yet).

Re callback for a window that doesn't yet exist though: I see, makes sense. In that case, the registering function would not have to check for `win` existence.

JackUrb: Oh yeah that's exactly how it would work, my difficulty would be how to expose an API for _removing_ a callback that has already been added. What do you think would be a good approach on that front?

README.md
lvdmaaten: I see. Most users will likely be accessing these variables from Python, so for them, it makes sense to use lower-case. You could "translate" the variable names in `check_origin()` if you want to keep camel-case in the JS code (admittedly, that is not very elegant though).

example/demo.py
lvdmaaten: Hmmm yeah that is tricky. Perhaps add a function `remove_callbacks(win)` that empties the table with callbacks for the specified `win` suffices? If users want more advanced behaviors, they can always implement the desired composition themselves (in the same way that they need to manually manipulate the content of `event_handlers` now)..?

README.md
JackUrb: JavaScript by nature isn't elegant, I'll leave the mismatched naming there 😆

PR #272, opened 2018-02-21
Expose get api for getting data for existing windows.

### LINKED ISSUE #266: Feature Request : API to pull scalar data
Tensorboard allows to curl down the data from specific plots (scalars particularly).
Not sure about the plotly blockers, but would be nice feature to gather data from 100s of envs.

JackUrb:
Creates a DataHandler to expose an endpoint that allows a user to get window data that exists on the server from the `visdom` class. 

Adds method to the visdom class that calls the data handler.

Updates docs with the new function

Quickly cleans up the server file b/c it was throwing linting errors

Testing:
screen shot 2018-02-21 at 11 30 03 am

Closes #266 by implementing the desired feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,7 @@ vis._send({'data': [trace], 'layout': layout, 'win': 'mywin'})
 ### Others
 - [`vis.close`](#visclose)    : close a window by id
 - [`vis.win_exists`](#viswin_exists) : check if a window already exists by id
+- [`vis.get_window_data`](#visget_window_data): get current data for a window
 - [`vis.check_connection`](#vischeck_connection): check if the server is connected
 
 ## Details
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -542,6 +543,15 @@ This function returns a bool indicating whether or not a window `win` exists on
...
 - `env`: Environment to search for the window in. Default is `None`.
 
+#### vis.get_window_data
+This function returns the window data for the given window. Returns data for all windows in an env if win is None.
+
+Arguments:
+- `env`: Environment to search for the window in.
+- `win`: Window to return data for. Set to `None` to retrieve all the windows in an environment.
+
+*Note:* This function is python-only and doesn't exist in the Lua version.
```
lvdmaaten: Should it not be easy to expose a wrapper of `get_window_data()` in the Lua client, also, or am I missing something?

py/server.py
```diff
@@ -730,6 +738,31 @@ def post(self):
...
 
+class DataHandler(BaseHandler):
+    def initialize(self, app):
+        self.state = app.state
+
+    @staticmethod
+    def wrap_func(handler, args):
+        eid = extract_eid(args)
+
+        if 'win' in args and args['win'] is not None:
```
lvdmaaten, line 749: `and args['win'] in handler.state[eid]['jsons']`?

README.md
JackUrb: It wouldn't be, though I wasn't quite sure what the use case would be in Lua. I could add this before I push if you think it'd be useful

py/server.py
JackUrb: I should put this as an assert - errors are passed from the server back to the user.

lvdmaaten: Even better!

PR #275, opened 2018-02-22
Better plotly feature parity

### LINKED ISSUE #197: Feature request: plot the legends within the graph
Is it possible to add an option to plot the legends within the graph itself, or to change the location of the legends? Thanks in advance!

### LINKED ISSUE #200: Feature request: is it possible to disable the background grid when plotting?
[sample.pdf](https://github.com/facebookresearch/visdom/files/1464398/sample.pdf)

I am wondering if it is possible to remove the grid from the final plot. Is there an option? Thanks!
There is an example cropped PDF attached.

### LINKED ISSUE #201: Feature request: error bars (basic and continuous)
Hello! I am interested in using error bars in my Visdom plots, ideally the continuous error bars (https://plot.ly/javascript/continuous-error-bars/), but just as well the basic error bars (https://plot.ly/javascript/error-bars/).

I tried adding basic error bars simply by adding,

```python
'error_y': {
    'type': 'data',
    'array': [10.0],
    'visible': True
},
```

(for example) to `scatter()` here: https://github.com/facebookresearch/visdom/blob/master/py/__init__.py#L648.

This works for the first point, but not for appending additional points (i.e. for the first point I get an error bar, but additional points get added to the plot with no error bars).

I was wondering if I might get some direction about what changes need to be made to get this functionality in, since Plotly supports it but it seems that Visdom does not. Do changes need to be made to server.py?

Thanks!

JackUrb:
Better supports general plotly features by adding opts to specify additional layout opts that we don't manage and additional trace options that we don't already provide.

On the downside it means there are now 2 ways to accomplish most of the things that we directly support, but those that use the plotly way are already versed enough in the plotly formatting to not necessarily need to use our implementation anyways, and our implementation is still better for people unfamiliar with plotly.

Closes #201, as error bars are implemented using trace options in plotly, so they are now possible for us as well.
Closes #197, as this is a layout option that we can now pass along.
Closes #200 for the same reason as above.

Prevents future "plotly does it but visdom doesn't" feature requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,8 +349,8 @@ The following `opts` are supported:
 - `opts.markercolor` : color per marker. (`torch.*Tensor`; default = `nil`)
 - `opts.legend`      : `table` containing legend names
 - `opts.textlabels`  : text label for each point (`list`: default = `None`)
-- `opts.layoutopts`  : dict of any additional options that plotly accepts for a layout
-- `opts.traceopts`   : dict mapping trace names or indices to dicts of additional options that plotly accepts for a trace.
+- `opts.layoutopts`  : dict of any additional options that the graph backend accepts for a layout. For example `layoutopts = {'plotly': {'legend': {'x':0, 'y':0}}}`.
+- `opts.traceopts`   : dict mapping trace names or indices to dicts of additional options that the graph backend accepts. For example `traceopts = {'plotly': {'myTrace': {'mode': 'markers'}}}`.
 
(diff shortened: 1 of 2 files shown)


JackUrb:
Definitely agreed on the override order though I have a decent argument - the layoutopts is already invoking a hard override of our formatting and functionality, so it would take precedence as it is the closest wrapper to the direct plotly implementation.

PR #283, opened 2018-02-26
Feature: Compare Environments

pareshmg:
* Added react component rc-tree-select which allows for searching and filtering through environments in dropdown
* Multiple environments can be selected to result in a comparison in plots with the same title (a legend is automatically created indexing the selected environments) [Note I have whited out some of the environment names]
compare

* The rc-tree-select is populated hierarchically by splitting according to the first underscore. This allows for some organization of environments and projects. [Note: I have whited out some of the environment names]
treeselect


JackUrb:
Hi @pareshmg, thanks for the contribution! I look forward to getting into it. First though, can you update your JS indentation settings to match our standard in this repo (2 spaces per indent)?

pareshmg:
@JackUrb : indent updated to 2.

Also, I wanted your input on something

Goal: On browser refresh, update the currently selected environments for compare.

Issue: React doesn't seem to execute callbacks on setState.

You can see some console.log debug messages in the setState part of socket.onClose. I'm not sure if this is a react bug (on googling, it seems like it was a while back), but the callback never executed for me. Because of this, I have currently disabled that feature but was hoping that someone with more experience in this would be able to help me out.

If you want, I can take out those debug console.log messages, but they never get called for me.

Also, thanks for Visdom.  I love it!

webpack.config.js
```diff
@@ -1,26 +1,30 @@
 module.exports = {
-  entry: [
```
JackUrb: Nit - spacing. This should also remain in the style we had

py/static/index.html
```diff
@@ -25,12 +25,15 @@
     <link rel="stylesheet" href={{ static_url("css/react-grid-layout-styles.css") }}>
 
     <!-- Other deps -->
-    <script src={{ static_url("js/react-react.min.js") }}></script>
-    <script src={{ static_url("js/react-dom.min.js") }}></script>
-    <script src={{ static_url("fonts/classnames") }}></script>
+    <script src={{ static_url("js/react.development.js") }}></script>
```
JackUrb: We shouldn't be committing development code, can you return this to min?

py/static/index.html
```diff
@@ -25,12 +25,15 @@
...
+    <script src={{ static_url("js/react-dom.development.js") }}></script>
+
     <script src={{ static_url("fonts/layout_bin_packer") }}></script>
-    <script src={{ static_url("js/react-grid-layout.min.js") }}></script>
```
JackUrb: What happened to these files?

py/static/css/style.css
```diff
@@ -182,3 +182,664 @@ doesn't override our desire to make them non-interactable */
 .vertical-line {
   border-left: thin solid #ccc;
 }
+
+
+
+
+/*  ///////////////////////////////////////////////// */
```
JackUrb: This is a _lot_ of css content here for this feature, is there a reason we need it all? How much of it is actually relevant to the feature you're committing here?

It also has very non-standard formatting, with extra newlines before all curly brace closes and no lines between classes.

```
this{
    formatting: is

}
strange
```
```
this {
  formatting: is;
}

correct
```

It also uses different spacing standards than what we already have in this file.

py/server.py
```diff
@@ -695,6 +696,94 @@ def gather_envs(state):
...
+            del res['jsons'][destWid]
+
+    tbl = """"<style>
+    table, th, td {{
+        border: 1px solid black;
+    }}
+    </style>
+    <table> {} </table>""".format(' '.join(["<tr> <td> {} </td> <td> {}</td></tr>".format(v,eidNums[v]) for v in eidNums]))
+
+    res['jsons']['window_comapre_legend'] = {"command": "window",
```
JackUrb: Nit: `window_compare_legend`

These formatting segments (this and the above table creator) should be moved to a helper function, as they make it difficult to see what's going on here.

py/server.py
```diff
@@ -695,6 +696,94 @@ def gather_envs(state):
...
+    for k in res['jsons']:
+        if not res['jsons'][destWid]['has_compare']:
+            del res['jsons'][destWid]
+
+    tbl = """"<style>
+    table, th, td {{
+        border: 1px solid black;
+    }}
+    </style>
+    <table> {} </table>""".format(' '.join(["<tr> <td> {} </td> <td> {}</td></tr>".format(v,eidNums[v]) for v in eidNums]))
```
JackUrb: Nit: 79 char lines. 

This construction in particular is incredibly vague as far as what's happening on first glance. Please try to format in such a way that it's clear where you're creating rows, and further clear where you're filling out their content, instead of having two `'{}'.format` constructions in the same line.

py/server.py
```diff
@@ -695,6 +696,94 @@ def gather_envs(state):
...
+                for di,d in enumerate(res['jsons'][destWid]['content']['data']):
+                    res['jsons'][destWid]['content']['data'][di]['name'] = '{}_{}'.format(eidNums[eid], d['name'])
+            else:
+                res['jsons'][destWid]['has_compare'] = True
+                for di,d in enumerate(win['content']['data']):
+                    d = copy.deepcopy(d)
+                    d['name'] = '{}_{}'.format(eidNums[eid], d['name'])
+                    res['jsons'][destWid]['content']['data'].append(d)
+
+    for k in res['jsons']:
```
JackUrb: If I understand this properly, the compare construction will only display windows that are shared in title by at least 2 of the environments being compared? Can you add comments here and above to display your logic? 

(It's a nice design decision though 👍 )

py/server.py
```diff
@@ -695,6 +696,94 @@ def gather_envs(state):
...
+            destWid = name2Wid[title]
+
+            if ix == 0:
+                res['jsons'][destWid]['has_compare'] = False
+                res['jsons'][destWid]['content']['layout']['showlegend'] = True
+                for di,d in enumerate(res['jsons'][destWid]['content']['data']):
+                    res['jsons'][destWid]['content']['data'][di]['name'] = '{}_{}'.format(eidNums[eid], d['name'])
+            else:
+                res['jsons'][destWid]['has_compare'] = True
+                for di,d in enumerate(win['content']['data']):
```
JackUrb: Nit: `di, d`

Appears here and elsewhere. 

Also can you use more descriptive names? I recognize that not all of the visdom codebase is up to the standard of doing so, but I'm slowly trying to make it more easily understood.

py/server.py
```diff
@@ -695,6 +696,94 @@ def gather_envs(state):
...
+    for eid in eids:
+        if eid in state:
+            envs[eid] = state.get(eid)
+        else:
+            p = os.path.join(FLAGS.env_path, eid.strip(), '.json')
+            if os.path.exists(p):
+                env = tornado.escape.json_decode(open(p, 'r').read())
+                state[eid] = env
+                envs[eid] = env
+
```
JackUrb: Nit: 1 newline

package.json
```diff
@@ -1,20 +1,33 @@
 {
-  "name": "visdom",
```
JackUrb: nit: this file should not change spacing either

js/main.js
```diff
@@ -1,13 +1,24 @@
 /**
-* Copyright 2017-present, Facebook, Inc.
-* All rights reserved.
-*
-* This source code is licensed under the license found in the
-* LICENSE file in the root directory of this source tree.
-*
-*/
+ * Copyright 2017-present, Facebook, Inc.
```
JackUrb: nit: spacing should not change

js/main.js
```diff
@@ -78,7 +94,7 @@ class App extends React.Component {
 
   colWidth = () => {
     return (this.state.width - (MARGIN * (this.state.cols - 1))
-      - (MARGIN * 2)) / this.state.cols;
+            - (MARGIN * 2)) / this.state.cols;
```
JackUrb: nit: spacing should not change

js/main.js
```diff
@@ -122,7 +138,7 @@ class App extends React.Component {
 
     this.setState({
       panes: newPanes,
-      layout: newLayout,
+      layout: newLayout
```
JackUrb: nit: this should not change

js/main.js
```diff
@@ -132,7 +148,7 @@ class App extends React.Component {
 
     if (!exists) {
       let stored = JSON.parse(localStorage.getItem(this.keyLS(newPane.id)));
-
+      if (this._bin == null) this.rebin();
```
JackUrb: Nit: avoid if statements with single executions following, please use braces even for single statements.

js/main.js
```diff
@@ -169,21 +185,26 @@ class App extends React.Component {
...
 
     socket.onmessage = this._handleMessage;
 
     socket.onopen = () => {
-      this.setState({connected: true});
-    }
-
-    socket.onerror = socket.onclose = () => {
-      this.setState({connected: false}, () => {
+      this.setState({connected: true}, () => {console.log('callback');});
```
JackUrb: Nit: Remove callbacks once migrating to `componentDidUpdate` for the functionality you wanted, then you can leave this socket section how it was.

js/main.js
```diff
@@ -192,41 +213,41 @@ class App extends React.Component {
     var cmd = JSON.parse(evt.data);
 
     switch (cmd.command) {
-      case 'register':
```
JackUrb: Nit: spacing appears to have broken here

js/main.js
```diff
@@ -261,7 +282,7 @@ class App extends React.Component {
       let focusedPaneID = this.state.focusedPaneID;
       // Make sure we remove the pane from our layout.
       let newLayout = this.state.layout.filter(
-         (paneLayout) => paneLayout.i !== paneID)
+        (paneLayout) => paneLayout.i !== paneID)
```
JackUrb: 👍  looks like you caught one of mine hahaha

js/main.js
```diff
@@ -283,19 +305,46 @@ class App extends React.Component {
...
+    var envID = null;
+    if (selectedNodes.length > 0) envID = selectedNodes[0];
     this.setState({
       envID: envID,
+      envIDs: selectedNodes,
       saveText: envID,
       panes: isSameEnv ? this.state.panes : {},
       layout: isSameEnv ? this.state.layout : [],
-      focusedPaneID: isSameEnv ? this.state.focusedPaneID : null,
+      focusedPaneID: isSameEnv ? this.state.focusedPaneID : null
```
JackUrb: Nit: retain trailing comma

js/main.js
```diff
@@ -283,19 +305,46 @@ class App extends React.Component {
...
     });
+    localStorage.setItem( 'envID', envID );
```
JackUrb: Nit: no spacing around function arguments

js/main.js
```diff
@@ -283,19 +305,46 @@ class App extends React.Component {
...
     // to handle here. We might want to surface the error state.
-    $.post(this.correctPathname() + 'env/' + envID,
-      JSON.stringify({'sid' : this.state.sessionID}));
+    console.log('posting for env', envIDs);
+    if (envIDs.length == 1 ) {
+      $.post(this.correctPathname() + 'env/' + envIDs[0],
+             JSON.stringify({'sid' : this.state.sessionID}));
+    }
+    else if(envIDs.length > 1) {
+      $.post(this.correctPathname() + 'compare/' + envIDs.join('+'),
```
JackUrb, line 346: 👍

js/main.js
```diff
@@ -306,6 +355,7 @@ class App extends React.Component {
     });
   }
 
+
```
JackUrb: nit: remove newline

js/main.js
```diff
@@ -336,7 +386,7 @@ class App extends React.Component {
     for (var envIdx in newEnvList) {
       if (!layoutLists.has(newEnvList[envIdx])) {
         layoutLists.set(newEnvList[envIdx],
-          new Map([[DEFAULT_LAYOUT, new Map()]]));
+                        new Map([[DEFAULT_LAYOUT, new Map()]]));
```
JackUrb: Nit - repair spacing

js/main.js
```diff
@@ -593,167 +644,182 @@ class App extends React.Component {
 
   renderEnvModal() {
     return (
-      <ReactModal
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -593,167 +644,182 @@ class App extends React.Component {
...
+      contentLabel="Environment Management Modal"
+      ariaHideApp={false}
+      style={MODAL_STYLE}
+        >
         <span className="visdom-title">Manage Environments</span>
         <br/>
         Save or fork current environment:
         <br/>
         <div className="form-inline">
-          <input
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -593,167 +644,182 @@ class App extends React.Component {
...
+      onClick={this.saveEnv}>
+        {this.state.envList.indexOf(
+          this.state.saveText) >= 0 ? 'save' : 'fork'}
+      </button>
         </div>
         <br/>
         Delete environment selected in dropdown:
         <br/>
         <div className="form-inline">
-          <select
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -593,167 +644,182 @@ class App extends React.Component {
...
+        Delete
+      </button>
         </div>
       </ReactModal>
     );
   }
 
   renderViewModal() {
     return (
-      <ReactModal
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -593,167 +644,182 @@ class App extends React.Component {
...
+      contentLabel="Layout Views Management Modal"
+      ariaHideApp={false}
+      style={MODAL_STYLE}
+        >
         <span className="visdom-title">Manage Views</span>
         <br/>
         Save or fork current layout:
         <br/>
         <div className="form-inline">
-          <input
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -593,167 +644,182 @@ class App extends React.Component {
...
+      onClick={this.saveLayout.bind(this)}>
+        {this.getCurrLayoutList().has(
+          this.state.saveText) ? 'save' : 'fork'}
+      </button>
         </div>
         <br/>
         Delete layout view selected in dropdown:
         <br/>
         <div className="form-inline">
-          <select
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -593,167 +644,182 @@ class App extends React.Component {
...
+      label: x,
+      value: x
+    };}));
+
     return (
-      <span>
+        <span>
         <span>Environment&nbsp;</span>
         <div className="btn-group navbar-btn" role="group" aria-label="Environment:">
-          <div className="btn-group" role="group">
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -766,98 +832,98 @@ class App extends React.Component {
...
-            {check_space}
-          </a>
-        </li>;
+          {view}
+        {check_space}
+        </a>
+          </li>;
       }
-    )
+                                          )
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -766,98 +832,98 @@ class App extends React.Component {
         }
         return <li>
           <a href="#" onClick={this.updateToLayout.bind(this, view)}>
-            {view}
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -766,98 +832,98 @@ class App extends React.Component {
...
+          </li>;
       }
-    )
+                                          )
     return (
-      <span>
+        <span>
         <span>View&nbsp;</span>
         <div className="btn-group navbar-btn" role="group" aria-label="View:">
-          <div className="btn-group" role="group">
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -766,98 +832,98 @@ class App extends React.Component {
...
+        </button>
+        </div>
+        </span>
     )
   }
 
   renderFilterControl() {
     return (
-      <div className="input-group navbar-btn">
+        <div className="input-group navbar-btn">
```
JackUrb: Nit: broken spacing in this function

js/main.js
```diff
@@ -871,19 +937,19 @@ class App extends React.Component {
       let panelayout = getLayoutItem(this.state.layout, id);
       let isVisible = pane.title.match(this.state.filter)
       return (
-        <div key={pane.id}
```
JackUrb: Nit: broken spacing through this function too

js/main.js
```diff
@@ -894,61 +960,61 @@ class App extends React.Component {
...
-      </div>
     )
   }
 }
 
 
 function load() {
   ReactDOM.render(
-    <App />,
+      <App />,
```
JackUrb: Nit: broken spacing

js/main.js
```diff
@@ -957,9 +1023,9 @@ function load() {
 document.addEventListener('DOMContentLoaded', load);
 
 $(document).ready(function(){
-    $('[data-toggle="tooltip"]').tooltip({
```
JackUrb: Nit: broken spacing

pareshmg:
I'm making some of the changes you asked. I just want you to know that I use emacs and it auto-indents everything for me. That's why a lot of the indentation stuff changed (unintentionally). For example, the switch-case statement indents that way in emacs for me. Similarly, the return statement indentation in line 97 (which I think I prefer, but you're the owner, so I can adapt).  

Do you use emacs? Is there a js config you can send me so it can be uniform? For reference, I use spacemacs with javascript-mode which does all the indentation.

I'll do my best to undo the other indentation changes you suggested.

py/static/index.html
pareshmg: Not at all a js expert here, but I think the way I have it configured, it gets compiled into the main.js that is generated. I did it this way because I was having a lot of version conflicts with downloading the js files. yarn should handle the dependencies and compile it automatically.

pareshmg:
@JackUrb : regarding the placement of this in the EnvironmentModal: I view this as a replacement for the current environment selector. As the tree-select has the ability to search, it would be a highly desired feature even when someone is selecting just one environment. Also, wrt the 'X', the rc-tree-select plugin has little 'x's that appear when you hover over selected environments, so it is fairly easy to clear selected environments. 

It seems like comparing is a fairly common task and should be fewer clicks away from the user right? If you look at tensorboard, the ability to compare run is very easily available to the user. 

Also, some of the UI mangling you see may be because of the width I have selected as default for the TreeSelect (l185 in main.js). Maybe if we dynamically change that width you would be more okay with it?

What are your views on this? Are you open to the selector I have replacing the existing one with some of the above mentioned modifications?

JackUrb:
Hi @pareshmg - I primarily use Atom and Sublime for my development work, so I don't have a config that would be able to help you there. I could quickly reformat and push to this branch if you'd prefer that though.

pareshmg:
That would really help me out @JackUrb. Especially for main.js. I think I've fixed most of the other files.

JackUrb:
Alright I'll try the formatting in just a little. Re the mangling - my issue is that the selector's height changes when you use the tree selector, changing the height of the main bar. I'd really be alright with any implementation that doesn't do that.

pareshmg:
Just to double check - the height of the main bar doesn't change when only one environment is selected, but does change when multiple environments are selected. And what you are asking me is to make sure that the height doesn't change even when multiple environments are selected. Is that right?

JackUrb:
Correct - I'm betting this is doable by using some kind of scrolling within the selector, and then maybe expanding the view with an overlay on hover? A user should know what they're comparing by their legend though, so perhaps that's not necessary.

JackUrb:
OH another note - how does your compare env feature interact with views? I haven't gotten to test that just yet.

pareshmg:
I haven't used the views feature, so I'm not sure.

pareshmg:
Okay, so now the height is fixed. I also enabled the "clear" option where an "x" now appears to clear all options at once.

JackUrb:
Alright this is a lot more manageable now. I have to get to work on some other tasks, but I'll get back to this for another round of review tomorrow!

pareshmg:
Thanks for your tip on using componentDidUpdate. State persists across refreshes :)

js/main.js
```diff
@@ -55,7 +66,8 @@ class App extends React.Component {
     sessionID: null,
     panes: {},
     focusedPaneID: null,
-    envID: ACTIVE_ENV,
+    envID: localStorage.getItem( 'envID' ) || null,
+    envIDs: JSON.parse(localStorage.getItem( 'envIDs' )) || {},
```
JackUrb: This default value is not working and causing the whole visdom page to not load on a clean pull. I tried using the default value `[]` but that hasn't worked either. It appears the render method of `TreeSelect` in `renderEnvControls` is returning null or undefined, which is making React error out.

Can you ensure this is working on a fresh pull without the saved environment data in localstorage and with a blank environment? I won't be able to give an in-depth review on what remains until I can test it out for myself.

download.sh
```diff
@@ -0,0 +1,28 @@
...
+wget https://cdn.rawgit.com/plotly/plotly.js/master/dist/plotly.min.js -O py/static/js/plotly-plotly.min.js
+wget https://unpkg.com/bootstrap@3.3.7/dist/css/bootstrap.min.css -O py/static/css/bootstrap.min.css
+wget https://unpkg.com/bootstrap@3.3.7/dist/fonts/glyphicons-halflings-regular.eot -O py/static/fonts/glyphicons-halflings-regular.eot
+wget https://unpkg.com/bootstrap@3.3.7/dist/fonts/glyphicons-halflings-regular.woff2 -O py/static/fonts/glyphicons-halflings-regular.woff2
+wget https://unpkg.com/bootstrap@3.3.7/dist/fonts/glyphicons-halflings-regular.woff -O py/static/fonts/glyphicons-halflings-regular.woff
+wget https://unpkg.com/bootstrap@3.3.7/dist/fonts/glyphicons-halflings-regular.ttf -O py/static/fonts/glyphicons-halflings-regular.ttf
+wget "https://unpkg.com/bootstrap@3.3.7/dist/fonts/glyphicons-halflings-regular.svg#glyphicons_halflingsregular" -O py/static/fonts/glyphicons-halflings-regular.svg#glyphicons_halflingsregular
+wget "https://cdnjs.cloudflare.com/ajax/libs/react-modal/3.1.11/react-modal.min.js" -O py/static/js/react-modal.min.js
+wget "https://cdnjs.cloudflare.com/ajax/libs/react-select/1.2.1/react-select.min.js" -O py/static/js/react-select.min.js
```
JackUrb: I'm fine having this in a download script, but we should also include it in `download_scripts` in the server file.

js/main.js
pareshmg: do you have the web console error for this? I tried this in chrome and firefox and they both work for me when localstorage is cleared.

Also, I think the default value should be `{}`, not `[]`

download.sh
pareshmg: Is there a way around the download_scripts? I think the versioning gets a bit complicated with yarn. Do you know a way to "publish" the js file links or something like that? I wasn't very happy with my download.sh either

js/main.js
pareshmg: also, just to double check, you did a `yarn install` right? Want to make sure that the rc-tree-select package was added to node_modules

download.sh
JackUrb: We unfortunately can't package the files with ours underneath our license, and can't distribute them directly, so I don't know of another way we'd 'publish' them. The `download_scripts` function was supposed to make the process at least a little easier all things considered.

js/main.js
JackUrb: I had not, though it isn't actually changing anything (just tried with a yarn install). Ideally we shouldn't need to have a user run the install step as we hadn't needed to in the past.

screen shot 2018-03-01 at 11 01 52 am

We may be running into dependency issues here - how are you building and running visdom locally?

pareshmg: I'm using yarn (as was suggested in the contributing to the UI section. But the yarn.lock should automatically update the dependencies right?

And for the downloaded js scripts, you definitely want to update because I updated to the next version of react. Did you run the download.sh script that I had? Do that for now and then we can add it into the python file. I'm sorry this is a pain :(

download.sh
pareshmg: The biggest issue I see with this is the clash in versions between yarn.lock and what we download, but I guess as long as we are consistent with the versions there, we should be ok. I'll double check and update all the versions.

js/main.js
pareshmg: btw, this is my first time using yarn/npm. So if I'm doing things incorrectly, please feel free to correct me

JackUrb: Ah okay things are working now on after downloading your specific versions.

No worries about difficulty using yarn/npm - we primarily use them for the contributing side to make webpack and react work properly, but we try to leave usage of them out of the final distribution. In that way we're somewhat atypical. You'll be able to work this out by updating the version numbers listed in the `download_scripts` function, but at least for now things are working so I can do some testing.

JackUrb:
Alright a few notes now that I've been able to play around a bit. Looks like it is really coming along as a feature, but I still have a few issues:

1. The width of the tree select is incredibly wide, and on anything other than a giant browser it starts to mangle the layout. Can you set a width relative to the browser width, thus ensuring all the navbar components can appear inline?
2. Submenus don't quite make sense for environments that will be the only one in there submenu. Is this fixable?
screen shot 2018-03-01 at 12 10 38 pm
3. Selected environment should default to 'main', and if there is no 'main' then it should default to something rather than requiring a selection. Clearing the bar should probably return to this default rather than blank.
4. The feature of forking environments has been broken by the introduction of this feature. When a user has a single environment selected, they should be able to "fork" that environment under a different name (use a clean clone of visdom to observe this behavior). This behavior should remain operational in the new feature (if only one environment is selected, it can be forked, and that becomes the new selected environment). If multiple environments are selected, the environment modal should be disabled.
5. The feature of view selection is now _inconsistent_. It still works, and isn't broken, but when multiple environments are selected you're still able to change views (which has no actual effect on anything). The view selector should be disabled and set to display "current" when multiple environments are selected, and the view modal should be disabled as well.

pareshmg:
@JackUrb : For (3), I updated it to set to main by default when no env is in localStorage or when the length of selected elements is zero on a page refresh. However, I am keeping the feature where "clear" removes all selections including main because otherwise it messes with the selection. If they accidentally clear all and want to quickly get back to main, they can simply refresh the page and main will be selected agian.

Without this,  if you want to select env1 and env2, you will first have to select env1 and then env2 and then manually clear out main.

Is that okay?

I've fixed 1,2,3 and I'm looking into 4&5 (I haven't used these features)

JackUrb:
This is really coming along! Thanks for putting up with all these changes 😅 
 
1. There's still room for improvement here - as of now it only updates the size on hover, not on window resize. It's likely you can get a better estimate of how large it should be based on the size of the other elements and how much size is left in the top bar in general to handle how big it should render
2. This is great, thanks for this fix 💯 
3. Your choice here seems perfectly valid to me, the feature definitely feels more user-friendly overall at this point. 
4. Good - I'm not sure this is necessarily better than having a toggle between the old select and a compare select, though I'm going to ask @lvdmaaten to weigh in on that
5. This is almost exactly as it should be - Little nit: can you make it set the text on the disabled view selector to "Compare"?

pareshmg:
1. Done. Set to max(50, window.width/3)
5. done. now says compare

wrt 4: I like the current selector way more than the old selector because of the search function. If you look at the original image I sent, you see how many environments I have. Having the ability to filter even for a single environment is a HUGE plus for me.

JackUrb:
Oh I agree that in many if not most cases, people will prefer the new selector. I'm positing though that some amount may prefer the old selector, and wondering if it would be worthwhile to have a setting to use that one instead. Don't think it's necessarily required. 

Will get a look at the improvements later today

ArunkumarRamanan:
Well, I See...

pareshmg:
@JackUrb : what did you think of the improvements?

JackUrb:
Hi @pareshmg - sorry for the delay here, other work came up. This is really almost there from my perspective. I'm stepping back from the requirement that the treeselector remains at just one line as the transition with the scroll bar is somewhat clunky. Treeselect seems to natively support having a scrollbar when the height exceeds two lines, so perhaps that could be used instead: http://react-component.github.io/tree-select/examples/big-data.html?

pareshmg:
Figured out how to set the height through css. It now is 35px which is the height of the buttons around it. So now it's naive scrolling at one line.

pareshmg:
bump

JackUrb:
Hi @pareshmg - Sorry for the delays here, I'm waiting to get some feedback from @lvdmaaten who is working on a deadline that I'm holding off for. Sometime early next week I should be able to get you more context.

JackUrb:
Hi @pareshmg in the meantime I just realized that the clear x of the environment selector is obscured by the scrolling portion behind it: It might be possible to give it a slight right pad/margin to move it away from the bar a little?

screen shot 2018-03-12 at 2 14 55 pm

Also main.js is still packing with documentation rather than the minified 1 line we currently have, though I'm not quite sure why that's happening.

JackUrb:
Finally in - Thanks again @pareshmg!

PR #281, opened 2018-02-26
Moving module to subdir of py folder to fix pip install -e

JackUrb:
Fixes #280, as well as all the other times `pip install -e .` issues have ever come up. Seems that setuptools ignores the package_dir param when the sourced directory _is_ the package directory rather than _contains_ the package directory. While this somewhat dirties our directory structure, it does fix the issue.

PR #287, opened 2018-03-04
Visdom flag to suppress or raise exceptions

### LINKED ISSUE #241: Feature request: _send raises exception instead of printing it
In the `Visdom` class's `_send` method, if the post request raises an exception, the exception handler prints the exception and returns false instead of raising it.

It would be nice if the exception handler raised the exception and let the user handle it based on the use case. The exception raised could have useful contextual information instead of printing out the exception, as printing the exception may pollute loggers set up for the code.

varunagrawal:
Fixes #241

### DIFF
--- a/py/__init__.py
+++ b/py/__init__.py
@@ -342,8 +342,11 @@ class Visdom(object):
             if self.raise_exceptions:
                 raise ConnectionError("Error connecting to Visdom server")
             else:
-                DeprecationWarning("Visdom is eventually changing to default to raising exceptionsrather than "
-                                   "ignoring. This change is expected to happen by July 2018")
+                if self.raise_exceptions is None:
+                    warnings.warn("Visdom is eventually changing to default to raising exceptions rather than "
+                                  "ignoring. This change is expected to happen by July 2018. Please set "
+                                  "`raise_exceptions` to False to retain desired behavior.",
+                                  PendingDeprecationWarning)
                 if not quiet:
                     print("Exception in user code:")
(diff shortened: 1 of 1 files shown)


py/__init__.py
```diff
@@ -337,11 +339,16 @@ def _send(self, msg, endpoint='events', quiet=False):
...
             return r.text
         except BaseException:
-            if not quiet:
-                print("Exception in user code:")
-                print('-' * 60)
-                traceback.print_exc()
-            return False
+            if self.raise_exceptions:
+                raise ConnectionError("Error connecting to Visdom server")
+            else:
```
JackUrb, line 344: Ah this is really close - can we wrap the warning in an if here, where in this else if self.raise_exceptions is None you warn that the default is going to change and that users should set `raise_expectations` to `False` to retain their desired behavior? (though `None` and `False` will have the same behavior other than the warning)

JackUrb:
Going to fix the issue inside.

py/__init__.py
varunagrawal: Oh I'm just seeing this. I can get it done by later tonight if that's fine.

JackUrb: That also works, no rush on it really. I'm just trying to clean up our PRs before I get #281 in next week so that I prevent annoying merge conflicts. If it's not done then I'll just do it 👍

varunagrawal: Alright, I've added the recommendation. @JackUrb let me know if this looks good.

PR #291, opened 2018-03-09
Modular start

dzimmerer:
Added start_server function which does not depend not parsed arguments, so that the start server function can be called from another python script


lvdmaaten:
Thanks for this @dzimmerer! We will need some time to test this to make sure it doesn't break anything in the version of visdom we use internally. We'll get back to you on this PR shortly after the weekend.

JackUrb:
```
from visdom import server
...
server.start_server()
```

PR #299, opened 2018-03-15
fix some bugs about plot multiple point in one line 

lixinsu:
According the documents

> . It takes as input an N or NxM tensor Y that specifies the values of the M lines (that connect N points) to plot 

I find that if I call line funciton as bellow 
   viz.line(X=np.array([1,2]), Y=np.array([1,2]), win='xx', name='linename') to plot (1,1)  (2,2) on one single  line 
it will  rename it as "1"  rather than "Linename"  

And when you want to add single point on single line as below
 viz.line(X=np.array([[1]]), Y=np.array([[1]]), win='xx', name='linename') 
It will  report error

So , I change the code as below 
1. when intput with Y.shape=(dim1, 1), We  reshape Y  as Y.shape=(dim1) 
2. The origin  code that decide  whether use the number to  name the line is done as below 
elif X.shape[0] == 1 and name is not None:
This will make a mistake when you try to plot multiple points in one line with name.

### DIFF
--- a/py/__init__.py
+++ b/py/__init__.py
@@ -801,6 +801,7 @@ class Visdom(object):
                     Y = None
 
         assert X.ndim == 2, 'X should have two dims'
+
         assert X.shape[1] == 2 or X.shape[1] == 3, 'X should have 2 or 3 cols'
 
         if Y is not None:
@@ -925,6 +926,11 @@ class Visdom(object):
             else:
                 assert X is not None, 'must specify x-values for line update'
         assert Y.ndim == 1 or Y.ndim == 2, 'Y should have 1 or 2 dim'
+        if Y.ndim == 2:
+            assert Y.shape[1] > 0, 'must plot one line at least'
+            if Y.shape[1] == 1:
+                Y = Y.reshape(Y.shape[0])
+                X = X.reshape(X.shape[0])
 
         if X is not None:
             assert X.ndim == 1 or X.ndim == 2, 'X should have 1 or 2 dim'
--- a/py/__init__.py
+++ b/py/__init__.py
@@ -926,14 +926,15 @@ class Visdom(object):
                 assert X is not None, 'must specify x-values for line update'
         assert Y.ndim == 1 or Y.ndim == 2, 'Y should have 1 or 2 dim'
         assert Y.shape[-1] > 0, 'must plot one line at least'
-        if Y.ndim == 2 and Y.shape[1] == 1:
-                Y = Y.reshape(Y.shape[0])
-                X = X.reshape(X.shape[0])
 
         if X is not None:
             assert X.ndim == 1 or X.ndim == 2, 'X should have 1 or 2 dim'
         else:
             X = np.linspace(0, 1, Y.shape[0])
+
+        if Y.ndim == 2 and Y.shape[1] == 1:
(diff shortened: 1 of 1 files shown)


lixinsu:
`from visdom import Visdom
import numpy as np
import time
import sys


viz = Visdom(port=8093)

# test single line



win = viz.line(
    X=np.array([1,2,3,4,5]),
    Y=np.array([1,2,3,4,5]),
    win="test",
    name='Line1',
)

win = viz.line(
    X=np.array([[1],[2],[3],[4],[5]]),
    Y=np.array([[1],[2],[3],[4],[5]]),
    win="test",
    name='Line2',
    update='append'
)

# tes single line single point
win = viz.line(
    X=np.array([6]),
    Y=np.array([6]),
    win="test",
    name='Line1',
    update='append'
)

win = viz.line(
    X=np.array([[6]]),
    Y=np.array([[6]]),
    win="test",
    name='Line2',
    update='append'
)

# test multiple line

win = viz.line(
    X=np.array([[7, 7]]),
    Y=np.array([[8, 8]]),
    win="test",
    update='append',
    opts={'legend':['Line1','Line2']}
)

win = viz.line(
    X=np.array([[8, 8]]),
    Y=np.array([[3, 1]]),
    win="test",
    update='append',
    opts={'legend':['Line1','Line2']}
)
`
this is my test code

JackUrb:
Pretty close on your test cases - can you try the case of multiple lines at once in the same call?

From the demo:
```
    win = viz.line(
        X=np.column_stack((np.arange(0, 10), np.arange(0, 10))),
        Y=np.column_stack((np.linspace(5, 10, 10),
                           np.linspace(5, 10, 10) + 5)),
    )
```

lixinsu:
newtest
this is the test result,it is ok

py/__init__.py
```diff
@@ -925,6 +925,10 @@ def line(self, Y, X=None, win=None, env=None, opts=None, update=None,
             else:
                 assert X is not None, 'must specify x-values for line update'
         assert Y.ndim == 1 or Y.ndim == 2, 'Y should have 1 or 2 dim'
+        assert Y.shape[-1] > 0, 'must plot one line at least'
+        if Y.ndim == 2 and Y.shape[1] == 1:
+                Y = Y.reshape(Y.shape[0])
+                X = X.reshape(X.shape[0])
```
JackUrb: Ack! `X` could be `None` here. You should reshape `X` _after_ line 936. Perhaps move this if `Y` block to after there?

JackUrb: Try
```
 win = viz.line(
        Y=np.column_stack((np.arange(0, 10), np.arange(0, 10))),
    )
```
and 
```
 win = viz.line(
        Y=np.arange(1, 10)
    )
```

lixinsu:
I modified the code, is this ok?

JackUrb:
Thanks for the fixes!

JackUrb:
I think DST is messing up our import script, I'll have to pull it in in an hour

JackUrb:
Okay the commit is _still_ appearing below my approve so it's going to take even longer

JackUrb:
It appears your commit has a time stamp of 3 hours and 20 minutes from now, so I can't land until then

PR #308, opened 2018-04-08
Image pan and zoom while allowing page to scroll

pareshmg:
I was getting annoyed when scrolling down would result in an image getting zoomed in. Also, sometimes, when I would scroll up, the image would zoom out to size zero and the only way to reset was to reload the page. Additionally, there was no way to pan the image after a zoom.

So I did the following:

* Shift + scroll = pan (shift + two finger move on trackpad)
* Ctrl + scroll = zoom (ctrl + two finger move OR pinch zoom on trackpad)
* Allow pan & zoom when on the Pane, not just on the image
* Double click on pane (not just image) to reset image
* Reset button in pane bar to reset image


pareshmg:
Sorry, I don't know why so many commits showed up here. It's actually just the last few. The other commits were from my previous pull request and me trying to merge my repo with the official repo (upstream)

.gitignore
```diff
@@ -2,3 +2,10 @@ node_modules
 build
 th/CMakeLists.txt
 th/build.luarocks
+dist/
+visdom*.tgz
+visdom.egg-info/
+setup.cfg
+py/static/fonts/
+py/static/css/
+py/static/js/
```
JackUrb: The main.js file produced into this folder is actually required for committing, so we probably shouldn't ignore the whole folder

js/ImagePane.js
```diff
@@ -24,37 +26,54 @@ class ImagePane extends React.Component {
...
-          onWheel={this.handleZoom.bind(this)}
-          onDoubleClick={this.resetZoom.bind(this)}
-        />
+        <div style={divstyle}>
+          <img
+            className="content-image cssTransforms"
+            src={content.src}
+            width={Math.ceil(1 + this.props.width * this.state.scale) + "px"}
+            height={Math.ceil(1 + this.props.height * this.state.scale) + "px"}
+            onWheel={this.handleZoom.bind(this)}
```
JackUrb: If we're changing zoom being scroll dependent this should also be removed right?

js/Pane.js
```diff
@@ -77,11 +89,14 @@ var classNames = require('classnames');
...
       <div className={windowClassNames}
         onClick={this.focus}
+        onDoubleClick={this.reset}
+        onWheel={this.zoom}
         ref={(ref) => this._windowRef = ref}>
         <div className={barClassNames}
           ref={(ref) => this._barRef = ref}>
           <button title="close" onClick={this.close}>X</button>
           <button title="save" onClick={this.download}>&#8681;</button>
+          <button title="reset" onClick={this.reset} hidden={!this.props.handleReset}>&#10226;</button>
```
JackUrb, line 99: Ah this is great! I can imagine other panes might use this reset feature somewhere down the road. 👍

js/ImagePane.js
pareshmg: This calls the function that separates the two and is harmless, but unnecessary. I took this out as it is cleaner to just pass the event up to the parent. (behavior remains same whether this line exists or not)

JackUrb:
Alright I've gotten a chance to try this out and its *awesome*. Only thing, can we swap ctrl and shift and check for the pan event before the zoom event? Shift scroll is already overloaded on many browsers as horizontal scroll, so you'll only be able to scroll left and right if using a mouse wheel under the current setting. The ordering is important as well, as shift control scroll should pan left/right rather than zoom

PR #319, opened 2018-04-20
Adding readonly support

### LINKED ISSUE #256: [Feature Request] Add 'readonly' flag to visdom.server
How can I share my work with other people such that they could only _view_ the plots? It'd be nice to have the ability to block any changes made by a user to the server.

xssChauhan:
Running visdom server with `readonly` flag starts the server in a mode that does not allow the client to perform any modifications. 
Using the flag: `python -m visdom.server -readonly`
This is achieved with the following:

- Block all the incoming `cmds` when started with `readonly` flag
- Following UI interactions are blocked as well

    - Closing single/all panes
    - Forking the environment
    - Deleting an environment
    - Broadcast of `keyPress` events
 
Closes #256


js/main.js
```diff
@@ -917,7 +925,7 @@ class App extends React.Component {
             title="Manage Environments"
             data-placement="bottom"
             className="btn btn-default"
-            disabled={!(this.state.connected && this.state.envID)}
+            disabled={!(this.state.connected && this.state.envID && !this.state.readonly)}
```
JackUrb, line 928: This is great! Should also be applied to 986 to disable the env modal as well.

xssChauhan: Thank you for the review. Shall add the suggested change as well.

PR #330, opened 2018-05-02
My idea for authentication method

rubiagatra:
I am adding authentication method when there is -with_login flag


JackUrb:
Hi @rubiagatra - Thanks for putting out a submission. Before I get started on reviewing, can you please remove the files that were supposed to be ignored by git (.DS_Stores, __pycache__, and /static/* files that are downloaded by the server rather than written by you). I can review afterwards

rubiagatra:
Okay I am working on it @JackUrb

py/visdom/server.py
```diff
@@ -928,15 +931,50 @@ def initialize(self, app):
...
 
     def get(self, args, **kwargs):
         items = gather_envs(self.state, env_path=self.env_path)
-        self.render(
-            'index.html',
-            user=getpass.getuser(),
-            items=items,
-            active_item=''
-        )
+        if self.with_login == False:
```
sampepose: Can you please merge this logic?

```
if (not self.with_login) or (self.with_login and self.current_user):
  # render index.html
else:
  # render login.html
```

Note that I used `self.current_user`, which is described here: http://www.tornadoweb.org/en/stable/guide/security.html#user-authentication

py/visdom/server.py
```diff
@@ -928,15 +931,50 @@ def initialize(self, app):
...
+                items=items,
+                active_item=''
+            )
+
+    def post(self, arg, **kwargs):
+        username = self.get_argument("username") 
+        password = hashlib.sha256(self.get_argument("password").encode("utf-8")).hexdigest()
+
+        if (username == self.user_login["username"]) and (password == self.user_login["password"]):
+            self.user_login["login"] = True
```
sampepose: Again, given http://www.tornadoweb.org/en/stable/guide/security.html#user-authentication, we should instead set a cookie here:

```
self.set_secure_cookie("user", self.get_argument("name"))
```

Please remove this boolean flag.

You'll also need to add

```
def get_current_user(self):
        return self.get_secure_cookie("user")
```

here: https://github.com/facebookresearch/visdom/pull/330/files#diff-ddbe57b92511a8307bcf5999eb519e6cR299

py/visdom/server.py
```diff
@@ -1081,7 +1122,18 @@ def main(print_func=None):
...
     logging.getLogger().setLevel(logging_level)
 
-    start_server(port=FLAGS.port, env_path=FLAGS.env_path, readonly=FLAGS.readonly, print_func=print_func)
+    if FLAGS.with_login:
+        username = input("Please input your username: ")
+        password = input("Please input your password: ")
+        user = {
+            "username": username,
+            "password": hashlib.sha256(password.encode("utf-8")).hexdigest(),
+            "login": False
```
sampepose: remove

rubiagatra:
please review my newest commit @JackUrb @sampepose

py/visdom/server.py
```diff
@@ -97,13 +99,15 @@ def serialize_all(state, env_path=DEFAULT_ENV_PATH):
 
 
 class Application(tornado.web.Application):
-    def __init__(self, port=DEFAULT_PORT, env_path=DEFAULT_ENV_PATH, readonly=False):
+    def __init__(self, port=DEFAULT_PORT, env_path=DEFAULT_ENV_PATH, readonly=False, with_login=False, user=False):
```
JackUrb: Rather than require passing a with_login parameter, you can use `user=None` and then check in the init method if `user` is set, and use that to determine if `with_login` is true. 

Also please use `login_enabled` and `user_credentials` as your variable names.

py/visdom/server.py
```diff
@@ -1039,6 +1075,8 @@ def main(print_func=None):
                              'level name or int (example: 20)')
     parser.add_argument('-readonly', help='start in readonly mode',
                         action = 'store_true')
+    parser.add_argument('-with_login', default=False,
```
JackUrb: Name the argument `-enable-login`, but store it in `login_enabled`

py/visdom/server.py
```diff
@@ -1053,7 +1091,17 @@ def main(print_func=None):
...
-    start_server(port=FLAGS.port, env_path=FLAGS.env_path, readonly=FLAGS.readonly, print_func=print_func)
+    if FLAGS.with_login:
+        username = input("Please input your username: ")
+        password = input("Please input your password: ")
+        user = {
+            "username": username,
+            "password": hashlib.sha256(password.encode("utf-8")).hexdigest(),
+        }
+    else:
+        user = False
```
JackUrb: Prefer `None` rather than `False` for uninitialized data

py/visdom/server.py
```diff
@@ -1053,7 +1091,17 @@ def main(print_func=None):
...
+    
+    start_server(port=FLAGS.port, env_path=FLAGS.env_path, readonly=FLAGS.readonly, print_func=print_func, with_login=FLAGS.with_login, user=user)
```
JackUrb: Line length (this line might've already been too long, but try to keep it <80)

py/visdom/server.py
```diff
@@ -887,15 +894,43 @@ def initialize(self, app):
...
     def get(self, args, **kwargs):
         items = gather_envs(self.state, env_path=self.env_path)
-        self.render(
-            'index.html',
-            user=getpass.getuser(),
-            items=items,
-            active_item=''
-        )
+
+        if (not self.with_login) or self.with_login and self.current_user:
```
JackUrb: Where is `self.current_user` set? Is this handled by tornado? As this is somewhat unclear (it isn't set in this file anywhere), please add a comment about where it's coming from for future contributors 👍.

py/visdom/static/css/style.css
```diff
@@ -199,4 +199,4 @@ doesn't override our desire to make them non-interactable */
 
 .rc-tree-select-selection--multiple .rc-tree-select-search__field__placeholder {
     top: 8px;
-}
```
JackUrb: keep trailing newline

py/visdom/server.py
```diff
@@ -1053,7 +1091,17 @@ def main(print_func=None):
 
     logging.getLogger().setLevel(logging_level)
 
-    start_server(port=FLAGS.port, env_path=FLAGS.env_path, readonly=FLAGS.readonly, print_func=print_func)
+    if FLAGS.with_login:
+        username = input("Please input your username: ")
+        password = input("Please input your password: ")
```
JackUrb: As a better practice, please use the `getpass` library to pull the password, this way it isn't just displayed in plaintext in terminal

py/visdom/server.py
```diff
@@ -295,6 +299,9 @@ def __init__(self, *request, **kwargs):
         self.include_host = True
         super(BaseHandler, self).__init__(*request, **kwargs)
 
+    def get_current_user(self):
```
JackUrb, line 315: Add a docstring here, as this isn't referenced anywhere else.

In combination with self.current_user I'm getting the idea that tornado calls this function to populate `current_user`, but that's unclear from the code.

py/visdom/server.py
```diff
@@ -80,7 +81,8 @@ def extract_eid(args):
     "debug": "/dbg/" in __file__,
     "static_path": get_path('static'),
     "template_path": get_path('static'),
-    "compiled_template_cache": False
+    "compiled_template_cache": False,
+    "cookie_secret": "YOUR SECRET HERE"
```
JackUrb: Try to retrieve this key from a `COOKIE_SECRET` file in the `'%s/.visdom/' % expanduser("~")` directory. If that file doesn't exist, have the user create the file the first time they try using a user/password near line 1094 (and then update the global `tornado_settings` there).

rubiagatra:
Working on it @JackUrb, Thank you for your feedback! 👍

rubiagatra:
Already fix the requested changes @JackUrb, please review the last commit. About the last one, I created docstring below the function. Thank you

py/visdom/server.py
```diff
@@ -80,10 +81,9 @@ def extract_eid(args):
     "debug": "/dbg/" in __file__,
     "static_path": get_path('static'),
     "template_path": get_path('static'),
-    "compiled_template_cache": False
+    "compiled_template_cache": False,
 }
 
-
```
JackUrb: keep line break

py/visdom/server.py
```diff
@@ -1053,7 +1102,23 @@ def main(print_func=None):
...
+    if FLAGS.enable_login:
+        username = input("Please input your username: ")
+        password = getpass.getpass(prompt="Please input your password: ")
+        user_credential = {
+            "username": username,
+            "password": hashlib.sha256(password.encode("utf-8")).hexdigest(),
+        }
+
+        if not os.path.isfile(DEFAULT_ENV_PATH + "COOKIE_SECRET"):
+            with open(DEFAULT_ENV_PATH + "COOKIE_SECRET", "w") as file:
```
JackUrb: Avoid the name `file` as it is nondescript (and is a reserved name for `file` objects). Use `cookie_file`.

py/visdom/server.py
```diff
@@ -1053,7 +1102,23 @@ def main(print_func=None):
...
+                file.write(user_credential["username"] + user_credential["password"])
```
JackUrb: Request a cookie secret here from the user rather than using one created by the provided credentials. 

Also please provide a flag `-force_new_cookie` that users can use that will allow them to re-create this cookie, as it will allow visdom users to invalidate the cookies of previous users and sessions. As things stand a user who is validated to a visdom server will be able to access any future servers for a month even if the login changes, and there's no clear way for the server owner to fix this.

py/visdom/static/incorrect.html
```diff
@@ -0,0 +1,29 @@
...
+
+  <body class="text-center">
+    <form class="form-signin"> 
+      <h1 class="h3 mb-3 font-weight-normal">Visdom Login</h1>
+      <label for="inputUsername" class="sr-only">Email address</label>
+      <input name=username type="username" id="inputUsername" class="form-control" placeholder="Username" required autofocus>
+      <label for="inputPassword" class="sr-only">Password</label>
+      <input name=password type="password" id="inputPassword" class="form-control" placeholder="Password" required>
+      <button class="btn btn-lg btn-primary btn-block" type="submit", formmethod="post">Login</button>
+      <p>YOUR PASSWORD INCORRECT</p>
```
JackUrb: Message should be "Invalid login information" and not all caps.

rubiagatra:
Hello @JackUrb, I fixed the requested changes. Please review my last commit.

This is the login page when a user enables `enable_login` flag

screen shot 2018-05-16 at 09 51 07

and this is when a user does not input correct username and password

screen shot 2018-05-16 at 09 51 23

Thank you

JackUrb:
@lvdmaaten makes a very good point here - just be sure to double-hash the password that is initially set up for the server as well!

rubiagatra:
working on it @JackUrb , @lvdmaaten , thank you for your feedback 👍

rubiagatra:
Hello @JackUrb, @lvdmaaten  I fixed the requested changes. Please review my last commit.

py/visdom/server.py
```diff
@@ -930,6 +979,8 @@ def download_scripts(proxies=None, install_dir=None):
         # here is another url in case the cdn breaks down again.
         # https://raw.githubusercontent.com/plotly/plotly.js/master/dist/plotly.min.js
         'https://cdn.plot.ly/plotly-latest.min.js': 'plotly-plotly.min.js',
+        # Stanford Javascript Crypto Library for Password Hashing
+        '%ssjcl@1.0.7/sjcl.js' % b: 'sjcl.js',
```
JackUrb, line 983: Likely the last comment - can you add this file to our `download.sh` script, so that those who download with the script will be able to get it as well?

rubiagatra: I am sorry, I am not aware with `download.sh`, working on it @JackUrb

rubiagatra: done @JackUrb

JackUrb:
Thanks for the updates @rubiagatra - going to do some testing internally and then I'll merge this!

Issue #329, opened 2018-05-01
get_window_data returns invalid json (with suggested fix)


davidwagnerkc:
https://github.com/facebookresearch/visdom/blob/8a3a6a4a8928693a240cf16efa6106f2a01a5827/py/visdom/server.py#L907-L917

I think 914-917 need to be removed. Looks like they were copied from `win_exists()`, but maybe left on accident. 

If the selected `win` does exist then the result is JSON + "true" appended on the end which makes the json invalid (json.loads() fails):

![image](https://user-images.githubusercontent.com/20519655/39486683-a0d4d126-4d42-11e8-852b-564bd714bc89.png)

If the selected `win` does not exist then the assert fails so "false" never gets written. The returned result is an empty string with an exception logged on the server. 

I removed the lines locally and `json.loads(vis.get_window_data('compare'))` works like a champ and passing in a `win` that does not exist prints exception on server logs and returns an empty string. This seems like fine behavior if the `win` doesn't exists, but I'd leave that decision you guys. 

Related to PR: #272

JackUrb:
This is a good catch, thanks! If you want to submit that fix as a PR I'd be happy to accept, otherwise I'll do it once I'm back at my laptop

JackUrb:
Closed by #334

PR #334, opened 2018-05-03
Return valid json data for get_window_data

davidwagnerkc:
Fix for #329
json.loads() now works correctly and getting a non-existent window will return an empty string

image


JackUrb:
Thanks a lot for this catch @davidwagnerkc!

PR #347, opened 2018-05-10
textlabels bugs in scatter?

cympfh:
When `Y` is set, `textlabels` not works expectedly.


JackUrb:
Hi @cympfh - This looks great and seems to fix our bug for supplying multiple text labels to a scatter plot. Can you put a test example into `demo.py`? With that and you signing the CLA I'd be happy to accept this 👍

PR #349, opened 2018-05-11
Use MathJax from cloudflare instead of cdn.mathjax

spMohanty:
Reference : https://www.mathjax.org/cdn-shutting-down/


spMohanty:
@JackUrb : `download.sh` already seems to point to `MathJax` from the CloudFlare CDN (reference : https://github.com/facebookresearch/visdom/commit/cc55d3c43b31d7b46f438f6ebf61999f26b247eb https://github.com/facebookresearch/visdom/blob/master/download.sh#L8 )

JackUrb:
@spMohanty Ah sorry about that, just assumed we had it wrong in both places. Pulling in now (last commit before 0.1.8!). Thanks again!

PR #352, opened 2018-05-14
Delay initialization until socket finished, make back socket optional

### LINKED ISSUE #350: [Bug] Issue with `Visdom.check_connection` method on the first call
Hi, 

There is an issue appeared with the version 0.1.8.1 such that in one terminal I run :
```
# python3 -m visdom.server
Downloading scripts. It might take a while.
It's Alive!
INFO:root:Application Started
You can navigate to http://68a369ae3552:8097
```
and without exiting the server in the next terminal I run multiple times `vis.check_connection()` to check the connection:  
```
# python3 -c "import time, visdom; vis=visdom.Visdom(); print([vis.check_connection() for _ in range(5)])"
> [False, True, True, True, True]

# python3 -c "import time, visdom; vis=visdom.Visdom(); print([vis.check_connection() for _ in range(5)])"
> [False, True, True, True, True]
```
same in the console:
```
Python 3.5.2 (default, Nov 23 2017, 16:37:01)
>>> import visdom; v = visdom.Visdom(); print(v.check_connection())
False
>>> import visdom; v = visdom.Visdom(); print(v.check_connection())
True
>>> import visdom; v = visdom.Visdom(); print(v.check_connection())
True
```
Probably, it is not a correct behaviour. Any suggestions to workaround, please ?
Thanks

JackUrb:
Adds an option to not use the backwards socket to receive data from the visdom server to the python client (though by default the socket is enabled).

Delays returning from the `__init__` call until a message has been rebounded by the socket or the connection has been refused (or 5 seconds have passed and the request times out).

Bumps version

Fixes #350


py/visdom/__init__.py
```diff
@@ -270,7 +271,7 @@ def __init__(
         self.send = send
         self.event_handlers = {}  # Haven't registered any events
         self.socket_alive = False
-        self.use_socket = True
+        self.use_socket = use_incoming_socket
```
lvdmaaten, line 274: What will happen when the user sets `self.use_socket` to `False` and then tries to register an event handler? Will the `register_event_handler` raise a not-supported-exception or something of that kind, or will it err out with a time-out or could-not-connect type of exception?

Also, we may have to warn the user here that by setting `self.use_socket` to `False`, certain functionality is not supported.

py/visdom/__init__.py
```diff
@@ -281,6 +282,19 @@ def __init__(
 
         if send:  # if you're talking to a server, get a backchannel
             self.setup_socket()
+        # Wait for initialization before starting
+        time_spent = 0
+        while self.use_socket and not self.socket_alive and time_spent < 5:
+            time.sleep(0.1)
```
lvdmaaten: It is not really important, but this is typically implemented by doubling the sleep time each time you sleep.

py/visdom/__init__.py
JackUrb: Ah this is a good point - I'll add more messaging in here surrounding this option.

py/visdom/__init__.py
JackUrb: Sounds good to me, have done that in the past but didn't think about it here.

py/visdom/__init__.py
JackUrb: Event handlers are registered locally, so this doesn't actually require a server connection to establish, as opening the socket means you get any events you register for when creating the windows. I've made sure you can only register event handlers when the socket will exist though

PR #361, opened 2018-05-23
Properties pane added

lopisan:
A properties-like grid pane was added to enable setting/changing values in python/lua backend. 

* A `PropertiesPane` was added, displayed using `vis.properties()`
* New `PropertyUpdate` event added to propagate events to backend

### Changes to existing code
* Key press events interfered with textbox input in property pane, so I was forced to move the key events from main canvas to `TextPane`. Main canvas still catches key events, but just publishes them to newly added `EventSystem` (simple event queue implementation). `TextPane` subscribes to the queue and thru `appApi` sends the event as was before.
* `appApi` was added to props of Panes containing just `sendPaneMessage` - a message from the pane that will be forwarded to backend (with the evironment and pane id added)
* CSS styling of focused pane changed - now just a widow header bar (title) is higlighted on focused window, not window content (it was ugly in properties pane and)

Example of usage is in `demo-properties.py`. It shows that even a backend validation can be done (when `PropertyUpdate` event is received, the `viz.properties()` can be called again with previous or updated values.


example/demo-properties.py
```diff
@@ -0,0 +1,93 @@
...
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from visdom import Visdom
+import numpy as np
```
JackUrb: Remove unused imports (numpy, math, os.path, getpass, platform, urllib)

example/demo-properties.py
```diff
@@ -0,0 +1,93 @@
...
+                new_value = value + '_updated'
+            elif prop_id == 1:
+                new_value = value + '0'
+            elif prop_id == 2:
+                new_value = 'Stop' if properties[prop_id]['value'] == 'Start' else 'Start'
+            else:
+                new_value = value
+            properties[prop_id]['value'] = new_value
+            viz.properties(properties, win=properties_window)
+            viz.text(f"Updated: {properties[event['propertyId']]['name']} => {str(event['value'])}",
```
JackUrb: Please use python2 compliant formatting (esp for demos) and keep line length < 80 for python (line length also an issue on lines 56 and 70)

js/TextPane.js
```diff
@@ -11,6 +13,35 @@ const Pane = require('./Pane');
 
 class TextPane extends React.Component {
 
+  onEvent = (e) => {
```
JackUrb, line 15: Global events make sense to me! Can this particular event also be added to `ImagePane`s as well? We have a fairly common use case of attaching listeners to images to be able to arrow through them

js/main.js
```diff
@@ -606,20 +610,22 @@ class App extends React.Component {
     this.setState({layoutLists: layoutLists, layoutID: layoutID});
   }
 
-  broadcastKeyEvent = (event) => {
+  publishEvent = (event) => {
+    EventSystem.publish('global.event', event);
+  }
+
+  sendPaneMessage = (data) => {
```
JackUrb: Can you add a comment about how sendPaneMessage is being used as the event system API?

py/visdom/__init__.py
```diff
@@ -514,6 +514,45 @@ def text(self, text, win=None, env=None, opts=None, append=False):
             'opts': opts,
         }, endpoint=endpoint)
 
+    def properties(self, data, win=None, env=None, opts=None):
+        """
+        This function shows editable properties in a pane. Properties are expected to be a List of Dicts e.g.:
```
JackUrb: Docblock line length

example/demo-properties.py
lopisan: fixed

example/demo-properties.py
lopisan: fixed

py/visdom/__init__.py
lopisan: fixed (also in `init.lua `)

js/TextPane.js
lopisan: I've added the event handling to `ImagePane`. Also a demonstration to `demo-properties.py` was added.

js/main.js
lopisan: Comments added. As I was writing it, I realized that only focused panes should be calling this method (it adds `target` and `eid` of focused pane - not of the calling pane). I've done it this way to minimize changes to code (and I think for now it could be used), but maybe it's worth consideration not to add `target` and `eid` in this method, but just let the Pane fill it (it has paneId, but what with envId?)

lopisan:
Hi, I'm glad you like the feature, I've fixed things you've mentioned in comments.

js/main.js
JackUrb: This is a really good point - I can see down the road it might be useful for panels that aren't focused to be sending events, however the current implementation is definitely workable for the events we have. A change to support those down the road will include telling panes what their env_id is, however that could all invoke a much harder expectation (like background panes from non-active environments being able to send/receive events, and we're definitely not there).

js/EventSystem.js
```diff
@@ -0,0 +1,43 @@
+class EventSystem {
```
JackUrb, line 10: Ah just realized this file is missing the copyright

lopisan: Fixed

JackUrb:
Awesome - thanks for the fixes and the feature in general! We'll push it out to PyPI once we have more done of our `0.1.9` plan

PR #364, opened 2018-05-27
Add 'resisable' option for matplot

### LINKED ISSUE #363: Resizable SVG for matplotlib plots
A simple change [here](https://github.com/facebookresearch/visdom/blob/master/py/visdom/__init__.py#L608) can be made to allow resizable SVG (it will take the size of the pane). All we need to do is to remove width and height attributes, something like this:
```python
if resizable:
    import bs4
    soup = bs4.BeautifulSoup(svg, 'xml')
    soup.svg.attrs.pop('width', None)
    soup.svg.attrs.pop('height', None)
    svg = str(soup)
```

dmitriy-serdyuk:
Fixes #363

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,8 +365,12 @@ the name of an SVG file `svgfile`. The function does not support any specific
 `opts`.
 
 #### vis.matplot
-This function draws a Matplotlib `plot`. The function does not support
-any plot-specific `opts`.
+This function draws a Matplotlib `plot`. The function supports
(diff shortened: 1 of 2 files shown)


JackUrb:
Hi @dmitriy-serdyuk! Sorry for the delay here - taking a look now 👍

dmitriy-serdyuk:
@JackUrb , done

JackUrb:
Thanks for the update!

Issue #367, opened 2018-06-06
Visdom & Pillow-simd


0phoff:
I am using pillow-simd, which is a drop-in replacement for pillow that offers a speedup for certain operations by using SIMD operations of your CPU.  
However, when installing visdom, it installs pillow, which overwrites my pillow-simd install.

They solved this in pytorch/vision, by checking if pillow-simd is installed on the system in __setup.py__.
```python
from pkg_resources import get_distribution, DistributionNotFound

def get_dist(pkgname):
    try:
        return get_distribution(pkgname)
    except DistributionNotFound:
        return None

requirements = [ ... ]
pillow_ver = ' >= 4.1.1'
pillow_req = 'pillow-simd' if get_dist('pillow-simd') is not None else 'pillow'
requirements.append(pillow_req + pillow_ver)
```

I think the only drawback of using this, is if you provide wheels, as it would then be unable to determine whether you use pillow-simd or not.  
I guess this is no concern here right?

JackUrb:
Hi @0phoff - this kind of fix seems like a valid thing for us to do as well. I'm unfamiliar with `pillow-simd` though, would this also need to change the import statements, or is the fix to setup.py sufficient? 

Also if pillow-simd is truly a drop-in replacement with the same functionality, I don't think that it is important for visdom to know which you are using.

0phoff:
That is the only thing required.  
Just like Pillow and PIL, pillow-simd also uses the PIL namespace.
Thus it does indeed not matter which one someone is using.

Pillow-simd is actually a fork of Pillow and they changed some of the underlying image functions.
You can find more information in their [repository](https://github.com/uploadcare/pillow-simd).

JackUrb:
Alright in that case I'm happy to take the pytorch/vision solution. Do you want to submit the PR?

JackUrb:
Implemented in #368

PR #368, opened 2018-06-07
Add pillow-simd as requirement if it is installed

0phoff:
As discussed in #367, this PR allows visdom to work with Pillow-simd if the end user has it installed.  
What this PR does, is that upon installation it checks if pillow-simd is installed and adds it to the requirements if it is. Otherwise, regular pillow is added.  
Pillow-simd is a drop-in replacement that uses the same namespace, so no further changes are required.

__Disclaimer:__  
I did not figure this one out on my own, but saw this solution in [pytorch/vision](https://github.com/pytorch/vision/pull/522).
This solution might have some problems I am not aware of, as I am no guru when working with setuptools. One thing I do know, is that this will cause a problem when creating pip wheels, so that should be taken into consideration. _(I dont think that this is a concern for this project)_

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,14 @@
 import os
 from setuptools import setup
+from pkg_resources import get_distribution, DistributionNotFound
+
+
+def get_dist(pkgname):
+    try:
+        return get_distribution(pkgname)
+    except DistributionNotFound:
+        return None
+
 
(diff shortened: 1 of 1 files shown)


JackUrb:
Seems good to me! Thanks @0phoff for the fix and the detailed PR explanation.

PR #370, opened 2018-06-10
Support Tensor that requires grad

LinChen-Cherry:
Add support for visualizations of Tensor that requires grad.

Note that Tensor.detach() is called inside this support.


JackUrb:
Hi @LinChen-Cherry - Thanks for the fix!

JackUrb:
Confirmed that detaching a detached tensor is a no-op. Merging

PR #373, opened 2018-06-11
Creates default env if it doesn't exist with Visdom(env=...)

### LINKED ISSUE #365: Cannot instantiate visdom with specified env name
When run the following code, an error is reported by the terminal running the visdom.server
vis = Visdom(env='test')

However, if the env name is not given, it is OK. The specific code is:
vis = Visdom()

I have tried to fix this error by install simplejson, but it does not work.

ERROR:tornado.application:Uncaught exception POST /win_exists (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:8097', method='POST', uri='/win_exists', version='HTTP/1.1', remote_ip='127.0.0.1')
  File "/usr/local/lib/python3.5/dist-packages/tornado/web.py", line 1541, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/usr/local/lib/python3.5/dist-packages/visdom/server.py", line 481, in post
    self.wrap_func(self, args)
  File "/usr/local/lib/python3.5/dist-packages/visdom/server.py", line 472, in wrap_func
    if args['win'] in handler.state[eid]['jsons']:
KeyError: 'test'
ERROR:root:ERROR: 500: {'exc_info': (<class 'KeyError'>, KeyError('test',), <traceback object at 0x7fca48f23608>)}

JackUrb:
Right now the env is only created once the visdom client actually posts something, which is somewhat unexpected behavior. Now if an env with `env_name` doesn't exist when `visdom.Visdom(env='env_name')` is called, an env is created for `env_name`.

Fixes #365

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,9 +272,11 @@ vis._send({'data': [trace], 'layout': layout, 'win': 'mywin'})
 
 ### Others
 - [`vis.close`](#visclose)    : close a window by id
+- [`vis.delete_env`](#visdelete_env) : delete an environment by env_id
 - [`vis.win_exists`](#viswin_exists) : check if a window already exists by id
 - [`vis.get_window_data`](#visget_window_data): get current data for a window
 - [`vis.check_connection`](#vischeck_connection): check if the server is connected
+
 
 ## Details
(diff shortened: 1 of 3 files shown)


JackUrb:
Forgot the readme edit in #372  so I added it here.

PR #374, opened 2018-06-11
Fixes modal default values

### LINKED ISSUE #292: Env delete button initially disabled when modal opened.
image

JackUrb:
By providing default values for the dropdowns in the modals, we prevent issues where the React component was in a degenerate state, which was causing issue #292. These values are pulled from the current state, so we are sure that the default values we provide will always be valid to render.

Fixes #292.


js/main.js
```diff
@@ -739,15 +747,23 @@ class App extends React.Component {
   }
 
   openEnvModal() {
-    this.setState({showEnvModal: true, saveText: this.state.envID});
+    this.setState({
+      showEnvModal: true,
+      saveText: this.state.envID,
+      modifyID: this.state.envList[0]
```
lvdmaaten: Add comma at end of line for blame.

PR #375, opened 2018-06-11
Local logging to file, replay from logs

### LINKED ISSUE #321: [Feature Request] Visdom python client save locally as well as to the server
Raised on reddit [here.](https://www.reddit.com/r/MachineLearning/comments/8ej2j4/d_facebook_visdom_vs_google_tensorboard/dxvnslq/)

Right now if the visdom remote server goes down, it's not really possible to continue logging to it, and logs sent during that time are simply lost. Saving logs locally in a format that can be pushed to the server later would alleviate this problem.

JackUrb:
Implements #321. Basically visdom would simply dispose of logs that failed, and if a user closed panes on the server there was no recourse to getting them back. This introduces a way to store the logs on the local visdom client's machine as well as the remote visdom server, ensuring the data remains safe. It also allows replaying of the data.

Basic implementation centers around logging the endpoint and json of every plotting event to a file, and then that file can later be replayed with that info. One gotcha is that all plots have to know what the window id was that they were plotted with in order to make sure that recorded updates to those plots will properly update the correct plot. This does mean that logs cannot be replayed multiple times in the same environment.

Demo:
visdomreplaylog

This PR also includes documentation for some of the Visdom object params

Closes #321

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,19 @@ th example/demo2.lua
 ## API
 For a quick introduction into the capabilities of `visdom`, have a look at the `example` directory, or read the details below.
 
+### Visdom Arguments (Python only)
+The python visdom client takes a few options:
+- `server`: the hostname of your visdom server (default: `'http://localhost'`)
+- `port`: the port for your visdom server (default: `8097`)
+- `env`: Default environment to plot to when no `env` is provided (default: `main`)
+- `raise_exceptions`: Raise exceptions upon failure rather than printing them (default: `True` (soon))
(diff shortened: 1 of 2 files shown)


jramapuram:
Sorry to regurgitate this thread, but is it possible to **both** log to file and send to the remote server? Also are disconnects / server errors handled gracefully? i.e. try/catch blocks? Ideally this feature would be useful in scenarios where the node is still able to continue training the model.

JackUrb:
@jramapuram at this point it's working to save locally and send to the remote server at the same time as long as you specify a log file, and I'm pretty sure errors will be handled without being raised, though you can guarantee this with `visdom.Visdom(raise_exceptions=False)`.

jramapuram:
Great! Thanks for that!

Issue #376, opened 2018-06-12
[Feature Request] Marker style for individual lines (including dashes)


lozino:
Hi!

Is there a way to plot different kinds of lines on the same plot?

For example, let's say I want to visualize the mean and the variance of different variables over time, and I would like to plot something like:

mean(x1)  --> blue continuous line
var(x1)      --> blue dashed line
mean(x2)  --> red continuous line
var(x2)      --> red dashed line

and so on...

Is there a simple way to do this?

Thank you!

JackUrb:
Unfortunately this isn't possible in any easy way just yet, as the line dash/dot style would be set [here](https://github.com/facebookresearch/visdom/blob/31f76766c09c7f2d381a09d624b8910cb866cbd7/py/visdom/__init__.py#L972) according to the [plotly format](https://plot.ly/javascript/line-charts/#line-dash) and we haven't set that up. It would make sense to include this as part of enhancing the entire markersize concept (which I've wanted to do since #149).

You can however package your own line data following the plotly guide and send it using `_send` as described [here](https://github.com/facebookresearch/visdom#generic-plots) and shown in the demo [here](https://github.com/facebookresearch/visdom/blob/31f76766c09c7f2d381a09d624b8910cb866cbd7/example/demo.py#L416).

lozino:
That works thanks! What if I wanted to append new data to the existing plot?

JackUrb:
You should be able to use the same `_send` operator but with `endpoint='update'` and sending the same `update='append'` parameter that visdom does in `visdom.scatter`. Note that I've never tried this, but it should work in theory.

On Jun 13, 2018, at 4:12 AM, lozino <notifications@github.com<mailto:notifications@github.com>> wrote:


That works thanks! What if I wanted to append new data to the existing plot?

JackUrb:
Ah yes to update using `endpoing='update'` you need to specify the name of the trace you are updating, as in the way that [scatter handles updates](https://github.com/facebookresearch/visdom/blob/31f76766c09c7f2d381a09d624b8910cb866cbd7/py/visdom/__init__.py#L1003). It needs this to know what specifically you're targeting

lozino:
I have tried doing that with no positive result. Here's the snippet i am using for testing:

  ```
for i in range(10):
    trace = dict(x=[i], y=[i], mode="markers+lines", type='custom',
                 marker={'color': 'red', 'symbol': 104, 'size': "10"}, name='1st Trace')
[...]

PR #377, opened 2018-06-13
Allowing updates to custom plots

JackUrb:
Before we didn't allow updates through the update endpoint to custom plots, which doesn't make sense really (as they can be whatever users want). This helps temporarily alleviate #376 until we improve the marker style format.

### DIFF
--- a/py/visdom/server.py
+++ b/py/visdom/server.py
@@ -602,7 +602,7 @@ class UpdateHandler(BaseHandler):
 
         if not (p['type'] == 'text' or
                 p['content']['data'][0]['type'] in ['scatter', 'custom']):
-            handler.write('win is not scatter or text, was {}'.format(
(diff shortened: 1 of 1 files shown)


py/visdom/server.py
```diff
@@ -601,8 +601,9 @@ def wrap_func(handler, args):
         p = handler.state[eid]['jsons'][args['win']]
 
         if not (p['type'] == 'text' or
-                p['content']['data'][0]['type'] == 'scatter'):
-            handler.write('win is not scatter or text')
+                p['content']['data'][0]['type'] in ['scatter', 'custom']):
+            handler.write('win is not scatter or text, was {}'.format(
```
lvdmaaten: `'win is not scatter, text, or custom; was {}'`

PR #379, opened 2018-06-13
[Docs] Firewall update

JackUrb:
Slight improvement to the docs based on the comment here: https://github.com/facebookresearch/visdom/issues/288#issuecomment-397096106

### DIFF
--- a/py/visdom/server.py
+++ b/py/visdom/server.py
@@ -601,9 +601,8 @@ class UpdateHandler(BaseHandler):
         p = handler.state[eid]['jsons'][args['win']]
 
         if not (p['type'] == 'text' or
-                p['content']['data'][0]['type'] in ['scatter', 'custom']):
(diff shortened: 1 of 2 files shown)


py/visdom/server.py
```diff
@@ -601,8 +601,9 @@ def wrap_func(handler, args):
         p = handler.state[eid]['jsons'][args['win']]
 
         if not (p['type'] == 'text' or
-                p['content']['data'][0]['type'] == 'scatter'):
-            handler.write('win is not scatter or text')
+                p['content']['data'][0]['type'] in ['scatter', 'custom']):
```
lvdmaaten: Note: This change is also in https://github.com/facebookresearch/visdom/pull/377#pullrequestreview-129403199

You may run into merge conflicts when landing these.

PR #384, opened 2018-06-18
Create Issue Templates

JackUrb:
We don't have any issue templates, and sometimes that adds friction in the first few rounds of trying to figure out what's wrong with somebody's setup. This should clarify things


.github/ISSUE_TEMPLATE/-bug-report-.md
```diff
@@ -0,0 +1,30 @@
...
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+Give a clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Client logs:**
```
lvdmaaten: "In Chrome, javascript logs can be found via View -> Developer -> JavaScript Console."

.github/ISSUE_TEMPLATE/-bug-report-.md
```diff
@@ -0,0 +1,30 @@
...
+For issues that make it to the point of reaching the frontend in a browser, please include the javascript logs from that browser. 
+
+**Server logs:**
+For any issue that involved launching the visdom server, please include the server logs:
```
lvdmaaten: You need to see server logs not just for server-launch problems, but also for certain runtime problems. 

It may also be good to remark here that the server doesn't currently log into a particular file (maybe it should?) but simply to stdout.

PR #387, opened 2018-06-18
Add 'visdom' script for starting server.

### LINKED ISSUE #382: Thoughts on binary launcher?
Hi,
I'm wondering if it would be good idea if `pip install visdom` installed a binary that when executed is equivalent to `python -m visdom.server`. ie, a user types

```
> visdom
```

instead of 

```
> python -m visdom.server
````

Seems nice since the former is shorter and easier to remember, and more in line with how other servers are typically started (eg, `tensorboard`).

malmaud:
Closes #382.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,13 +166,15 @@ luarocks make
 
 ## Usage
 
-Start the server (probably in a  `screen` or `tmux`) :
+Start the server (probably in a  `screen` or `tmux`) from the command line:
 
 ```bash
-visdom
+> visdom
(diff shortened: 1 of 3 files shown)


malmaud:
Something like this? Feel free to push edits to the PR or merge and edit if there's a particular wording you favor.

JackUrb:
Perfect - thanks for the neat enhancement!

malmaud:
Thanks for reviewing!

PR #390, opened 2018-06-19
Clarifying colormap for plotly plots

### LINKED ISSUE #380: scatter does not apply colormap
https://github.com/facebookresearch/visdom/blob/d1c95845c784b1cd1a99bc64afd68efcaa178efc/py/visdom/__init__.py#L926

example:
`vis.scatter(X=torch.randn(10,2).numpy(), opts=dict(colormap='Hot')`

JackUrb:
closes #380.

Plotly never got around to drawing color schemes from colormaps for line and scatter plots, so these arguments don't actually do anything for these plot types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,7 +408,6 @@ reflected in the colors of the markers.
 
 The following `opts` are supported:
 
-- `opts.colormap`    : colormap (`string`; default = `'Viridis'`)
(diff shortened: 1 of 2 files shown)


JackUrb:
It's also possible plot.ly deprecated them at some point. I saw some reference to it elsewhere but not in the code/api

PR #391, opened 2018-06-19
Allows 'append' on non-existing plot for option re-use

### LINKED ISSUE #100: Feature Request: Plot a BlankBoard.
Can we just plot a blank board, and alway use updateTrace to add new points or new lines?

JackUrb:
Closes #100.

People want a blankboard so that they can use the same command in every call of an evaluation loop to plot rather than needing to instantiate the graph before they can update. Existing workarounds are doing a regular plot call on the first iteration and appends for all others, or initializing the plot with a garbage value (like 0,0 on a line) and then appending from that point. Having an option that can create or append solves the issue.

This leverages `win_exists` to determine whether a particular call should be a create or append when passed the `'createappend'` option. It then does the correct action. Tested for both line and scatter, and existing behavior is maintained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,7 +404,11 @@ scatter plot. An optional `N` tensor `Y` containing discrete labels that
 range between `1` and `K` can be specified as well -- the labels will be
 reflected in the colors of the markers.
 
-`update` can be used to efficiently update the data of an existing plot. Use 'append' to append data, 'replace' to use new data, or 'remove' to remove the trace specified by `name`. If updating a single trace, use `name` to specify the name of the trace to be updated. Update data that is all NaN is ignored (can be used for masking update).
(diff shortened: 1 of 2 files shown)


malmaud:
You can't currently pass in `win=None` and `update='append'`, which is where this is useful:

From __init__.py:913:

```python
        elif update is not None:
            assert win is not None
```

malmaud:
Sounds fine by me. 

I suppose database APIs usually make a strong distinction between updating and upserting so users are immediately alerted to logic errors in their code if they try to update a non-existing record, but in the context of visdom the increase in API complexity of the distinction might not be warranted.

JackUrb:
Will make the change to append creating by default

Issue #389, opened 2018-06-19
Switch from `setattr` to decorators?


malmaud:
I notice in `__init__.py`, you use `setattr` metaprogramming to make `Visdom` methods work on both Torch tensors and numpy arrays. This seems to be confusing static analysis with Jedi, causing autocomplete for `Visdom` methods to fail in various IDEs. 

How would you feel about switching to using a decorator for methods that  take can a tensor or array instead of `setattr`? That would be more transparent to static analysis tools and restore autocomplete. Arguably it's also cleaner, since the annotation that a method works on tensors and arrays would be inline with the method instead of that logic happening in a seperate block. 

I can submit a PR for discussion.

JackUrb:
I've always found our current use of `setattr` to be a rather hacky support, but I haven't been able to take the time to work through an alternate solution. If you have something else in mind that works I'd be more than happy to take a look at it!

malmaud:
Closed by https://github.com/facebookresearch/visdom/commit/2305c7ceca35e19a4d8411556ed2f8d535a37fe8

PR #392, opened 2018-06-19
Switch from 'setattr' to decorator.

malmaud:
First pass at resolving #389.

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -30,6 +30,7 @@ import warnings
 import time
 import errno
 import io
+from functools import wraps
 try:
     import bs4
     BS4_AVAILABLE = True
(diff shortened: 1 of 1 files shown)


JackUrb:
Hi @malmaud - Have you tested this on np arrays and tensors? (And has it helped the autofill issues from before?)

This seems like a really clean solution though, and it's given me a chance to learn about `wraps`. If it's working well I'd be happy to take it in as-is, though I'll need to ensure functools works internally before I can land it.

malmaud:
I have, although it requires PyTorch 0.4 or greater for the `isattribute(a, torch.Tensor)` to work robustly since before that, variables were their own type. Although that reminds me that we also have to check for `torch.nn.Parameter`. 

It does enable autocomplete.

JackUrb:
Hm so I suppose this will be breaking for <0.4. We might want to hold off on this for a bit until more people have migrated or implement some solution for that case.

py/visdom/__init__.py
```diff
@@ -248,25 +249,30 @@ def _assert_opts(opts):
...
-            if type(kwargs[k]).__module__.startswith('torch'):
-                kwargs[k] = kwargs[k].cpu().detach().numpy()
-
-        return fn(*args, **kwargs)
-    return result
-
+def _torch_to_numpy(a):
+    for kind in torch_types:
+        if isinstance(a, kind):
+            return a.cpu().detach().numpy()
```
lvdmaaten: This will fail for PyTorch <0.4 `Variable`s: for those, the tensor is wrapped in `a.data`. To remain backwards compatible, perhaps you can check `hasattr(a, 'data')` to see whether you need to get that field? (Please add a comment stating which bits are for backwards compatibility and a deprecated warning -- that makes it easy for us to remove the <0.4 code later.)

malmaud: The `hasattr` method won't work since 'data' is defined on non-variables in PyTorch 0.3; it will just raise an exception if you try to access it. I tried in my subsequent commit to come up with the most readable version of backwards compatibility and successfully tested it in a virtualenv with PyTorch 0.3.1.

Feel free to push to this PR or edit after merging  if you want to tweak the wording of anything.

Issue #25, opened 2017-03-19
interactive features?


tachim:
Hi, are there any plans to include interactive features like sliders?

ajabri:
Hi @tachim,

Thanks for your post! Can you provide an example of 'sliders'? In general, we're looking forward to plugging in various visualization APIs into visdom.

tachim:
Hi @ajabri, I'm referring to visualizations like these: https://nbviewer.jupyter.org/gist/jackparmer/239c00f721097deb11fd. Not high priority at all, just a thought.

shubhamjain0594:
Hi @ajabri 

Is there any work being done on this. What are possible ideas for the slider and what exactly is it's utility? Issue #114 looks at much broader usage of slider, is that something we are looking. I am interested in trying to implement it out.

lvdmaaten:
We're not actively working on this right now (busy with other things), but we welcome PRs implementing (part of) this functionality!

PR #393, opened 2018-06-19
Image History sliders

### LINKED ISSUE #286: [Feature Request] Add slide bars to show multiple images over time
Hi, now I'm using both visdom and TensorBoard. Is it possible to add slide bars as the attached image below to show multiple images over time? Now, if you display GANs' generated images at each epoch, for example, the window is filled with them.

screenshot-2018-3-3 tensorboard

### LINKED ISSUE #369: opts={'caption':"Windows Title Showed?"} The Caption  NOT show in windows?
Hi, I have several windows showing different images, and I need the `caption` property showing in windows title bar to recognise each other. However,
```python
vis.image(img, win='img', opts={'caption':"Windows Title Showed?"})

```
When i set the opts.caption parameters, the caption dont appear as :
image

Am I wrong?

JackUrb:
Closes #286 by implementation.
Fixes #369 for real.

Adds a way to store image history. The `store_history` option, when provided repeatedly for the same window will allow for appending multiple images to the same pane, accessed using a slider. The most recently sent image will be selected by default, and updates from the python client will reset the selection to the most recent. This doesn't break any existing behavior:
screen shot 2018-06-19 at 5 42 28 pm
This also begins the process for adding 'widgets' to arbitrary panes, allowing for functionality like long requested in #25. (Standardizing these widgets and allowing users to attach them to variables in the plot/layout state will be the last part of implementing #25)

Full implementation flow:
- Adds an option to `image` that attaches a store_history option to the request. Sends the request to the update endpoint if the window already exists. (Note: It'd be nice to make it so that the server doesn't require the `store_history` flag every time an additional image is sent, but it complicates the flow a lot. May revisit down the road)
- Creates flows in the create and update handlers of the server that recognize `image_history` types and append data to the content rather than making it new.
- Adds logic to `ImagePane` that allow them to track the state of the slider, if it exists
- Adds logic to `Panes` that allows specialized panes to render additional content 'widgets' at the bottom of the pane.

Gotchas:
- ~~This PR doesn't have a demo, will include in a follow-up~~ Added in #593 
- This PR doesn't have feedback sent back to the python client on slider updates, will implement in a follow-up
- This PR doesn't have a method for the python client to select a slider value, will implement in a follow-up alongside a demo that can sync the contents of multiple image panes to one slider!
- ~~This PR brings up the issue of the size of content being sent during window updates to a head, as each additional image added to a history forces a send of the entire history every single time. Furthermore each update to any window in the same environment will force this same send.~~ This was implemented in #395 
- This PR doesn't properly handle margins, even though I can calculate how tall I want them to be. Need to review some CSS magic.
screen shot 2018-06-19 at 5 50 16 pm


talesa:
is there any timeline for accepting this PR?
It would be great to have it!

JackUrb:
Hi @talesa - sorry for the delays here, but I haven't had the time to finish #393 which is a prerequisite for this otherwise the visdom server becomes unstable while this feature is inactive for long periods of time. Originally had some community involvement on that one but the last leg of the task never finished because we hadn't found a great `json-diff` library that had a shared format between python and JavaScript.

tarrencev:
Any plans to merge this? Would be a cool functionality!

JackUrb:
Hi @tarrencev - I haven't gotten to reinvestigate this since landing #395, but I hope to get to it soon! It should mostly be a matter of resolving merge conflicts.

JackUrb:
@pringshia I've rerun tests and ensured the new demo is working. 

This has been a long-standing proposed change that I'm sure a bundle of people would like to see merged since I introduced it mid last year, but I never got to merge it. If the code makes sense to you and the widget implementation appears valid/extensible enough, I think we should pull it in.

js/ImagePane.js
```diff
@@ -148,13 +155,54 @@ class ImagePane extends React.Component {
...
+      selected: evt.target.value,
+    });
+  };
+
   render() {
     let content = this.props.content;
+    let type = this.props.type;
+    let widgets = [];
+
+    if (type == 'image_history') {
```
pringshia: In this case I don't think it's likely an issue, but I usually prefer to use triple equals (`===`) for equality checks to avoid accidental implicit typing

py/visdom/static/css/style.css
```diff
@@ -129,6 +129,13 @@ button {
   overflow: hidden;
 }
 
+.widgets {
+  position: absolute;
+  bottom: 0;
```
pringshia, line 134: Seems that by being absolutely positioned, there is the possibility that the widgets panel can be overlaid on top of the content. Could this cause any issues where the underlying content is not "interactive-able"? Specifically in this case, could the widgets pane cover up the bottom part of the image and clip the bottom content?

Instead of making `widgets` a first class citizen and exposing it through the `Pane` API what are the trade-offs of having instead the higher abstraction (in this case `ImagePane`) define it as part of the children prop instead, similar to the rest of the content? Are there any benefits in having widgets be differentiated and elevated as a specific additional prop?

JackUrb: This debate is exactly the reason I didn't want to merge these changes without having a discussion with someone haha. One advantage I imagined down the line is as people start wanting more interactive components, they'd have the ability to wire them directly into the widgets of relevant plots/panes. 

Of course the position absolute on the bottom at the moment can cover up content, and I couldn't come up with a different solution to be able to append additional content greater than the specified pane height.

pringshia: Agreed on the idea of extensibility, though I'm still tossing around the execution in my mind. I think for a v0 this is fine, though I'd like to revisit this at some point. The overlay issue isn't also a major blocker here since I see that the image is draggable.

As an aside, I'm noticing out-of-place text (repeated caption plus an errant `<`) underneath the image. Unrelated to this PR:
visdom-image-move

JackUrb: Well *that* explains where the old captions went 😂

psmaragdis:
Any chance we can add the same functionality for audio as well? It would be nice to compare past and current audio outputs by moving an epochs slider.

PR #396, opened 2018-06-20
Add Python type stub.

malmaud:
First pass at adding type annotations.

Closes https://github.com/facebookresearch/visdom/issues/388.


JackUrb:
Seems solid to me - did you get a chance to test with `pyre`?

malmaud:
Not yet, I'll give that a shot.

malmaud:
It does work with pyre in my testing, with the caveat that pyre needs to be explicitly told the directory where the stub lives with the `search_path` option. That won't be necessary once https://github.com/facebook/pyre-check/issues/70 is resolved.

malmaud:
I rebased. I don't think any other changes are needed since #392 didn't change the type information of any `Visdom` arguments, as far as users are concerned.

JackUrb:
Ah fair enough - I recall a time where some of this overlapped with #392 but I suppose that went away with the separate stub file. Thanks!

PR #403, opened 2018-06-25
Fix 'plotlyplot' for numpy arrays.

malmaud:
I realize that the `plotlyplot` function will fail if the Plotly figure's attributes are numpy arrays instead of plain lists, since the numpy arrays cannot be serialized to JSON in `_send`. 

I thought the easiest solution is to make use of the Plotly Python client's `PlotlyJSONEncoder`, which properly deals with that as well as proper handling of NaNs and Infs. I load the JSON back as a Python dict so `_send` can add other attributes to it before JSON-serializing the entire message. 

With this, `plotly.figure_factory.create_distplot` works:

screenshot 2018-06-25 18 05 11

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,8 @@ packages installed to use this option.
 
 This function draws a Plotly `Figure` object. It does not explicitly take options as it assumes you have already explicitly configured the figure's `layout`. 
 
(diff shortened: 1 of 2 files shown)


malmaud:
Done.

PR #418, opened 2018-07-12
Use wss when serving over https.

### LINKED ISSUE #415: Support WSS
When serving visdom over HTTPS (such as with a reverse proxy), the websocket connection between the client and server continues to be insecure since the `ws` instead of `wss` protocol is used. Chrome will warn you and require you to opt-in to enable the insecure websocket connection.

A command-line option to the visdom server that causes the served javascript to replace `ws://` with `wss://`. The Python client could use `wss` instead of `ws` if it sees that the Visdom host starts with `https` instead of `http`.

malmaud:
Closes #415.


JackUrb:
Thanks for the help here! Quick question though - what happened to the `static/main.js` file here? 7 lines of change in `main.js` doesn't seem to match 42k lines in the output.

malmaud:
Ah ya, I just built it with `npm run dev` when I guess I should have used `npm run build` at the end. But as a generated file, is it a good idea for `static/main.js` be in version control at all?

JackUrb:
I'd certainly prefer the static main file to be generated at run, but I haven't found a good way to do that internally at fb yet.

lvdmaaten:
Also, it would require every visdom user to install npm + a bunch of JS dependencies. This seems undesired as very few of these users are likely to ever touch the Javascript code.

malmaud:
Well, I was thinking you would build it when uploading a new  release, as any Python package that has binary artifacts would.

malmaud:
At any rate, fixed now.

PR #498, opened 2018-10-25
Allow scatter plots with more legend items than number of classes

### LINKED ISSUE #480: Allow scatter plots with more legend items than number of classes
This line forces the number of items in the legend to be the same as the number of classes.

https://github.com/facebookresearch/visdom/blob/master/py/visdom/__init__.py#L1140

In my case, not all the classes can be sampled in a particular batch. `len(opts['legend'])` becomes greater than `K`.

Allow number of legend items *at least* the number of classes 

I can manually remove legend items not used in this particular batch, but this is cumbersome.

vtantia:
## Description
Change two assert statements in the scatter function from == to >=. This is done in order to -
- allow more legend items than number of classes
- allow the label 1 to be missing and assert that the minimum label is not less than 1

## Motivation and Context
Closes #480 
Also allows scatter function calls with label 1 missing. This could happen in a case in which the label 1 is not included in the scatter function call

## How Has This Been Tested?
This has been tested by making 3d scatter plots by modifying demo.py. The 3d scatter demo in demo.py contains 2 legend items. I have increased it to 3 legend items and then made 3 plots -
- with the same labels as demo.py, ie, [1, 2]
- with labels [2, 3]
- with labels [1, 2]
Screenshot is attached below for all 3 plots.

In addition to this, I have tested the demo.py file on both my branch and a clean branch and everything is the same.

image

## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -1137,7 +1137,7 @@ class Visdom(object):
         _assert_opts(opts)
 
         if opts.get('legend'):
-            assert type(opts['legend']) == list and len(opts['legend']) >= K, \
+            assert type(opts['legend']) == list and K <= len(opts['legend']), \
                 'largest label should not be greater than size of the ' \
                 'legends table'
 
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -1110,7 +1110,7 @@ class Visdom(object):
             Y = np.ones(X.shape[0], dtype=int)
 
         assert np.equal(np.mod(Y, 1), 0).all(), 'labels should be integers'
-        assert Y.min() == 1, 'labels are assumed to be between 1 and K'
+        assert Y.min() >= 1, 'labels are assumed to be between 1 and K'
 
         K = int(Y.max())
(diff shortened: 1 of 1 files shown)


py/visdom/__init__.py
```diff
@@ -1137,7 +1137,9 @@ def scatter(self, X, Y=None, win=None, env=None, opts=None, update=None,
         _assert_opts(opts)
 
         if opts.get('legend'):
-            assert type(opts['legend']) == list and len(opts['legend']) == K
+            assert type(opts['legend']) == list and len(opts['legend']) >= K, \
```
JackUrb:
```suggestion
            assert type(opts['legend']) == list and K <= len(opts['legend']), \
```
Extremely small style nit: Swapping this order makes the point of this check more clear as in the corrected form it reads "K (the largest label) should not be larger than the legend items", which matches the assertion message more directly than "there should be more legend options than the largest label", which is also correct but matches the message less.

vtantia: Thanks for the guidance! Accepted the suggested change

PR #526, opened 2018-12-14
Make Image Pane scrolling browser independent

dzimmerer:
Instead of the browser-dependent scroll wheel value, only use the sign value, and allow to side scroll image with Alt+Shift.

## Description
Use only the sign of the scrolling value/speed to always get a fixed scrolling offset.
Also added, if Alt+Shift is pressed to allow scrolling image in X direction instead of Y.

## Motivation and Context
https://stackoverflow.com/questions/5527601/normalizing-mousewheel-speed-across-browsers
Mouse wheel speed is different for different browsers, leading to unusable zooming and scrolling values in different browsers.

## How Has This Been Tested?
Tested on Chrome + Firefox (on ubuntu).

## Types of changes

## Checklist:


js/ImagePane.js
```diff
@@ -61,23 +61,35 @@ class ImagePane extends React.Component {
   }
 
   handleZoom = (ev) => {
+
```
JackUrb: nit: remove

js/ImagePane.js
```diff
@@ -61,23 +61,35 @@ class ImagePane extends React.Component {
...
+      else {
+        this.setState({tx: this.state.tx + scrollDirectionX * direction*50});
+        this.setState({ty: this.state.ty + scrollDirectionY * direction*50});
+      }
       ev.stopPropagation();
       ev.preventDefault();
     } else if (ev.ctrlKey) {
       // get the x and y offset of the pane
       var rect = this._paneRef._windowRef.children[1].getBoundingClientRect();
+      // Get browser independendgit scaling factor
```
JackUrb: nit: spelling

JackUrb:
Hi @dzimmerer - thanks for the fix, it looks great! One question I have about the shift implementation is that Shift->scroll in most browsers is registered as horizontal scrolling already, so isn't the special case redundant?

dzimmerer:
I thought so as well, but at least in the cases I tried the horizontal scrolling worked on the whole window/grid but not on the image pane.

PR #544, opened 2019-01-12
Remove non-ASCII characters in README.md

bouthilx:
## Description

Simply replace characters ‘ by \' in `README.md` and encode the `README.md` file in `setup.py` to avoid further problems if non-ASCII characters were to be added again.

## Motivation and Context

Reading a file with non-ascii characters will raise a UnicodeDecodeError if the system locale is not set to UTF-8. Note that the characters in question are not coherent with the rest of the documentation and should be ` rather than ‘.

## How Has This Been Tested?

Tested in a container where locale is not UTF-8. 

## Types of changes

## Checklist:


JackUrb:
Thanks for catching this @bouthilx - I'll pull this in and update the distribution package!

Issue #395, opened 2018-06-19
[Core Improvement] Reign in the amount of content sent to the frontend client when things are updated.


JackUrb:
Noted in #393, a lot of data gets sent to the client as an environment grows. As seen in #386, this is not a sustainable practice. Moving forward we'll need to be able to send update packets from the server rather than the entire contents of an environment's json every time a window in that json gets updated. The client should only request the complete contents when the environment changes or if an inconsistency is detected (can be done using a hash of the stored data).

xssChauhan:
@JackUrb I would like to help with this

JackUrb:
Hi @xssChauhan - this is a fairly involved change that I've already started work for, but one segment I haven't touched is the last part (the hashing step). This involves sending a request from the web client to a new endpoint with an env id and window id and receiving in return a hash/checksum of the layout and data of that window. You can then also implement the same hash/checksum function on the web side. Does this seem like something you'd want to work on?

xssChauhan:
@JackUrb This is certainly interesting. Would be glad to work on it. Where do i begin?

JackUrb:
Alright so you'll start in the server file by introducing a new endpoint. This should be something like the exists handler: https://github.com/facebookresearch/visdom/blob/ec91e738e100a9cbf946c332c7783ffd12c644b5/py/visdom/server.py#L155
It will take a window and eid in the same way:
https://github.com/facebookresearch/visdom/blob/ec91e738e100a9cbf946c332c7783ffd12c644b5/py/visdom/server.py#L506
Then it should perform a hash operation on the json stored in `handler.state[eid]['jsons']`. Preferably this will be an MD5 of the json string. It should then return that hash by using `handler.write()`.

xssChauhan:
@JackUrb Understood. Is there a branch for this already? Or do i make one  from master only?

JackUrb:
You can start a new branch for this - it's somewhat independent of the rest of the work

xssChauhan:
Cool. Getting started now. Thank you for the starting pointers.

xssChauhan:
@JackUrb Here's my [implementation](https://github.com/xssChauhan/visdom/commit/44a0dc835f67c81ac554c0a25ecf8b2bd29ff5f5) for the hash handler. Will you please take a look at it and tell me your review?

JackUrb:
@xssChauhan - feel free to open it as a pull request and I can better comment there!

xssChauhan:
[...]

PR #552, opened 2019-01-17
Update network io

suvrat96:
Improves network IO by sending diff packets instead of the entire window data when a window is updated.
## Description
Calculates the diff packet when a window is updated via jsondiff. 
(Can be further optimized in the future via Delta Format)
Updated hashing function to use stable ordering stringify version taking into account JS representation of numbers for consistent hashing of windows. 
Modified main.js to handle window update diff packets.
Modified state to maintain two copies of panes in order to maintain consistency (The original copy of panes is contaminated by plotly which prevents consistency checking.)

## Motivation and Context

#395 

## How Has This Been Tested?
python example/demo.py
Tested on firefox and chrome browsers. 
Ensured that hashes match and update happens via the diff packet where appropriate via console logs. 

## Types of changes

## Checklist:


py/visdom/server.py
```diff
@@ -573,6 +575,34 @@ def post(self):
         self.wrap_func(self, args)
 
 
+def order_by_key(kv):
+    key, val = kv
+    return key
+
+
+# Based on json-stable-strinfy-python from @haochi with some usecase modifications
```
JackUrb: Nit: `json-stable-stringify-python`

py/visdom/server.py
```diff
@@ -582,6 +612,14 @@ def initialize(self, app):
         self.env_path = app.env_path
         self.login_enabled = app.login_enabled
 
+    @staticmethod
+    def update_packet(p, args):
+        old_p = copy.deepcopy(p)
+        p = UpdateHandler.update(p, args)
+        p['contentID'] = get_rand_id()
+        patch = jsonpatch.make_patch(old_p, p)
```
JackUrb, line 621: I'd add a comment here that this is creating the patch by doing a diff, and that it could be a good candidate for optimization if down the road we need it.
`# make_patch isn't high performance. If bottlenecked we should build the patch ourselves.`
Or something like that.

py/visdom/server.py
```diff
@@ -665,10 +703,19 @@ def wrap_func(handler, args):
                 p['content']['data'][0]['type']))
             return
 
-        p = UpdateHandler.update(p, args)
-
-        p['contentID'] = get_rand_id()
-        broadcast(handler, p, eid)
+        p, diff_packet = UpdateHandler.update_packet(p, args)
+        if len(stringify(p)) <= len(stringify(diff_packet)):
```
JackUrb, line 709: nit: simple comment to describe this block, like `# send the smaller of the patch and the updated pane`

js/main.js
```diff
@@ -279,6 +306,14 @@ class App extends React.Component {
           this.addPaneBatched(cmd);
         }
         break;
+      case 'window_update':
+        if (this.state.envIDs.length > 1 && cmd.has_compare !== true) {
+          this.postForEnv(this.state.envIDs);
+        } else {
+          let numTries = 3;
+          setTimeout(this._checkWindow, 0, cmd, numTries);
```
JackUrb, line 316: Nit: `# Check to see if the window exists before trying to update`

js/main.js
```diff
@@ -259,6 +263,29 @@ class App extends React.Component {
     this._socket = socket;
   }
 
+  _checkWindow = (cmd, numTries) => {
+    if (cmd.win in this.state.consistent_pane_copy) {
+      let windowContent = this.state.consistent_pane_copy[cmd.win];
```
JackUrb, line 269: Nit: `# Apply patch and check hash. Re-fetch if final window doesn't match hash`

Issue #563, opened 2019-02-19
Visdom crashes when given window title is None


ohwi:
**Bug Description**

OS: Ubuntu 16.04
Browser: Chrome / Safari
Visdom version: 0.1.8.8

If the "title" opts passed to a visdom audio is None, the client crashes without an error message and displays only the background.

**Reproduction Steps**

```{.python}
v = Visdom()

def visdom_audio(audio, env="test", disc=None):
	v.audio(
		audio,
		env=env,
		opts=dict(
			sample_frequency=args.sample_rate,
			title=disc,
		)
	)

visdom_audio(path)
```

**Client logs:**

> main.js?v=4316a77ded15b92b6e884b60b72f6513:11 Warning: Each child in an array or iterator should have a unique "key" prop.
> 
> Check the top-level render call using <ul>. See https://fb.me/react-warning-keys for more information.
>     in li
> a @ main.js?v=4316a77ded15b92b6e884b60b72f6513:11
> :8097/extensions/MathMenu.js?V=2.7.1:2 Uncaught SyntaxError: Unexpected number
> :8097/extensions/MathZoom.js?V=2.7.1:2 Uncaught SyntaxError: Unexpected number
> react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:162 TypeError: Cannot read property 'match' of null
>     at main.js?v=4316a77ded15b92b6e884b60b72f6513:35
>     at Array.map (<anonymous>)
>     at t.value (main.js?v=4316a77ded15b92b6e884b60b72f6513:35)
>     at h (react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:130)
>     at beginWork (react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:134)
>     at d (react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:158)
>     at f (react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:159)
>     at g (react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:159)
>     at t (react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:167)
>     at x (react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:166)
> h @ react-dom.min.js?v=950495cc51ccb90612cf0fe0bb44f8f3:162
[...]

PR #566, opened 2019-02-21
Show errors when pane rendering fails

pringshia:
Issues like #563 cause the entire app to "crash" without displaying much visual feedback to users. This is a step towards making error cases more informative, apparent, and graceful.

## Description
Errors will be caught and displayed within a TextPane:
screen shot 2019-02-21 at 1 27 35 pm

## Motivation and Context
- Pane errors do not cause the entire app to "crash" to the blue screen, so other panes are still functional
- Errors messages are displayed in the UI - makes reporting bug reports easier as well as have it be more apparent and obvious that something went wrong.

## How Has This Been Tested?
- I locally reversed #561 to generate an error and observed that it was caught and logged appropriately without bringing down the whole app.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the code style of this project.
- [X] For JavaScript changes, I have re-generated the minified JavaScript code.


pringshia:
@JackUrb I wanted to investigate using React's `ErrorBoundary`s to handle this sort of situation but I think that may be more suited as part of a larger refactor later on as we component-ize more of this codebase. Until then this should be pretty functional. I'll rebase.

pringshia:
Should be fixed now, not sure why Github was attempting to merge the two files. Perhaps we should also considering making this `main.js` file a git [binary file](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes) since it's a compiled build artifact that will never need an intricate merge resolution. Making it binary will only let you merge with a mine-or-theirs resolution.

If you are ok with it, I can add that in as part of this PR.

JackUrb:
Sounds good to me!

PR #597, opened 2019-04-05
Error when `name` is used with multiple lines/labels

SsnL:
## Description
Error when `name` is used with multiple lines/labels

## Motivation and Context
https://github.com/facebookresearch/visdom/issues/596

## How Has This Been Tested?
Honestly speaking, I haven't tested, but this seems simple enough. LMK if I need to.

## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -1076,7 +1076,7 @@ class Visdom(object):
         This function plays a video. It takes as input the filename of the video
         `videofile` or a `LxHxWxC` or `LxCxHxW`-sized `tensor` containing all the frames of
         the video as input, as specified in `dim`. The color channels must be in BGR order.
-        
+
         The function does not support any plot-specific `opts`. The following video `opts` are supported:
 
         - `opts.fps`: FPS for the video (`integer` > 0; default = 25)
@@ -1205,8 +1205,6 @@ class Visdom(object):
             if update == 'append':
                 if win is None or not self.win_exists(win, env):
                     update = None
-                else:
-                    update = 'append'
 
             # case when X is 1 dimensional and corresponding values on y-axis
             # are passed in parameter Y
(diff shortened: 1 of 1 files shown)


py/visdom/__init__.py
```diff
@@ -1322,6 +1320,10 @@ def scatter(self, X, Y=None, win=None, env=None, opts=None, update=None,
 
                 data.append(_scrub_dict(_data))
 
+        if len(data) > 1:
```
JackUrb: Technically this could actually be determined the moment we set `K`, as `K` is the number of lines.

ssnl: This is a bit awkward, because `K` is computed as `int(Y.max())`. So technically it is the max label index. However, we also seemed to have assumed that it is the number of labels/lines at line 1284.

So it seems to me that the correct fix is to: 

1.  Use `np.unique` to compute the set of labels.
2. Change the `for`-loop above to iterate over the set. 
3. Change conditions in line 1284 and this line to be that the set has size 1.

This is a bigger change. Let me know how you think!

JackUrb: Ah you are correct, for the purpose of `name` we used K for the name implementation above, but that isn't actually the correct thing to be checking. I like your proposed solution!

ssnl: I updated the PR in new commit :)

JackUrb:
Thanks @SsnL this looks great! All I'd want to ensure is that `examples/demo.py` still runs. Other than that I suppose we still need a CLA refresh to pull the code in. Once that goes through I'll land this.

SsnL:
Thanks Jack. `demo.py` runs for me, although I can't get the interactive text window working. Does that sound related to you?

I have sent `cla@fb.com` an email, and am currently waiting for them to reply.

Issue #323, opened 2018-04-25
views for multiple envs


0phoff:
I use multiple environments for different trials and modifications of my model trainings.
Usually I run in the 'main' environment, fork it when it is complete and start a new training in the 'main' environment. This way, I only define views in the main environment and once I fork the environment, the currently selected view becomes the 'only' view for the fork.

While this setup works quite fine, I do feel the current way of handling views is quite limiting:
  - If I accidently drag one of the panels, distorting the view in one of my environments, I lost that view on the data. As these forked environments do not have the same saved views, I have to recreate it.
  - If you plot directly in different environments, but want to have the same views for different envs, the only possible way as of now is to recreate your view in each environment.

---

I feel there are 2 ways you could go about fixing this:
  - Implement something like global views, that work for any environment.
    While this might seem weird at first, because not every environment has the same panels, it does have the merit of being useful when comparing 2 environments. The global views could namely be created when comparing `env A` with `env B`, saved, and then reused at a later stage when comparing `env C` with `env D`.
  - Implement a way to copy over views from one environment to another. This is probably easier to implement, but would not help when comparing environments.

JackUrb:
Hi @0phoff - I definitely see where you're coming from with this one. While I see the merit of having global views, I feel it becomes a UX issue either clouding an environment's view list with a bunch of views unrelated to that environment or it requires coming up with a design that allows one to easily mark a view as global to be shared. 

Thus I feel like a copy view function would probably be the better approach. How would you suggest trying to fit the old view into the new environment? Are you using the same window ids? The same chart titles?

To solve the latter issue with the copy approach, I feel it would be simplest to just allow a user to use the union of the views of the environments being compared. Would that improve the workflow?

0phoff:
I use the same window IDs, as it is just the same script running with eg. different parameters or different input data, so copying views would indeed be easy here.

You could check which windows are actually in the new env when copying over a view, but that would be somewhat limiting as you could not copy over the view before all the panels are created. (eg. I run tests every few epochs, so the testing panels are not yet created upon startup)  
I think the best approach would be to just copy over the view with all existing panel IDs, and leave it up to the user to integrate new panels into that view etc.

With union of the views you mean that all the views of the envs being compared can be selected from the dropdown? That would indeed solve most use cases I can see, without being too complicated to implement! 

---

I think this issue can be split up in 2 parts:
  - [ ] Allow a user to copy over views from one env to another
[...]

PR #609, opened 2019-04-19
Expose fork method on python API and sync windows state with backend

ricardofbarros:
## Description
This PR will add the following features: 
- expose a `.fork()` method on the client python API with the same functionality as the UI.
- sync layout state with the backend. We can resize the windows and reload the page that the windows will keep the layout state.

## Motivation and Context
#323

## How Has This Been Tested?
Installing visdom directly from my repo and testing by hand.

## Types of changes

## Checklist:


ricardofbarros:
cc @JackUrb

ricardofbarros:
I just need some guidance on what should the server reply back to the client on the following cases:

- https://github.com/facebookresearch/visdom/pull/609/files#diff-ddbe57b92511a8307bcf5999eb519e6cR860
- https://github.com/facebookresearch/visdom/pull/609/files#diff-ddbe57b92511a8307bcf5999eb519e6cR863

README.md
```diff
@@ -631,6 +631,14 @@ This function deletes a specified env entirely. It takes env id `eid` as input.
 > **Note**: `delete_env` is deletes all data for an environment and is IRREVERSIBLE. Do not use unless you absolutely want to remove an environment.
 
 
+#### vis.fork
```
JackUrb: I think it would be more appropriate to use `fork_env` as we may have something specifically for forking layouts in the future. CamelCase locations should then use `ForkEnv`.

py/visdom/server.py
```diff
@@ -840,6 +845,29 @@ def post(self):
...
+        self.state = app.state
+        self.login_enabled = app.login_enabled
+
+    @staticmethod
+    def wrap_func(handler, args):
+        prev_eid = args.get('prev_eid')
+        eid = args.get('eid')
+
+        if prev_eid not in handler.state:
+            return handler.write('TODO: env doesn\'t exit')
```
JackUrb: You can use an assert in this case, the error should propagate to the visdom client, which I think is an appropriate outcome to trying to fork an environment that doesn't exist. Note in the docs that such an exception will occur if an env that doesn't exist is forked.

py/visdom/server.py
```diff
@@ -840,6 +845,29 @@ def post(self):
...
+
+        handler.state[eid] = copy.deepcopy(handler.state[prev_eid])
+        handler.write('TODO: missing proper reply')
```
JackUrb: Proper reply here is probably either `True` or the new env_id. As the failure case will be an exception, there's no real requirement for the success case.

py/visdom/server.py
```diff
@@ -856,11 +856,10 @@ def wrap_func(handler, args):
         prev_eid = args.get('prev_eid')
         eid = args.get('eid')
 
-        if prev_eid not in handler.state:
-            return handler.write('TODO: env doesn\'t exit')
+        assert prev_eid in handler.state, 'env to be forked doesn\'t exit'
 
         handler.state[eid] = copy.deepcopy(handler.state[prev_eid])
```
JackUrb, line 863: Just realizing you're missing a few small bits based on the implementation of fork that's present in SocketHandler:
```
eid = escape_eid(eid)
```
is used to make sure that the eid is valid before it's put in.
```
serialize_env(self.state, [eid], env_path=self.app.env_path)
```
ensures the new env is written to file once it's created, and 
```
broadcast_envs(self)
```
should ensure that open windows will receive the forked environment.

ricardofbarros: Fixed :)

ricardofbarros:
@JackUrb I thank you for guiding me through this feature :) 

I've fixed what you requested and I also noticed I coded a bug on handling `layout_item_update` on backend. 

It was overwriting the layout state of the env with a single window. I fixed that as well and now the feature works perfectly in any scenario of a fork. 

Before that fix if I resized a window and if I immediately forked that env, without saving, the new env wouldn't pick up those changes.

ricardofbarros:
Any ETA for this @JackUrb ?

JackUrb:
Sorry for the delay @ricardofbarros - I've been split across a number of different tasks. I'll get to it this week in my conference downtime

ricardofbarros:
Thanks for the reply :)

PR #611, opened 2019-04-23
Support for visualizing embeddings

pringshia:
Showing zoom capabilities and summary view:
embeddings-1

Showing the lasso tool and colored entities based on classification:
visdom-lasso-optimized

### To-Dos

- [X] Visualize point cloud data in 2d form
- [X] Provide a hover event which provides a way to relate to the original data
- [x] Allow ability to select a subset of points and only perform visualization on those points
- [x] Store a history of selected points to allow for a push/pop history view
- [X] Auto-scale perplexity
- [ ] Provide a mechanism to select embedding type and parameters to send an event to the Python-client algo (deferred)

## Description

We introduce a new `vis.embeddings` method that allows visualizing embeddings via running the t-SNE algorithm. The visualization runs using WebGL, allowing it to handle thousands of points at once. The user can pan, scroll-zoom in and out, and hover over points to get additional information. Each entity can optionally specify an HTML representation defined server-sided. If specified, this will get pulled in and displayed when the user hovers over an entity. In our mnist-example, this is used to show an image of the handwritten number!

We also allow re-running t-SNE on a sub-selection of points after the initial plot. For this we include the lasso tool. A minimum of 22 points must be selected for this to work.

## Motivation and Context

This enhances visdom by adding the ability to visualize embeddings via running the t-SNE algorithm. It's goal is to make the ML debugging/interpretability experience easier.

## How Has This Been Tested?

Example usage (from `/example/mnist-embeddings.py`):
```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals

import visdom
import numpy as np
from PIL import Image  # type: ignore
import base64 as b64  # type: ignore
from six import BytesIO

features = np.loadtxt("example/data/mnist2500_X.txt")
labels = np.loadtxt("example/data/mnist2500_labels.txt")
vis = visdom.Visdom()

image_datas = []
for feat in features:
    img_array = np.flipud(np.rot90(np.reshape(feat, (28, 28))))
    im = Image.fromarray(img_array * 255)
    im = im.convert('RGB')
    buf = BytesIO()
    im.save(buf, format='PNG')
    b64encoded = b64.b64encode(buf.getvalue()).decode('utf-8')
    image_datas.append(b64encoded)


def get_mnist_for_index(id):
    image_data = image_datas[id]
    display_data = 'data:image/png;base64,' + image_data
    return "<img src='"+display_data+"' />"


vis.embeddings(features, labels, data_getter=get_mnist_for_index, data_type='html')

input('Waiting for callbacks, press enter to quit.')
```

## Types of changes

## Checklist:


JackUrb:
Current status from the added mnist demo:
Screen Shot 2019-04-25 at 6 47 46 PM

pringshia:
The `bhtsne` algorithm requires a minimum # of points equal to 3 * perplexity: https://github.com/lvdmaaten/bhtsne/blob/master/tsne.cpp#L75

The [default perplexity](https://github.com/lvdmaaten/bhtsne/blob/master/bhtsne.py#L176) is set to 50.

Hence, currently the lasso tool silently fails when < 150 points are selected. Perhaps we should surface something more actionable when this occurs.

JackUrb:
Nice find - perhaps we can scale perplexity down when under 150 points? and then set a hard minimum of like 21 points? (because who is going to actually be running these algorithms any lower than that? embeddings are mostly useful in aggregate)

pringshia:
Agreed - this sounds like a good approach to me.

pringshia:
One additional small thing that I think would be big in terms of usability.

When the user invokes `vis.embeddings` from Python code or triggers a lasso re-processing, I think it would be great for the backend to send an initial ping to the window before conducting the `do_tsne` work. This way, the UI can display some sort of loading indicator that informs the user that the backend has received and acknowledged the request, and work has begun (since t-sne can sometimes take a while to run).

We can bikeshed on the format of the ping, but the main goal here is to have an instant response first, albeit indicating a loading state, followed by the actual data load, to give a sense of responsiveness.

Some UX-related points this will address:
- Currently when I run `vis.embeddings`, I am unsure if things "actually worked". Sometimes things silently fail and I'm not aware immediately, I wait thinking that the embeddings are still loading.
- Sometimes the back-end disconnects between the initial load and me using the lasso-tool. For example, in cases where there's a timeout disconnect. There's no clear instruction to me at this point whether me using the lasso tool actually kicked off work or not.

py/visdom/__init__.py
```diff
@@ -42,6 +44,30 @@
 except ImportError:
     BS4_AVAILABLE = False
 
+try:
+    # TODO try to import https://github.com/CannyLab/tsne-cuda first? will be
+    # faster but requires more setup
+    import visdom.extra_deps.bhtsne.bhtsne as bhtsne
+
+    def do_tsne(X):
+        Y = bhtsne.run_bh_tsne(X, initial_dims=X.shape[1], verbose=True)
```
pringshia: @JackUrb - any thoughts on leaving `verbose` set to `True` so it's easier to see the status of the processing on the server? I can't think of any disadvantages after cursory reflection, other than if there are a lot of points it may drown out other console logging from the server.

JackUrb: That's fine by me 👍

js/main.js
```diff
@@ -1411,7 +1453,12 @@ class App extends React.Component {
               isFocused={pane.id === this.state.focusedPaneID}
               w={panelayout.w}
               h={panelayout.h}
-              appApi={{ sendPaneMessage: this.sendPaneMessage }}
+              width={this.w2p(panelayout.w)}
+              height={this.h2p(panelayout.h) - 14}
```
JackUrb: At some point this magic number should end up in a constant

package.json
```diff
@@ -18,19 +18,25 @@
...
     "lint-staged": "^8.1.4",
-    "prettier": "1.16.4",
+    "prettier": "1.17.0",
     "react-select": "^2.0.0-alpha.3",
     "webpack": "^1.14.0"
   },
   "scripts": {
     "dev": "webpack --watch --progress --colors",
-    "build": "webpack -p",
+    "build": "webpack",
```
JackUrb: Any reason we're dropping the p flag?

pringshia: Ah it was removed while I was debugging, will add it back in. Nice catch.

js/main.js
pringshia: Moved it to a constant for now. Ideally the css layout will be done in a way in the future where this won't be needed.

PR #631, opened 2019-06-10
Adds support for transferring env state to/from client

andyljones:
## Description
Currently Visdom handles serialization with `save`, but this means that log files end up in a different place to the rest of the project. This PR adds a `set_window_data` function that, together with the existing `get_window_data`, can be used to serialize data on the client.

It also fixes `get_window_data`/`close` by changing `win=None` to `win='all'`, as `_send` assumes `None` means 'create a new window' and helpfully fills in `win=None` with a random string.

## How Has This Been Tested?
I've manually checked it works for setting the data of environments that don't currently exist on the server. It doesn't yet work for updating environments that already exist - what's the right way to trigger a reload on all the clients?

## Types of changes

## Checklist:


JackUrb:
In theory this is a breaking change, as `win='all'` would mean windows can no longer be named `'all'`. It works here, but I think the need for this is exposing that we've outgrown the setup for `_send` and should likely be abstracting the sending functionality for send types that are intentionally trying to target all windows.

As far as being able to force reloads on the frontends, you can use the globally defined function `broadcast_envs(self)` from any handler.

andyljones:
Good point about it being a messy API. I've changed it to passing `create=False` to `_send` to indicate that `win=None` means 'broadcast'. Made for a smaller modification than splitting the `_send` function in two.

PR #633, opened 2019-06-12
Adding optional polling-based comms to frontend

JackUrb:
## Description
Creates piping to use the same API for both socket-based and request-based communication between the frontend client and the server.

## Motivation and Context
Up until now, the only way to communicate with a visdom backend has been through sockets. Unfortunately, in some network configurations, socket-based communications are not scalable or not feasible. To this end, this PR introduces a method to have both.

## How Has This Been Tested?
Ran the demo both with `-use_frontend_client_polling` and without it.

With:
Screen Shot 2019-06-12 at 6 52 07 PM
Without:
Screen Shot 2019-06-12 at 6 50 23 PM

Note that the batching element of the query-based version has led to a different ordering of windows. In observation this has to do with the number of times repack is called on different batch sizes.

## Types of changes

## Checklist:

- [x My code follows the code style of this project.


py/visdom/server.py
```diff
@@ -382,7 +386,7 @@ def on_message(self, message):
                 logging.info('closing window {}'.format(msg['data']))
                 p_data = self.state[msg['eid']]['jsons'].pop(msg['data'], None)
                 event = {
-                    'event_type': 'Close',
+                    'event_type': 'close',
```
JackUrb, line 389: Note - this fixed a bug where close events weren't being picked up on the frontend due to capitalization

webpack.config.js
```diff
@@ -14,6 +14,10 @@ module.exports = {
   output: {
     filename: 'py/visdom/static/js/main.js',
   },
+  node: {
+    net: 'empty',
+    dns: 'empty',
```
pringshia, line 19: Is this needed because of the `fetch` library?

py/visdom/server.py
```diff
@@ -428,6 +432,142 @@ def on_close(self):
             self.subs.pop(self.sid, None)
 
 
+# TODO condense some of the functionality between this class and the
+# original SocketHandler class
```
pringshia, line 436: Agreed that it would be nice to reuse functionality between both implementations, but that can be done later

py/visdom/server.py
```diff
@@ -428,6 +432,142 @@ def on_close(self):
...
+
+        if self.readonly:
+            return
+
+        if cmd == 'close':
+            if 'data' in msg and 'eid' in msg:
+                logging.info('closing window {}'.format(msg['data']))
+                p_data = self.state[msg['eid']]['jsons'].pop(msg['data'], None)
+                event = {
+                    'event_type': 'Close',
```
pringshia: Does this need to be a lowercase `'close'` as well?

py/visdom/server.py
```diff
@@ -428,6 +432,142 @@ def on_close(self):
...
+    def write_message(self, msg):
+        self.messages.append(msg)
+
+    def get_messages(self):
+        to_send = []
+        while len(self.messages) > 0:
+            message = self.messages.pop()
+            if type(message) is dict:
+                # Not all messages are being formatted the same way (JSON)
+                # TODO investigate
```
pringshia, line 564: what is the symptom of this, i.e. in what unintended way are these messages being formatted? what are potential implications?

webpack.config.js
```diff
@@ -29,7 +33,10 @@ module.exports = {
         test: /\.css$/,
         loaders: ['style', 'css'],
       },
+      {
+        test: /\.json$/,
+        loader: 'json-loader',
+      },
     ],
   },
-  plugins: [new webpack.BannerPlugin('@generated')],
```
pringshia: Why is this being removed? Is this an accident? It was originally added to prevent lint from validating the output static bundle which reports a ton of errors.

js/main.js
```diff
@@ -263,10 +361,18 @@ class App extends React.Component {
...
 
+  setupPolling = () => {
+    this._socket = new Poller(this);
+  };
+
   connect = () => {
     if (this._socket) {
       return;
     }
+    if (USE_POLLING == 'True') {
```
pringshia: Is it an easy fix to have this be set to the js native value of `true`? This seems a bit magic string-y.

webpack.config.js
JackUrb: Yup

webpack.config.js
JackUrb: This was a mistake! I'll add it back

py/visdom/server.py
JackUrb: The symptom was that sometimes the packet sent to the frontend via polling would not be properly encoded for sending, and would fail to be decoded on the other end. This is actually super inefficient to be Json encoding windows, so if it turns out to be possible to not encode them individually I want to swap to doing that. For some reason the socket script was doing fine at decoding both cases, which I didn't take the time to fully understand during this pass

py/visdom/server.py
JackUrb: In theory it should be - I hadn't observed it as the pane is already closed on the frontend someone is observing from 😅

js/main.js
JackUrb: It just amounts to changing the location where I do the magic. I'll put it in the place where USE_POLLING is set

py/visdom/server.py
pringshia: hmm okay, if it's not causing any major issues then it's fine by me.

JackUrb: Yeah if anything this just happened to expose that JSON encoding magic has already been going on behind the scenes and I should examine that at some point.

wi1k1n:
I am experiencing a bug with this feature while using visdom with login flags. Here are my observations:

`visdom -p 6008 -use_frontend_client_polling`

This line runs the server in long polling mode and everything looks to work fine

`VISDOM_COOKIE=mycookie VISDOM_USERNAME=myuser VISDOM_PASSWORD=mypassword VISDOM_USE_ENV_CREDENTIALS=1 visdom -enable_login -port 6008`

This line also seem to be ok: server asks for credentials and then starts socket connection.

`VISDOM_COOKIE=mycookie VISDOM_USERNAME=myuser VISDOM_PASSWORD=mypassword VISDOM_USE_ENV_CREDENTIALS=1 visdom -enable_login -port 6008 -use_frontend_client_polling`

This line throws an exception. Here I am trying to run server with enabled login feature and in polling mode instead of sockets but the server runs well only until client tries to open connection (enters his credentials or opens http://localhost:6008 in browser with already saved cookies).


I have tested this on Ubuntu 19.1 and Windows 10 with both python 2.7 and 3.7 (in both cases visdom 0.1.8.9). All these tests showed the same behavior.

The exception message is as follows:
> File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 582, in open
    if self.login_enabled and not self.current_user:
AttributeError: 'ClientSocketWrapper' object has no attribute 'current_user'

The whole traceback stack is below:
> ERROR:tornado.application:Uncaught exception GET /socket_wrap (::1)
HTTPServerRequest(protocol='http', host='localhost:6008', method='GET', uri='/socket_wrap', version='HTTP/1.1', remote_ip='::1')
  File "/home/mazlov/.local/lib/python3.7/site-packages/tornado/web.py", line 1701, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 78, in _check_auth
    f(self, *args, **kwargs)
  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 1172, in get
    new_sub = ClientSocketWrapper(self.app)
  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 551, in __init__
    self.open()
  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 582, in open
    if self.login_enabled and not self.current_user:
AttributeError: 'ClientSocketWrapper' object has no attribute 'current_user'
ERROR:root:ERROR: 500: {'exc_info': (<class 'AttributeError'>, AttributeError("'ClientSocketWrapper' object has no attribute 'current_user'"), <traceback object at 0x7f1241ff2780>)}
INFO:root:Traceback: ['Traceback (most recent call last):\n', '  File "/home/mazlov/.local/lib/python3.7/site-packages/tornado/web.py", line 1701, in _execute\n    result = method(*self.path_args, **self.path_kwargs)\n', '  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 78, in _check_auth\n    f(self, *args, **kwargs)\n', '  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 1172, in get\n    new_sub = ClientSocketWrapper(self.app)\n', '  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 551, in __init__\n    self.open()\n', '  File "/home/mazlov/.local/lib/python3.7/site-packages/visdom/server.py", line 582, in open\n    if self.login_enabled and not self.current_user:\n', "AttributeError: 'ClientSocketWrapper' object has no attribute 'current_user'\n"]
ERROR:root:rendering error page
ERROR:root:[Errno 2] No such file or directory: '/home/mazlov/.local/lib/python3.7/site-packages/visdom/static/error.html'`

JackUrb:
Thanks for flagging @wi1k1n. It should be sufficient to just remove the whole `if self.login_enabled and not self.current_user:` block from `open` as the auth has already happened by this point. Feel free to submit a PR with this change!

PR #638, opened 2019-06-18
[tsne] Auto-scale perplexity, enforce 22 pts min on UI, add loading screen on first call

pringshia:
This will be merged into the WIP t-SNE branch #611 

It addresses a few things we discussed before:
- a loading screen on the initial python call to `vis.embeddings` (I was not able to get this to work on lasso selection however)
- a minimum of 22 points must be lassoed on the UI to run embeddings
- perplexity will be autoscaled for any # of points 22 and above such that perplexity = floor(# of points / 3), capping at 50

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -50,7 +50,12 @@ try:
     import visdom.extra_deps.bhtsne.bhtsne as bhtsne
 
     def do_tsne(X):
-        Y = bhtsne.run_bh_tsne(X, initial_dims=X.shape[1], verbose=True)
+        num_entities = len(X)
+
+        # the number of entities provided must be 
+        perplexity = 50 if num_entities >= 150 else \
+            num_entities // 3 if num_entities >= 21 else 7
+        Y = bhtsne.run_bh_tsne(X, initial_dims=X.shape[1], perplexity=perplexity, verbose=True)
         xmin, xmax = min(Y[:, 0]), max(Y[:, 0])
         ymin, ymax = min(Y[:, 1]), max(Y[:, 1])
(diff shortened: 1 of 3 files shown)


py/visdom/__init__.py
```diff
@@ -50,7 +50,12 @@
     import visdom.extra_deps.bhtsne.bhtsne as bhtsne
 
     def do_tsne(X):
-        Y = bhtsne.run_bh_tsne(X, initial_dims=X.shape[1], verbose=True)
+        num_entities = len(X)
+
+        # the number of entities provided must be
```
pringshia: whoops forgot to finish this comment. will fix

py/visdom/__init__.py
```diff
@@ -1073,6 +1078,17 @@ def embeddings(self, features, labels, data_getter=None, data_type=None,
         _title2str(opts)
         _assert_opts(opts)
 
+        loading_message = {
```
JackUrb, line 1081: You _could_ add almost this exact setup before the `do_tsne` call in the `elif event['event_type'] == 'RegionSelected':` branch to make this work with region selected as well. (the primary difference being sending the already existing data but setting `isLoading`. Ultimately this isn't particularly efficient though, I think long term we're going to need to support the same kind of event system that we have in the reverse direction into the forward direction. Might be worth tabling the want to have a loading indicator for selecting regions until we work out what that might look like.

pringshia: I tried mirroring the setup but was getting a cryptic error. Will show you in person.

pringshia: Discussed offline to table this for now. Marking as resolved.

PR #643, opened 2019-06-27
Integrate PR feedback for embeddings vis

pringshia:
- Download file renaming and updating the payload so the file actually has valid data.
- Remove mnist data from repo and instruct users on where to download from when running the demo code.
- Fixed bug where componentWillMount was being used instead of componentWillUnmount.
- Remove all UNSAFE_* methods from the EmbeddingsPane file.


example/mnist-embeddings.py
```diff
@@ -14,9 +14,15 @@
 from PIL import Image  # type: ignore
 import base64 as b64  # type: ignore
 from six import BytesIO
+import sys
+
+try:
+    features = np.loadtxt("example/data/mnist2500_X.txt")
+    labels = np.loadtxt("example/data/mnist2500_labels.txt")
+except:
+    print("Unable to find files mmist2500_X.txt and mnist2500_labels.txt in the example/data/ directory. Please download from https://github.com/lvdmaaten/lvdmaaten.github.io/blob/master/tsne/code/tsne_python.zip")
```
JackUrb, line 23: I feel like we should conditionally ask the user if they want to download and then do it ourselves, but this works fine for now

pringshia: +1, I'll slot that as an enhancement after this lands.

Issue #399, opened 2018-06-22
Scale images to window size?


malmaud:
When plotting small images with `Visdom.image`, it would be nice if the image resized to fit its window (and/or have zoom widgets) since small (32x32) images are pretty hard to see on a 4k display if there's a 1-to-1 correspondence between display pixel and image pixel.

JackUrb:
Hi @malmaud - we actually do have zoom (ctrl-scroll or pinch controls) on image windows currently, however fitting the image to the window on resize would be pretty cool to have as well. Will investigate.

JackUrb:
A quick note to anyone who is interested in implementing this - the fix relies on altering scale similarly to how it's done: 
https://github.com/facebookresearch/visdom/blob/f340284d5e5671908f3645139b3385d1a2585913/js/ImagePane.js#L17
but specifically when the `h` or `w` props change. You'll need to query the height and width of the image itself directly from the source (which is often done by prepping a dummy image component and getting its size, but never rendering it).

JackUrb:
This was implemented in #657. Thanks @ml7

PR #657, opened 2019-08-14
Address issue #399

ml7:
## Description

Allows image to resize itself to better fit the window. Resizing is done dynamically and matches its window resize activity. It is also now centered along the larger axis for better presentation.

When the user resizes the window, we want the zoom/drag settings to also reset themselves. This is because the image may go off-frame and disappear. We also handle edge cases where, for example, the container window is taller than it is wide, and the image itself is even taller than the window but has the same width. 

## Motivation and Context

This change improves the Visdom user experience.

https://github.com/facebookresearch/visdom/issues/399

## How Has This Been Tested?

Manual test

Rebuild Visdom from modified source code. Run demo.py and add a small 32x32 px image to view changes.

## Types of changes

## Checklist:


js/main.js
```diff
@@ -1553,22 +1555,24 @@ class App extends React.Component {
...
-              w={panelayout.w}
-              h={panelayout.h}
-              width={this.w2p(panelayout.w)}
-              height={this.h2p(panelayout.h) - PANE_TITLE_BAR_HEIGHT}
-              appApi={{
-                sendPaneMessage: this.sendPaneMessage,
-                sendEmbeddingPop: this.sendEmbeddingPop,
-              }}
-            />
+            <ReactResizeDetector handleWidth handleHeight>
```
JackUrb, line 1558: As a question of curiosity and for the record - was the RGL `onResize` event not sufficient to handle this kind of event? What does the new dependency grant us that wasn't supported?

ml7: The `onResize` event seemed to apply to the entire Grid Layout itself. I was more interested in detecting individual `onResize` events for each of the `<div>` elements, and triggering a re-render for individual elements only on resize. Plus the new dependency offers a really clean solution as we don't have to set up an `EventListener` and attach it to the corresponding DOM Element for each item (it's instead only 2 lines of extra code). Hope this offers some clarity!

PR #715, opened 2020-03-25
Unexpect Error due to MathJax Extension removed

ubaidsworld:
The issues in the console log was remove.
I got the reference from here: support.google.com/chrome/forum/AAAAP1KN0B0Uj6DitMGK1A

## Description
There was an error:
image
After my edit the error is gone:
image
## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/static/index.html
+++ b/py/visdom/static/index.html
@@ -30,7 +30,7 @@ LICENSE file in the root directory of this source tree.
     <script src={{ static_url("js/layout_bin_packer.js") }}></script>
 
     <!-- Mathjax -->
-    <script type="text/javascript" async src={{ static_url("js/mathjax-MathJax.js") }}></script>
+    <script type="text/html" async src={{ static_url("js/mathjax-MathJax.js") }}></script>
 
     <!-- Plotly -->
     <script src={{ static_url("js/plotly-plotly.min.js") }}></script>


ubaidsworld:
Hello Can someone take a look at this branch please?

py/visdom/static/index.html
```diff
@@ -30,7 +30,7 @@
     <script src={{ static_url("js/layout_bin_packer.js") }}></script>
 
     <!-- Mathjax -->
-    <script type="text/javascript" async src={{ static_url("js/mathjax-MathJax.js") }}></script>
+    <script type="text/html" async src={{ static_url("js/mathjax-MathJax.js") }}></script>
```
JackUrb, line 33: This is likely hiding the real issue - the problem is that `mathjax-MathJax.js` is not being properly copied into the `static/js` directory when the `download_scripts` function is called (which occurs here):
https://github.com/facebookresearch/visdom/blob/87419959811e2e0fbfeb42417d33a811ca6c8922/py/visdom/server.py#L1656
And because of that, what's being returned is likely a 404 not found page rather than the javascript resource. The actual mathjax script available [here](https://github.com/facebookresearch/visdom/blob/87419959811e2e0fbfeb42417d33a811ca6c8922/py/visdom/server.py#L1656) is surely a valid js file, so there's definitely a real problem lying underneath here, as we shouldn't need to say that this file is `html`. 

If you're still interested in debugging this issue, you can check the network tab of your browser's developer/inspector panel to see what the return code is for this query, and then go from there.

ubaidsworld: Thank you fr your reply, @JackUrb .
I will try it out once again will suggest you the best. I thought that the issue might get resolve if we just remove it from the console. As I have seen many issues complain regarding the console message.
Hence, I thought it would be better if we do this. But as it is not acceptable solution, I will make another attempt.

ubaidsworld: I have proposition @JackUrb , can we change the url from `https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG` to `https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js `in the server.py file?
I have got relieved after I changed the cdn url in the index.html, and later I have downloaded the file manually and replaced it in place of MathJax.js. It worked.

ubaidsworld: Hello @JackUrb , how about replacing the URL in server.py from: `https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG` to: `https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js`?

ubaidsworld: I guess I have solved it.. I will make another pull request with the new solution that I tried. May be that will help.

PR #726, opened 2020-04-03
Removed MathJax js console error issue

ubaidsworld:
Mathjax error were visible in the console and now the error is gone.

## Description

Mathjax error were visible in the console and now the error is gone.

## Motivation and Context

There were many issues regarding the js library used in the visdom and it was showing console error. I do not like the console errors. Hence, thought of removing it.
## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/server.py
+++ b/py/visdom/server.py
@@ -1653,8 +1653,8 @@ def download_scripts(proxies=None, install_dir=None):
             'react-dom.min.js',
         '%sreact-modal@3.1.10/dist/react-modal.min.js' % b:
             'react-modal.min.js',
-        'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG':  # noqa
-            'mathjax-MathJax.js',
+        #'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG':  # noqa
+            #'mathjax-MathJax.js',
         # here is another url in case the cdn breaks down again.
         # https://raw.githubusercontent.com/plotly/plotly.js/master/dist/plotly.min.js
         'https://cdn.plot.ly/plotly-latest.min.js': 'plotly-plotly.min.js',
@@ -1742,6 +1742,12 @@ def download_scripts(proxies=None, install_dir=None):
             except URLError as exc:
                 logging.error('Error {} while downloading {}'.format(
                     exc.reason, key))
+
+    filename = '%s/static/%s/%s' % (install_dir, "js", "MathJax.js")
+    if not os.path.exists(filename):
+        mathjax_js = requests.get("https://raw.githubusercontent.com/mathjax/MathJax/master/es5/tex-mml-svg.js")#("https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG")
+        with open(filename,"w") as file:
+            file.write(mathjax_js.text)
 
     if not is_built:
         with open(built_path, 'w+') as build_file:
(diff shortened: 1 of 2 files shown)


py/visdom/server.py
```diff
@@ -1743,6 +1743,12 @@ def download_scripts(proxies=None, install_dir=None):
                 logging.error('Error {} while downloading {}'.format(
                     exc.reason, key))
 
+    filename = '%s/static/%s/%s' % (install_dir, "js", "MathJax.js")
+    if not os.path.exists(filename):
+        mathjax_js = requests.get("https://raw.githubusercontent.com/mathjax/MathJax/master/es5/tex-mml-svg.js")#("https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG")
```
JackUrb: Is it not sufficient to just replace the url in line 1656 with this one?

JackUrb:
Thanks for taking another stab at this - almost ready to accept it out right, just one question of curiosity on the implementation. 

If my suggestion doesn't work, just remove the commented lines 1656-1657 entirely and then remove the additional comment on 1748. Also, you should use a more descriptive variable name than `filename` considering it's specifically the mathjax file, such as `mathjax_filename`.

py/visdom/static/index.html
```diff
@@ -29,12 +29,12 @@
...
 
-    <!-- Mathjax -->
-    <script type="text/javascript" async src={{ static_url("js/mathjax-MathJax.js") }}></script>
-
     <!-- Plotly -->
     <script src={{ static_url("js/plotly-plotly.min.js") }}></script>
 
+    <!-- Mathjax -->
+    <script type="text/javascript" async src={{ static_url("js/MathJax.js") }}></script>
+
```
JackUrb: It's important that the `MathJax` library is imported _before_ the `Plotly` library, so please don't swap the ordering here.

ubaidsworld: @JackUrb  I do not ant to swap the lines. But what I found I will share with you.
Actually, while I was testing the edited code for mathjax, I have found that Plotly suddenly gave some error and that were logged in the console.
But when I swapped the loading positions of the JS, it was working fine.
But if you think that it is not appropriate then I will make the changes. 
Here take a look at this error in the console when plotly is placed after the Mathjax:
image

Let me know and I will make the changes if you say so.

py/visdom/server.py
ubaidsworld: I guess it works.

py/visdom/static/index.html
JackUrb: Ah okay so there's still something weird going on with this import - The entire goal is to be able to use `LaTeX` in Plotly, and `MathJax` is supposed to enable that. When `MathJax` is imported before Plotly, Plotly attempts to integrate with the library.

py/visdom/server.py
ubaidsworld: Just give me a little time I will fix this.

py/visdom/static/index.html
ubaidsworld: ok .. I got it ... just give me some time and it will be fixed.. Insha allah..

JackUrb:
Alright digging into this a little more, I suspect the attempt to move MathJax to a local download in order to help those operating behind firewalls was somewhat incomplete. `MathJax.js` relies on the directory structure hosted on a cdn like `https://cdn.jsdelivr.net/npm/mathjax@2.7.1/` to load the correct resources, thus the fix in 2d12134 broke `MathJax`. If we wanted to get this to work locally, we'd need to download the entire package to a local directory.

For now, I think it'd be sufficient to go back to importing directly from a CDN in py/visdom/static/index.html. `https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_SVG` or `https://cdn.jsdelivr.net/npm/mathjax@2.7.1/MathJax.js?config=TeX-AMS-MML_SVG` would both be sufficient.

ubaidsworld:
@JackUrb Yes, I got it . I will add the MathJax folder separately in the project and that will do the magic. I am closing this request and will pull a new request, with the proper addons.

PR #675, opened 2019-09-25
[Big change][WIP] Server refactor

JackUrb:
## Description
The `server.py` file gradually grew from a bootstrapped tornado copy-paste into a full fledged server, but at no point had it ever been truly refactored. As such, the whole application was somewhat difficult to understand or debug, and making contributions to the server wasn't particularly clear. 

This PR aims to refactor the server in a number of ways, with a few core changes that will be coming up in follow-on PRs that I will merge into this branch as they are accepted. 

## Explanation of changes
This specific PR starts the process by making 0 code logic changes. It simply splits up what used to be ~2000 lines of mostly undocumented code in one file into ~2000 lines of now documented code in separate files that describe their purposes.

## Upcoming refactor steps
- [x] Split the monolithic `server.py` file into a number of more self-contained classes
- [ ] Create a `data_model` that will abstract the operations that the server does on panes away from being directly in control of the handlers. Handlers should not be dereferencing arbitrary dicts. Move some of the `server_utils` into appropriate places following this.
- [ ] Standardize the socket protocol between the forward and backwards sockets. Make an expected type for messages. Propagate this standardization to the javascript and python clients. Update the `socket_handlers` to use this new functionality.
- [ ] Clean up the `web_handlers`.
- [ ] Celebrate a nicer server, then move on to fixing `__init__.py`.

## How Has This Been Tested?
Ran demo.py, all of the panes plot as expected. Will continue to test as such for each step. `python -m visdom.server` still functions as expected as well.

## Types of changes

- [x] Better Engineering
- [x] Bug fixes

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] For JavaScript changes, I have re-generated the minified JavaScript code.

PR #770, opened 2020-11-22
check types with isinstance; allow tuple, where possible

dizcza:
Dear Jack,

I stumbled upon a bad practice of checking the object type with `type(x) == y`. Also, a handy-used scenario is when the user populates a dict with keys as legend labels and values as the payload:
```python
names, values = list(zip(*dict.items()))
viz.line(..., opts=dict(legend=names))
```
In short, legend sometimes can be passed as a tuple rather than a list, and I thought there is nothing wrong with allowing tuples.

## How Has This Been Tested?
Tested by running the demo script.

## Types of changes
- [x] Other (non-breaking change which improves code readability)

## Checklist:

P.S. It ain't much, but it's honest work.
P.P.S. When I was fixing the bad practice here and there, I noticed some other small issues with the coding style like code duplication, old python usage, etc. At first, I felt tempted to dive deeply into fixing them all but noticed https://github.com/facebookresearch/visdom/pull/675, which, of course, should be merged before any further sugar-code fixes.
P.P.P.S. One could set up Travis or CircleCI builds like I'm doing [here](https://github.com/dizcza/pytorch-mighty/blob/master/mighty/tests/test_trainer.py#L24-L26) to finally test the changes properly on each PR. I can write CircleCI build script if you set up a server.

Best,
Danylo

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -140,7 +140,7 @@ def _title2str(opts):
 
 
 def _scrub_dict(d):
-    if type(d) is dict:
+    if isinstance(d, dict):
         return {k: _scrub_dict(v) for k, v in list(d.items())
                 if v is not None and _scrub_dict(v) is not None}
     else:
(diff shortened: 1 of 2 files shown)


JackUrb (approved):
Style fixes are always appreciated! Definitely hoping to get #675 in sometime in the very near future as we prepare transitioning the project. And following that I'm hoping standardizing some CI for Visdom will be next on the docket. Happy to tag you with an up-for-grabs when we get CI set up.

As always, thanks for the contribution!

Issue #755, opened 2020-09-10
Maybe visdom.line() has a bug


IMIMyzj:
**Bug Description**
When I use visdom.line() to draw a pic, the code give me the information "X and Y should be the same shape". So I go to see the source code, to find where I get wrong with. After thinking and testing, I think there may exist a bug.
The bug code is here:
```python
if Y.ndim == 2 and Y.shape[1] == 1:
    Y = Y.reshape(1, Y.shape[0])
    X = X.reshape(X.shape[0])

if Y.ndim == 2 and X.ndim == 1:
    X = np.tile(X, (Y.shape[1], 1)).transpose()

    assert X.shape == Y.shape, 'X and Y should be the same shape'
```
So if I send the function line(X=x_input, Y=y_input), where x_input is an 'N' tensor or 'Nx1' tensor , and y_input is an 'Nx1' tensor, that is I just want to draw a line.  When x_input and y_input go through following code,**y_input will be an '1xN' tensor, and x_input will be an 'N' tensor**.
```python
if Y.ndim == 2 and Y.shape[1] == 1:
    Y = Y.reshape(1, Y.shape[0])
    X = X.reshape(X.shape[0])
```
Then they can go through following code, after this, **X become an 'NxN' tensor**, so it will cause assert above.
```python
if Y.ndim == 2 and X.ndim == 1:
    X = np.tile(X, (Y.shape[1], 1)).transpose()
```
So I think the code should be:
```python
if Y.ndim == 2 and Y.shape[1] == 1:
    Y = Y.reshape(Y.shape[0])        # I modify here
    X = X.reshape(X.shape[0])

if Y.ndim == 2 and X.ndim == 1:
    X = np.tile(X, (Y.shape[1], 1)).transpose()

    assert X.shape == Y.shape, 'X and Y should be the same shape'
```

**Reproduction Steps**
1. We can reproduce using following code:
```python
x = np.array([[1], [2],[3]])   # x is 3x1, or use x=np.array([1,2,3]) where x is 3
y = np.array([[6], [6], [6]])  # y is 3x1
vis = visdom.Visdom(server="http://localhost", port=8097)
vis.line(y, x)
```
2. Then the code will return
```shell
  File "C:\Users\MelonEater\Miniconda3\envs\denoi\lib\site-packages\visdom\__init__.py", line 1694, in line
    assert X.shape == Y.shape, 'X and Y should be the same shape'
```

**Expected behavior**
See bug description.

**Screenshots**
[...]

PR #778, opened 2021-01-08
Enable plotting a single line in plot

ShirAmir:
## Description

Removed the following code lines from the line function:

  if Y.ndim == 2 and Y.shape[1] == 1:
                Y = Y.reshape(1, Y.shape[0])
                X = X.reshape(X.shape[0])

## Motivation and Context

Now supports plotting a single line rather than just multiple lines.

Solves the issue: "Maybe visdom.line() has a bug" https://github.com/fossasia/visdom/issues/755

## How Has This Been Tested?

I tested it on a personal script that introduced me to the problem. 

windows 10, python 3.7, anaconda, pytorch 1.4.0, visdom 0.1.8

## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -1688,11 +1688,7 @@ class Visdom(object):
             assert X.ndim == 1 or X.ndim == 2, 'X should have 1 or 2 dim'
         else:
             X = np.linspace(0, 1, Y.shape[0])
-
-        if Y.ndim == 2 and Y.shape[1] == 1:
-                Y = Y.reshape(1, Y.shape[0])
-                X = X.reshape(X.shape[0])
-
+     
         if Y.ndim == 2 and X.ndim == 1:
             X = np.tile(X, (Y.shape[1], 1)).transpose()
 


JackUrb:
Hi @ShirAmir - thanks for taking on resolving this issue.

If I recall correctly, this was [added](https://github.com/fossasia/visdom/commit/4d9f7c0e6a896943707fd255e0d46bc6798a42e3) to fix a bug relating to plotting a single point. Is it still possible to have a one point line plot after this change?

ShirAmir:
Hi @JackUrb, thanks for the quick reply!

To the best of my knowledge, plotting a single line with a single point fails both before and after my fix.

I did the following test to examine whether its possible to plot a single line with multiple points and a single line with one point. I saw the `line` function wraps the `scatter` function so I applied the same tests on it as well.

```
x = np.array([[1], [2], [3]])  # np.array([[1]]) 
y = np.array([[10], [20], [30]])  # np.array([[10]])
vis = visdom.Visdom(server="http://localhost", port=8097)
vis.line(y, x)
X = np.array([[1, 10], [2, 20], [3, 30]]) # np.array([[1, 10]])
vis.scatter(X)
```
image

Apparently a single point can't be visualized without setting `opt['markers'] = True`. This makes sense since plotting a single point in a line, without showing the markers themselves shouldn't visualize the point.

I'd consider setting the default behavior in `line` to show markers when there is a single point plotted. This can be done by applying this fix: 

```
if Y.ndim == 2 and Y.shape[1] == 1:
    Y = Y.reshape(Y.shape[0])   # instead of Y = Y.reshape(1, Y.shape[0])
    X = X.reshape(X.shape[0])

....

opts['markers'] = True if Y.ndim == 1 else opts.get('markers', False)        
```

Regardless, removing the above three lines as I originally suggested or applying the fix above is necessary to plot a single line with any amount of points.

JackUrb (approved):
The biggest issue I could find in the original post wasn't just that they couldn't _see_ a point plotted (which in my opinion, is _okay_ behavior), but rather that plotting would throw an error. So long as `viz.line(X=np.array([[1]]), Y=np.array([[1]]))` works (which failed in the original issue, but appears to work fine in your testing), this looks good to me.

PR #792, opened 2021-03-07
efficiency

al1216:
increased the efficiency of program by reducing namespace

## Types of changes

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,8 +65,8 @@ setup(
     packages=['visdom'],
     package_dir={'': 'py'},
     package_data={'visdom': ['static/*.*', 'static/**/*', 'py.typed', '*.pyi']},
-    include_package_data=True,
-    zip_safe=False,
+    include_package_data=bool(1),
+    zip_safe=bool(0),
     install_requires=requirements,
     entry_points={'console_scripts': ['visdom=visdom.server:download_scripts_and_run']}
 )
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@
 <!--- both on your branch and a clean branch and ensure that none of the functionality -->
 <!--- appears different. Be sure to install from source when testing. -->
 
-## Screenshots (if appropriate):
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -65,8 +65,8 @@ def get_dist(pkgname):
     packages=['visdom'],
     package_dir={'': 'py'},
     package_data={'visdom': ['static/*.*', 'static/**/*', 'py.typed', '*.pyi']},
-    include_package_data=True,
-    zip_safe=False,
+    include_package_data=bool(1),
+    zip_safe=bool(0),
```
JackUrb, line 69: How is this an improvement?

JackUrb (changes requested):

setup.py
al1216: @JackUrb In this instead of True or False I used bool(1) or bool(0) respectively. For better clarity and faster processing of program used this.

JackUrb: I'm entirely certain that `True` and `False` are clearer and more readable than `bool(1)` and `bool(0)`, and I'd wager making this call is less efficient. Can you point to any resource that suggests that replacing the python bool constant with _a bool cast on an int_ would be faster?

al1216: @JackUrb   No I don't have any resource that points out this . I learnt in my school days with my teacher that it's good to use bool(1) or bool(0) over True or False that sounds good in terms of technical aspects.

JackUrb: I'd refrain from applying that idea to Python in the future - you can always test this out yourself to see it's not true:
```
>>> import timeit
>>> from numpy import mean
>>> def check_false_speed():
...   if False:
...     return "hi"
...   else:
...     return "bye"
... 
>>> def check_bool_speed():
...   if bool(0):
...     return "hi"
...   else:
...     return "bye"
... 
>>> mean(timeit.repeat(check_false_speed, repeat=10))
0.05913775699999633
>>> mean(timeit.repeat(check_bool_speed, repeat=10))
0.09170382250000557
```

al1216: Okk. I got it . Thanks @JackUrb

PR #815, opened 2021-08-30
add flag to make server accessible only from localhost

beimbusch:
## Description
In this PR, I added a flag to the server CLI allowing to bind to localhost, i.e. `127.0.0.1`, denying external requests.

## Motivation and Context
In several situations, it might be sensible to have the server only accessible from localhost instead of the whole network, especially when one does not control the firewall of a machine. Granularly regulated external access is still feasible via SSH port forwarding.
Enabling the new option does not have an impact on the overall ease-of-use in many use cases but significantly reduces the potential attack surface.

## How Has This Been Tested?
This PR has been tested by locally running `server.py` in both configurations and by checking that the command line arguments parser still works correctly.

## Types of changes

## Checklist:


JackUrb (approved):
Looks like a fairly lightweight but definitely useful feature to me, thanks for the contribution!

Issue #749, opened 2020-07-21
get_env_list raises TypeError in offline mode


vfdev-5:
**Bug Description**
If call `get_env_list()` in offline mode, it raises `TypeError`.

**Reproduction Steps**
```python
>>> import visdom
>>> vis = visdom.Visdom(offline=True, log_to_filename="/tmp/log.visdom")
>>> vis.get_env_list()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.7/site-packages/visdom/__init__.py", line 822, in get_env_list
    return json.loads(self._send({}, endpoint='env_state', quiet=True))
  File "/opt/conda/lib/python3.7/json/__init__.py", line 341, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not bool
```

JackUrb:
In offline mode, we don't even send anything to the frontend:
https://github.com/fossasia/visdom/blob/e0a912d39f88e637f483c229c0fbf3d4da119ac5/py/visdom/__init__.py#L680-L682
As such, in offline mode we'd probably need to be keeping track of a local env list to solve this problem. You could do so right in the above location, intercepting sends that are setting different `env`s.

ath34-tech:
Ya, I thought the same,  can you point what function is responsible for creating env? 
I am thinking that whenever env is created we will add its data to the list and return that list. Or there is something that work different and I am still missing some point then please tell me.
Thank you!

JackUrb:
The local client doesn't do anything to create a new env, or create them itself, it has to do with the `env` value set by the time the `send` method is called. The server then creates an env.

ath34-tech:
Hey there,i manage to solve this issue, please check if it is right or wrong. 
Thank you!

ath34-tech:
hey @JackUrb , please review:https://github.com/fossasia/visdom/pull/824
thank you!

ath34-tech:
Hey @JackUrb I made changes to PR, please check them,
Thank you!

ath34-tech:
Hey @JackUrb , i formatted the code, please check if it is working.

ath34-tech:
Hey @JackUrb , please check my code on latest commit.

ath34-tech:
@JackUrb Please merge my PR.thank you

ath34-tech:
[...]

PR #821, opened 2021-11-05
Solved issue #749

ath34-tech:
I simply look into server.py gather_envs function and found out that this function is exactly what is needed to be added so I created a copy of that function in __init__.py and return a list from that function.
## Description
I added a function called get_all_env, which is similar to server.py's gather_envs functions, it simply return the name of all envs in list format. In get_all_env function I returned a sorted list which return name of all env.
## Motivation and Context
It solves issue #749 

## How Has This Been Tested?
I simply create a script that was reproducing issue(that script is told in issue I mentioned) :
`from visdom import Visdom

#  instantiate the window class 
```
vis = Visdom(offline=True, log_to_filename="./log.visdom")
print(vis.get_env_list())`

```
And output for this:
`
{'env list': "['main', 'main12', 'main13', 'main15', 'main2', 'main3', 'main6', 'myown']"}
`

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -2130,31 +2130,41 @@ class Visdom(object):
 
 
     @pytorch_wrap
-    def sunburst(self, X,parents,labels, win=None, env=None, opts=None):
+    def sunburst(self, labels, parents, values=None, win=None, env=None, opts=None):
         opts = {} if opts is None else opts
         _title2str(opts)
         _assert_opts(opts)
-        
+
         font_size=opts.get("size")
         font_color=opts.get("font_color")
         opacity=opts.get("opacity")
         line_width=opts.get("marker_width")
 
-        X = np.squeeze(X)
-        assert X.ndim == 1, 'X should be one-dimensional'
-        assert np.all(np.greater_equal(X, 0)), \
-            'X cannot contain negative values'
-
-
-        data = [{
-            'values': X.tolist(),
-            'labels': labels.tolist(),
-            "parents":parents.tolist(),
-            "outsidetextfont":{"size":font_size,"color":font_color},
-            "leaf":{"opacity":opacity},
-            "marker":{"line":{"width":line_width}},
-            'type': 'sunburst',
-        }]
+        assert len(parents.tolist())==len(labels.tolist()), "length of parents and labels should be equal"
+        
+        if(values!=None):
+            values = np.squeeze(values)
+            assert values.ndim == 1, 'values should be one-dimensional'
+            assert len(parents.tolist())==len(values.tolist()), "length of values should be equal to lenght of labels and parents"
+            data = [{
+                'values': values.tolist(),
+                'labels': labels.tolist(),
+                "parents":parents.tolist(),
+                "outsidetextfont":{"size":font_size,"color":font_color},
+                "leaf":{"opacity":opacity},
+                "marker":{"line":{"width":line_width}},
+                'type': 'sunburst',
+            }]
+        
+        else:
+            data = [{
+                'labels': labels.tolist(),
+                "parents":parents.tolist(),
+                "outsidetextfont":{"size":font_size,"color":font_color},
+                "leaf":{"opacity":opacity},
+                "marker":{"line":{"width":line_width}},
+                'type': 'sunburst',
+            }]            
         return self._send({
             'data': data,
             'win': win,
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -405,7 +405,8 @@ class Visdom(object):
             'base_url should not end with / as it is appended automatically'
 
         self.ipv6 = ipv6
-        self.env = env              # default env
+        self.env = env      
+        self.env_list={f'{env}'}      # default env
         self.send = send
         self.event_handlers = {}  # Haven't registered any events
         self.socket_alive = False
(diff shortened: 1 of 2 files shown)


py/visdom/__init__.py
```diff
@@ -679,8 +692,9 @@ def _send(self, msg, endpoint='events', quiet=False, from_log=False, create=True
 
         if self.offline:
             # If offline, don't even try to post
-            return msg['win'] if 'win' in msg else True
-
+            list_of_env=self.get_all_env()
+            return msg['win'] if 'win' in msg else json.dumps({"env list":f'{list_of_env}'})
```
JackUrb: You shouldn't be making this change in the `_send` call, as our expected behavior in offline mode is that sends return `True`. Instead the change should be not calling `_send` to begin with from in the `get_env_list` method.

py/visdom/static/js/layout-bin-packer.js.map
```diff
@@ -0,0 +1 @@
...5B,EAwDA,EAAE,IAAI,MAAM,KAAK,CAAC,IAAI,KAAK,GAAG,MAAM,EAAE;AAxDtC,EAyDA,IAAI,UAAU,CAAC,MAAM,EAAE,KAAK,CAAC;AAzD7B,EA0DA;;AA1DA,EA4DA,EAAE,IAAI,OAAO,GAAG,IAAI,CAAC,KAAK,CAAC,KAAK,GAAG,IAAI,CAAC,YAAY,CAAC,KAAK,CAAC,CAAC,IAAI,CAAC;;AA5DjE,EA8DA,EAAE,IAAI,CAAC,GAAG,KAAK,GAAG,OAAO,GAAG,IAAI,CAAC,YAAY,CAAC,KAAK,CAAC,GAAG,CAAC;AA9DxD,EA+DA,EAAE,IAAI,CAAC,GAAG,IAAI,CAAC,KAAK,CAAC,KAAK,GAAG,OAAO,CAAC,GAAG,IAAI,CAAC,aAAa,CAAC,KAAK,CAAC;;AA/DjE,EAiEA,EAAE,OAAO,CAAC,CAAC,CAAC,CAAC,EAAE,CAAC,CAAC,CAAC,CAAC;AAjEnB,EAkEA,CAAC;;AAlED,EAoEA,SAAS,CAAC,SAAS,CAAC,MAAM,GAAG,UAAU,YAAY,EAAE;AApErD,EAqEA,EAAE,IAAI,OAAO,YAAY,KAAK,QAAQ,EAAE;AArExC,EAsEA,IAAI,MAAM,IAAI,SAAS,CAAC,8DAA8D,CAAC;AAtEvF,EAuEA;AAvEA,EAwEA,EAAE,IAAI,MAAM,GAAG,IAAI,CAAC,MAAM,EAAE;AAxE5B,EAyEA,EAAE,IAAI,MAAM,KAAK,CAAC,EAAE,EAAE,OAAO,CAAC,CAAC;;AAzE/B,EA2EA,EAAE,IAAI,WAAW,GAAG,IAAI,CAAC,GAAG,CAAC,IAAI,CAAC,KAAK,CAAC,YAAY,CAAC,IAAI,CAAC,YAAY,CAAC,CAAC,CAAC,EAAE,CAAC,CAAC,CAAC;AA3E9E,EA4EA,EAAE,WAAW,GAAG,WAAW,GAAG,CAAC,GAAG,WAAW,GAAG,CAAC;AA5EjD,EA6EA,EAAE,IAAI,IAAI,GAAG,IAAI,CAAC,IAAI,CAAC,MAAM,CAAC,WAAW,CAAC;AA7E1C,EA8EA,EAAE,IAAI,WAAW,GAAG,IAAI,GAAG,IAAI,CAAC,aAAa,CAAC,CAAC,CAAC;;AA9EhD,EAgFA,EAAE,OAAO,WAAW;AAhFpB,EAiFA,CAAC,;;;;,;;"}
```
JackUrb: You shouldn't be adding anything to the static directory. This and below files aren't relevant to this PR.

py/visdom/__init__.py
```diff
@@ -503,6 +507,15 @@ def register_event_handler(self, handler, target):
...
         self.event_handlers[target] = []
 
+
+    def get_all_env(self):
+
+        if self.DEFAULT_ENV_PATH is not None:
+            items = [i.replace('.json', '') for i in os.listdir(self.DEFAULT_ENV_PATH) if '.json' in i]
+        else:
+            items = []
+        return items
```
JackUrb: This misses the mark - usually `visdom`'s client is run on a different machine than the `visdom` server, and as such the contents of `DEFAULT_ENV_PATH` will usually be empty on a client machine. You should instead be creating an `offline_env_list` or something like that in `__init__` if `offline`, and then in `send` if `offline`, check the `env` that visdom is trying to write to.

JackUrb (changes requested):
Hi @ath34-tech, thanks for the attempt here. Unfortunately this implementation is missing the mark. I've left some comments inline to describe what we're really going for.

To find out more about what's being sent in `_send`, try adding some `print` statements and running the `demo.py` - you should be able to see where in the arguments you can pull `env` from to populate the local env list.

py/visdom/__init__.py
ath34-tech: can you elaborate what you meant by check the "env" that visdom is trying to write to.

JackUrb: Check the args of what `visdom` is sending when calling `send` - an env can be included in every one of visdom's functions, like here: https://github.com/fossasia/visdom/blob/06f9cb22fdc4da2cb74f65faf0d6ec64d89e03fc/py/visdom/__init__.py#L884
It ends up being managed as an `eid`: https://github.com/fossasia/visdom/blob/06f9cb22fdc4da2cb74f65faf0d6ec64d89e03fc/py/visdom/__init__.py#L902
And send sets it to the default if none is provided:
https://github.com/fossasia/visdom/blob/master/py/visdom/__init__.py#L667-L668

ath34-tech (commented):
Hey did you asked me to do something like this?

PR #832, opened 2022-02-28
Lazy Loading of Environments

da-h:
## Description
This PR proposes to change all envs to be loaded lazily by default.

## Motivation and Context
In the case of many json files the `-env_path` folder, startup time and memory utilization increases linearly.
For example, the env created by `demo.py` and copied 1000 times takes about 52.7G of RAM and 277 seconds to start the visdom server.
(See below to recreate this simple used case).

Using this tools mainly for inspection of torch-experiments myself, I end up waiting up to half an hour until the visdom ui is only accessible.
This PR proposes to change all envs to be loaded lazily. Additionally, **this new feature will be activated by default**, as the feature induces no change of functionality and provides a much smoother user experience.

In the example above, lazy loading of the envs collapses startup time to 0.002 seconds and also the memory reduces to the "no-env" case RAM consumption of about 250M on my system.

**Note**: Personally, I rate this change as important enough for each user of visdom to always enable this feature. Thus, this PR adds a flag `-eager_data_loading` to pre-load the whole environment-folder as done previously.  
Please let me know, in case another behavior should be the default. I.e. If the change is not in the current direction of the project, I am also happy to change this PR in order to turn this feature on using a flag named `-lazy_load_data` for instance.

## How Has This Been Tested?

Using the `demo.py`, we create 1000 symlinks in `/tmp/logfiletest`.
1. Create env dir for the used case test: `mkdir -p /tmp/logfiletest`
2. Start the visdom server `visdom -env_path /tmp/logfiletest`
3. Execute the `demo.py` `python ./example/demo.py`
4. Save the created environment in the visdom web-ui. (Open http://localhost:8097/, click on the folder icon and press "save")
5. Stop visdom server using `^C`.
6. Create symlinks to simulate many envs. `cd /tmp/logfiletest; for i in $(seq 1000); do ln -s main.json ${i}.json; done`
7. Restart visdom server & check the loading time before and after this PR.

## Types of changes
    -> I expect no case to behave differently after this PR.

## Checklist:
    -> README.md adapted
    -> no JavaScript changed.


py/visdom/server.py
```diff
@@ -129,6 +130,40 @@ def hash_password(password):
...
+                'reload': env_data['reload']
+        }
+
+    def __getitem__(self, key):
+        if not self._raw_dict:
+            self._load_data()
+        return self._raw_dict.__getitem__(key)
+
+    def __iter__(self):
+        self._load_data()
```
JackUrb: Shouldn't this and the `__len__` implementation both mirror the `__getitem__` implementation in only loading if the load hasn't already happened? 

It may be cleaner to move the `None` check into `_load_data` and update it to only attempt the load if it hasn't already been done (after which, you could consider calling the function `_lazy_load_data` to better capture the fact that it doesn't always force a load).

JackUrb (changes requested):
This is a *great* addition @da-h, and I'm definitely supportive of making it the default option for visdom. One comment inline about potentially unintended repeat loads, then I'm happy to merge this in.

py/visdom/server.py
da-h: Good point, I've missed this!

I have changed the function `_load_data` as proposed and renamed the function to `lazy_load_data`.
(I feel that there is a utility in making it *public* in case some user needs to pre-fetch it in their script.)

Issue #82, opened 2017-05-10
`heatmap` append and more data cache/handling in visdom?


geyang:
Hi,

it looks like the both `line` and `scatter` has an `append` option in plotly, but `heatmap` does not.

In our experiment, we are using heatmap to render incoming streams of spectrum data, and it becomes quite slow to have to send over 100 kB of heatmap image for each additional column.

Is it within the scope of this project to add datacaching and functionalities like smoothing in this project?

If so, I can implement these inside this little helper module we use and make it available to everyone.

https://github.com/episodeyang/visdom_helper
Best,
Ge

ajabri:
Hi @episodeyang !

Yes, you're right, its something that IMO should definitely be there. In general, I think it would be useful to have a plot object that can be updated efficiently; I suppose the question is where should the data caching happen -- server side, client side? I guess there are tradeoffs. If you do it on the server, you still have to send the whole thing to the browser. If you do it in the browser, the state won't be updated on the server. One solution could actually be to make the update on the server and just send a diff of the actual message string so it's cheaper. 

I actually indeed already came across your visdom_helper, it looks cool. If you're up for it, we're all for it!

Allan

episodeyang:
@ajabri awesome! I'm up for it! I can keep the discussions here and link to the issues here!

JackUrb:
Fixed in #834

PR #834, opened 2022-03-02
Adding Update Function to Heatmap (in accordance to vis.line)

da-h:
## Description
The method `vis.line` enables efficient online-data tracking by using the `update=append` option.
This PR expands `vis.heatmap` accordingly to also allow appending data.
(*This should solve issue #82*).

**Note**: Before this can be merged, please check #836. It looks like the visdom client <-> server connection may be buggy in the case of multi-dimensional arrays.

**In more detail**:
- `vis.heatmap` gets a new argument named `update`. Similarly to `vis.line`, `update` can take one of the following values
    - `update="appendRow"`
    - `update="appendColumn"`
    - `update="prependRow"`
    - `update="prependColumn"`
    - `update="replace"`
    - `update="remove"`
- Data will be appended or replaced just as in `vis.line`. (Before data gets added, tha validity of the operation is checked with a respective error message in case it doesn't.)
- To maintain consistency of update-plots `vis.heatmap` requires the user to specify rownames or columnnames for the new data in case the have been specified upon initial `vis.heatmap`-call.
- In case the user did not specify names matplotlib expands the names canonically.
- Additional arguments given by `opts` overwrite the old ones. (I.e. the given example changes the colorscheme).
- To achieve automatic re-calculation of the colorscale, This PR adapts the `zmin`/`zmax` api pre-calculations, leaving the colorscale to matplotlib/plotly if not specified by the user. (Functionality wise, the result is the same as with `vis.line`).
- The PR also adds a small show case of the new features to `demo.py`.

**Note**:
- during the work on this PR, I have encountered a bug related to updating the plots in the respective javascript code. (additional PR coming very soon).
- Additionally, I have a question where I am a bit unsure how to tackle this in this project. Is there a way to send a message back to the client in case there has been a problem in `server.py`? Currently, this PR just dismisses operations that would create errors and the server (using `logging.error`) just displays what the error was. I would expect the API-call to get an error back, however, I did not find this feature yet.

## Motivation and Context
From all plots, appending data by replacing, becomes pretty quickly slow for heatmaps.
The goal is to improve the lack of performance in this used case by keeping the API as close as possible to the already existing "update"-feature of some of the other plots.

## How Has This Been Tested?
This PR adds tests to `demo.py` to showcase all of the cases above.

## Types of changes
    -> If not using the new `update`-argument, no function-changes should occur.

## Checklist:


py/visdom/__init__.py
```diff
@@ -1778,13 +1795,25 @@ def heatmap(self, X, win=None, env=None, opts=None):
...
             'eid': env,
-            'layout': _opts2layout(opts),
+            'layout': _opts2layout(opts) if update is None else {},
             'opts': opts,
-        })
+        }
+        endpoint = 'events'
+        if update:
+            # data_to_send['name'] = name
+            data_to_send['append'] = update in ['appendRow', 'appendColumn', 'prependRow', 'prependColumn']
```
JackUrb: It could make sense to add an assertion that `update` is in the expected value at the start of the function rather than having invalid values acting essentially as noops.

da-h: True, i've adapted the code accordingly.

da-h:
I like your suggestion how errors can be forwarded & I will think about it further and maybe make a suggestion when the opportunity arises.

PS: As i mentioned in another PR, i'm just starting to understand the project better, the one or the other nudge in the right direction is quite helpful. after i've been using the project for quite some time now, i thought i could put something back into the project whenever i have some free time. ;)

JackUrb:
Happy to have the help! I haven't had a ton of time to put into visdom so the contributions are greatly appreciated. I'll do my best to review and try to provide advice where I can 👍

PR #837, opened 2022-03-04
Enable Live-Edits of any Property of any Plot

da-h:
## Description
**Adding possibility to change all available plotpanes (e.g. axis-limits) on-the-fly**.
The feature added can be most easily explained by the short video added below.

## Motivation and Context
Pretty often I find myself restarting a script because a specific setting of a plot cannot be adapted easily in visdom.
This feature tries to solve this by adding the option to edit all `pane.content` variables on-the-fly.

**Some more details on the implementation**:
- to make it easier to add this feature, i have reused functionality of `PropertiesPane` by adding a `AbstractPropertiesList` Component.
- The `PropertyList` is added to the general `Pane`. I see that one could argue for a more specific implementation, possibly addit it only to `PlotPane` or so, but I felt that this feature could be expanded to a more general feature, usable by other types of panes as well.
- To enable this feature on a per-`Pane`-object basis i have added the option `enablePropertyList`, and `PlotPane` is using it to enable the feature in all plots.
- All simple data (boolean, ints, strings, and nested simple data) are shown in the list when clicking the new `editParameters`-Button in the top-right corner of every plot. The button turns red when enabled and closes the window again when pressed a second time.

## How Has This Been Tested?
I have to admit that I have tried some, but not all parameters used in `demo.py`.

record

## Types of changes

## Checklist:
   -> This is just a proposal for a place to mention this feature. I have used the previous video and placed it in the README.md where it fitted best in my opinion. AFAIK upload from PRs cannot be removed and will be accessible even after closing the PR.

### DIFF
--- a/js/AbstractPropertiesList.js
+++ b/js/AbstractPropertiesList.js
@@ -23,7 +23,7 @@ class MyRef {
   };
 }
 
-class Text extends React.Component {
+class EditablePropertyText extends React.Component {
   constructor(props) {
     super(props);
     this.textInput = new MyRef();
@@ -102,14 +102,14 @@ class AbstractPropertiesList extends React.Component {
     switch (prop.type) {
       case 'text':
         return (
-          <Text
+          <EditablePropertyText
             value={prop.value}
             submitHandler={value => this.updateValue(propId, value)}
           />
         );
       case 'number':
         return (
-          <Text
+          <EditablePropertyText
             value={prop.value}
             submitHandler={value => this.updateValue(propId, value)}
             validateHandler={value => value.match(/^[0-9]*([.][0-9]*)?$/i)}
--- a/js/AbstractPropertiesList.js
+++ b/js/AbstractPropertiesList.js
@@ -1,0 +1,156 @@
+/**
+ * Copyright 2017-present, The Visdom Authors
+ * All rights reserved.
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import React from 'react';
+
+class MyRef {
+  constructor() {
+    this._ref = null;
+  }
+
+  getRef() {
+    return this._ref;
+  }
+
+  setRef = ref => {
+    this._ref = ref;
+  };
+}
+
+class Text extends React.Component {
+  constructor(props) {
+    super(props);
+    this.textInput = new MyRef();
+    this.state = {
+      propsValue: props.value,
+      actualValue: props.value,
+      isEdited: false,
+    };
+  }
+
+  handleChange = event => {
+    let newValue = event.target.value;
+    if (this.props.validateHandler && !this.props.validateHandler(newValue)) {
+      event.preventDefault();
+    } else {
+      this.setState({
+        actualValue: newValue,
+      });
+    }
+  };
+
+  handleKeyPress = event => {
+    if (event.key === 'Enter') {
+      let ref = this.textInput.getRef();
+      if (ref) ref.blur(); // Blur invokes submit
+    }
+  };
+
+  onBlur = () => {
+    this.setState({ isEdited: false }, () => {
+      if (this.props.submitHandler) {
+        this.props.submitHandler(this.state.actualValue);
+      }
+    });
+  };
+
+  onFocus = () => {
+    this.setState({
+      isEdited: true,
+    });
+  };
+
+  UNSAFE_componentWillReceiveProps(nextProps) {
+    if (this.state.propsValue !== nextProps.value || !this.state.isEdited) {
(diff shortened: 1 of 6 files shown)


js/AbstractPropertiesList.js
```diff
@@ -0,0 +1,157 @@
...
+  getRef() {
+    return this._ref;
+  }
+
+  setRef = ref => {
+    this._ref = ref;
+  };
+}
+
+class Text extends React.Component {
```
JackUrb: Can we rename this class while we're moving it? `Text` was fairly nondescript before, perhaps `EditablePropertyText`?

js/AbstractPropertiesList.js
```diff
@@ -0,0 +1,157 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import React from 'react';
+
+class MyRef {
```
JackUrb, line 12: Seeing this again suggests that someday it'll be worthwhile to update to hook-based react. Certainly has made a lot of my more recent projects tractable. Unsure if it's something we can really adopt gradually though.

js/Pane.js
```diff
@@ -114,13 +126,108 @@ class Pane extends React.Component {
...
 
+class PropertyList extends AbstractPropertiesList {
+  _windowRef = null;
+  _barRef = null;
+
+  // updates the property of the window dynamically
+  // note: this.props refers in this content to the Components directly responsible
+  //       to the key, e.g. Text object from AbstractPropertiesList
+  updateValue = (key, value) => {
+    this.props.content[key] = value;
```
JackUrb, line 183: This is likely the source of the delays - editing the props directly like this isn't going to trigger react to do a re-render. Unfortunately I don't see a clean way to trigger such a thing short of having panes use both a `props` and `state` content, and then allowing property lists to do an `onUpdate` of sorts to change the state. Property lists would also want to maintain the list of properties they've edited, such that if a new set of content comes in via props, the edited properties can be re-applied.

JackUrb (commented):
This PR is really cool! As with all cool features at this scale though, I have a few questions/concerns to discuss before moving forward with it. Overall I think the choice to reuse the properties list was really clever, but there's a little work to be done for proper state/props management here. (I think in some ways you're being held back by the component-style react that Visdom has been built on, as it can make state maintenance harder to follow, and updates less clear.) In any case we can start the discussion primarily around the direct changes to `props.content` in `Pane.js`.

js/AbstractPropertiesList.js
da-h: Sure, I adapted the code accordingly.

js/AbstractPropertiesList.js
da-h: I've just read the reference of the react hooks API. I would like to rewrite some of the easier classes to the new API, but I suggest to create new PR for this for simplicity reasons.

js/Pane.js
da-h: Interesting. I have just re-checked when delays occur. Changing the variables, let's say the plot-type does trigger directly. However, opening & closing the Properties view, even though I have used a state property for this, has a delay of one additional click sometimes.

After some more testing I found the problem has to do with the focus change of the pane. When focused, the button doesn't seem to work.

I think I need some more time to find the source of the delays. I will get back to you when i have a clean solution. :)

da-h:
One additional thing:
- the last change switched the properties list to be shown inside the pane (`overflow-y: scroll`), instead of overlapping the lower border. I found this fits the current project state better and also is needed as in the multiple-windows-case, some windows could overlap the open list of another window.

js/Pane.js
da-h: Just found the bug in my break!
I did not update the function `shouldComponentUpdate` to include the new button. :facepalm:

The triggers now work as expected. :muscle:

da-h:
Final Note:
- I found some minor bugs due to missing integrity tests of the objects to create property lists from. I have added some minor checks before passing the objects to `AbstractPropertiesList` in the last push.

js/AbstractPropertiesList.js
JackUrb: Agreed - not important for this PR, but may make working in this codebase easier for you and everyone moving forward!

js/Pane.js
JackUrb: Ah okay this makes a lot of sense - it's been so long since I've worked with `shouldComponentUpdate` flows that I didn't even think about it. Thanks for chasing this bug down!

JackUrb (approved):
Thanks for the fixes, and for the overall feature here! This is a really cool addition to Visdom and I'm sure it'll see use 👍

Issue #487, opened 2018-10-11
Feature request -- Compare images from different environments


lavoiems:
**Is your feature request related to a problem? Please describe.**
I like the feature that compare the plots from different environments. I'd like to see a similar feature for comparing images.

**Describe the solution you'd like**
Creating a new window that would contains the images from the selected environments. A second window containing the absolute position of each image in the new window would also be needed.

**Describe alternatives you've considered**
Downloading the images and comparing them side by side

JackUrb:
Seems like something that would be valid to implement if you're (or someone else is) interested. Plot comparisons are created in https://github.com/facebookresearch/visdom/blob/635897f47470961791f907c80742ae4b901ddf6f/py/visdom/server.py#L765
I'm unsure a separate pane with the absolute positions would be needed though, it might be possible to just create some kind of MultiPane that renders two ImagePanes inside. Basic idea would be that if the plot type is an image, the `compare_env` function would package up the data to display that multipane and then it would be rendered.

vedansh-5:
Raised a PR #1204 for this issue.

Issue #477, opened 2018-09-18
Adding a new plot from existing plot in the ui itself


TheShadow29:
Suppose I already have two different plots, say of Metric1 and Metric2. Now I want to plot both of these in the same graph. To my knowledge, one would have to do this in the python file itself. However, it would be nice to have a quick comparison between the two graphs without having to change the python code

A simple add_graph like option in visdom gui should suffice and taking values from the two different graphs.

rajatvd:
I've written a [package](https://github.com/rajatvd/VisdomObserver) which somewhat accomplishes this, but it relies on having a `metrics.json` file from which to plot graphs. It also integrates with the `sacred` package. I'm currently working on trying to make it a bit more independent from this. 

If this doesn't fit your needs exactly, you could try using the `properties` pane and write up a callback to do what you require. Have a look at the `demo_properties.py` [example](https://github.com/facebookresearch/visdom/blob/master/example/demo-properties.py) if this isn't clear.

Another approach is to have the two plots in different envs with the same window title, and using the compare envs feature to plot both the metrics on the same graph.

TheShadow29:
Seems like properites_callback is worth digging into. The compare env wouldn't be good as it would require one to have the same variable name for all the metrics. Your package also looks good, I will definitely refer to it. Thank you. However, I still believe this is something that should be supported natively in visdom.

TheShadow29:
@JackUrb Thank you for the detailed roadmap. Definitely seems worth implementing. I can give it a try however, I too have a bit of limited bandwidth. I will definitely try but no promises. 

A couple of clarifications:
1. `Add Chart` option in the UI should be good enough. It should have the option to select which variable names to plot with checkboxes.
2. I am not sure what `ComparePane` is. Is it the name of the new pane to be created?
3. Did you mean `main.json` instead of `main.js`?
4. I am having a bit of trouble understanding the terms `events`, `trace`, `layout`. I am guessing (please correct me if I am wrong) `events`: is the Callback events namely close, keypress and propertyupdate, and `trace` is a dictionary containing the data points, and has graph properties, finally `layout`is the title of the graph, and axes names. I am not sure where can I get these values from. `layout.json` is in `$HOME/.visdom/view/layout.json`. But where do I get the `traces` from.

Thanks

JackUrb:
[...]

Issue #470, opened 2018-09-01
Display text windows when comparing environments


rajatvd:
Currently, comparing environments compares only plots with the same title across envs. It would be useful to have a similar feature for text windows. In my setup, I create an env for each run of my experiment, and display some plots and a text windows with some config info for each run. Comparing envs is an awesome feature which allows me to easily compare the plots across different runs. However, I can't easily compare the configs across runs.

A solution is to just display all text windows with the same title, and append an index (0, 1, 2,...) in the title of the compare env window as is the case for legends in plots.

It might be possible to simulate this behavior by displaying the configs in a 'plot' using a text annotation of some sort, but this seems too hacky.

Pcmhacker-hero:
Hi @vedansh-5 ,
I took a look at issue #470, and was able to reproduce the behavior against the current dev branch. Currently, compare_envs() ignores text windows with the same title, although the frontend already supports displaying text windows in comparison views. If this is still actual I would like to work on this issue. 

Would it be alright for me to take it on?

PR #840, opened 2022-03-07
Compare everything with everything else

da-h:
## Description
This PR enables to compare any pane with any other pane using the same title.
(See video below).

## Motivation and Context
Visdom's compare feature is one of its selling points, but unfortunately is only supported by scatter plots.
There are issues asking for such a feature, i.e.
* #470 
* #487
  (Works even with image-sliders enabled).
* In principle, the new implementation can easily be extended to also tackle #477.

This PR tries to increase the usability of this feature by enabling to compare any two window contents that share the same title.  
*(This design decision is due to its previous behavior, where scatter plots were merged based on their title.)*

**In more detail**:
- Essentially, the `compare_env` from `server.py` has been rewritten completely, mostly falling back to the previous behavior in case of scatter plots.
- This also solves a minor bug: when the first `env` in the compare-list did not contain a window a later env did contain, the window would not show up. The new behavior shows a window if any `env` contains it.

## How Has This Been Tested?
Added a `title` to every plot in my `demo.py`-copy. However, I strongly think that this feature should be tested thoroughly before merging. ;)

(The textbox says: "Actually, the new feature can change any content to any other content using the same title.")
record2

## Types of changes

## Checklist:


JackUrb (commented):
HI @da-h - this is an interesting feature, but it's also a fairly complicated one to get right. I really like the eager approach here to get all of the pane types functioning and represented, though some notes on a design that could be considered:
It could be reasonable to create some kind of abstract `ComparePane` which would be fed data when in compare mode, and then the `ComparePane` can contain both specialized behavior (treating scatterplots in a merge way, perhaps images in side-by-side), alongside default behavior (either what you've implemented currently, or side-by-side as well). `compare_envs` in general is in need of a refactor, so if design changes are required for something like this I'm happy to work through them (and ensure everything is tested!). What are your thoughts here?

da-h:
Thanks the feedback!

I too thought about a side-by-side feature as well, not only for images, but for arbitrary plots as well. I am yet unsure how to implement this in a fitting way: by including a global switch for side-by-side vs. tabbed view or by giving every Pane a button to switch between those two modes, what are the defaults, etc.

I am also unsure in which way to go here abstractly. I see two possibilities
1. As of now, you are right, this feature should be a part of a specialized `ComparePane` that handles the content of choice, possibly merging data if possible. This would keep `Pane` more clean and more importantly, the new Component could use its own state to keep track of the current view. (Currently, the main-file remaps the data into the respective Panes.) The only downside, as far as I can tell, is that the order of windows would change after re-initializing all `Panes` as `ComparePanes`, as currently, the content gets replaced in already available instances. (Is my guess correct?)
2. After thinking about this a bit longer I see another more intrusive suggestion to implement this feature into the project, but surely beyond the scope of this PR. The idea is to make the "Compare Mode" the default mode in which data is presented, thus making the `Pane` just a view for a selection of data. This would broaden visdom's "dashboard"-like appearance to enable the user to *connect* arbitrary plots & windows into a set of views that represent the user-selected data. In the default "view"/case, it would show the data just as visdom does now, namely show every plot in a seperate Pane. The new feature would, however, allow a user to connect any selection of plots to be represented in the same Pane. I see allowing the user to group & show exactly the information that is needed at the respective moment as an extension of the already existing functionality used right now by the current compare feature, which is currently only limited to group the same plot of different environments.

In any case or outcome of the discussion, I am happy to implement a concrete proposal that takes backward compatibility into account and resubmit the PR.

JackUrb:
Sorry for the delays on the continued review here - this is a hard one to navigate, especially when considering things like how it interacts with some of the other quality features you're adding (like #837 or #846). 

I think generally option (1) is the easier route, and that ultimately combining it with some of the other functionality we already have (like saved views) then it provides a similar type of capability that (2) would provide, without the headache of a substantial refactor. 

I can outline a clearer sample architecture next week for what I imagine this would look like if you'd like to pursue that!

da-h:
> I think generally option (1) is the easier route, and that ultimately combining it with some of the other functionality we already have (like saved views) then it provides a similar type of capability that (2) would provide, without the headache of a substantial refactor.

This is true. Also, (1) could lead a more gradual route to (2) with the possibility to change the defaults later (i.e. allowing every Pane to be comparable or so).

> I can outline a clearer sample architecture next week for what I imagine this would look like if you'd like to pursue that!

Sure, your suggestions are very welcome! ;)
Where i am particularly unsure in the context of react is how to enable each subclass of `Pane` to be comparable using the new `ComparePane` that extends Pane itself. Presumably you would have to create `ComparePane` as a wrapper for another `Pane`-Object, right? But how would the function `compare_envs` trigger the switch of the classes (`Pane` to `ComparePane`) ?

Saksham-Sirohi (commented):

- Before review for merge, please split or mark dependencies on the server refactor stack (#675 / #1384).

PR #842, opened 2022-03-08
Custom Machine- & Project-specific Styles

da-h:
## Description
This PR adds the possibility to style visdom on a per-user or per-project fashion.

**In more detail**:
- the style files are loaded during the server initialization to keep load-times low
- for windows & osx the phat changes to their platform-specific config paths

## Motivation and Context
In my case, I use visdom on multiple machines, sometimes even in multiple `env_paths` at the same time.
Specifying visually seperating the machines & projects can help to find more quickly what I was looking for.

## How Has This Been Tested?
- create user specific css `~/.config/visdom/style.css` & adapt
- create env-path specific `style.css` & adapt

2022-03-08-090919_1618x792_scrot

## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/server.py
+++ b/py/visdom/server.py
@@ -23,6 +23,7 @@ import time
 import traceback
 import uuid
 import warnings
+import platform
 from os.path import expanduser
 from collections import OrderedDict
 from collections.abc import Mapping
@@ -204,6 +205,7 @@ class Application(tornado.web.Application):
         self.env_path = env_path
         self.state = self.load_state()
         self.layouts = self.load_layouts()
+        self.user_settings = self.load_user_settings()
         self.subs = {}
         self.sources = {}
         self.port = port
@@ -243,6 +245,7 @@ class Application(tornado.web.Application):
             (r"%s/win_hash" % self.base_url, HashHandler, {'app': self}),
             (r"%s/env_state" % self.base_url, EnvStateHandler, {'app': self}),
             (r"%s/fork_env" % self.base_url, ForkEnvHandler, {'app': self}),
+            (r"%s/user/(.*)" % self.base_url, UserSettingsHandler, {'app': self}),
             (r"%s(.*)" % self.base_url, IndexHandler, {'app': self}),
         ]
         super(Application, self).__init__(handlers, **tornado_settings)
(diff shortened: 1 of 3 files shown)


py/visdom/server.py
```diff
@@ -319,6 +322,36 @@ def load_state(self):
...
 
+    def load_user_settings(self):
+        settings = {}
+
+        """Determines & uses the platform-specific root directory for user configurations."""
+        if platform.system() == "Windows":
+            base_dir = os.getenv('APPDATA')
+        elif platform.system() == "Darwin": # osx
+            base_dir = os.path.expanduser('~/Library/Preferences')
+        else:
```
JackUrb, line 333: What is this else case - general unix users?

py/visdom/server.py
```diff
@@ -319,6 +322,36 @@ def load_state(self):
...
+            base_dir = os.getenv('XDG_CONFIG_HOME', os.path.expanduser('~/.config'))
```
JackUrb, line 334: I'm not familiar with `XDG_CONFIG_HOME`, is this a standard directory?

py/visdom/server.py
```diff
@@ -319,6 +322,36 @@ def load_state(self):
...
+        user_css = ""
+        logging.error("initializing")
+        home_style_path = os.path.join(config_dir, "style.css")
+        if os.path.exists(home_style_path):
+            with open(home_style_path, "r") as f:
+                user_css += "\n" + f.read()
+        project_style_path = os.path.join(self.env_path, "style.css")
+        if os.path.exists(project_style_path):
+            with open(project_style_path, "r") as f:
+                user_css += "\n" + f.read()
```
JackUrb, line 347: Would it ever be the case where we'd want to combine both of these styles rather than directly overwrite? Not necessarily that we need to support this, but proposing the thought. If we decide against it you should include a note about this behavior in the docs.

JackUrb (commented):
This is a cool feature! I have some questions, but this kind of customization I'm sure many will find useful. Want to get in a round of discussion before merging though.

py/visdom/server.py
da-h: Of course, both options (extending and replacing style-file) don't exclude each other. However, I am unsure, if this makes the feature too complicated.

Let's run through the idea:

-----

In case of both features, I think, the best search order for the style to serve is:
1. replacing style-file for current env-path
2. replacing style-file for user
3. default style-file with extending style-file, first, by extending using the user style-file, then by extending using the env-path style file.
   (Currently, this PR implements only this option.)

What would be the naming scheme?
1. `style.css` for replacing style-files
2. `extend.css` for extending style-files

----

What do you think?

py/visdom/server.py
da-h: Yes, AFAIK this comes from X-Applications initially, but many CLI-tools (e.g. `git` `curl`, `vim`, ...) use the environment variable as well. (See [here](https://wiki.archlinux.org/title/XDG_Base_Directory#Support) for a more complete list.)

py/visdom/server.py
da-h: Just double checked this. `platform.system()` typically returns one of basically three OS Types.

The documentation of `platform.system()` says:
> Returns the system/OS name, such as 'Linux', 'Darwin', 'Java', 'Windows'. An empty string is returned if the value cannot be determined.

[Digging a bit deeper on the pros/cons of `sys.platform` vs `os.name` vs `platform.system()`](https://stackoverflow.com/questions/1854/python-what-os-am-i-running-on/58071295#58071295) shows that `platform.system()` checks the system type at run-time, while the other get their value at python compile-time.

On any system, using `~/.config` should work, however specifying another config-directory for windows and osx seems to me to be a frequently used approach in applications.

py/visdom/server.py
JackUrb: Ah thinking about this more with your context, your version is already best inline with CSS standards - repeated rules already take the later declaration, so you're set up in a case where overriding will occur in the correct direction here. If a user really wants to completely override they're free to do so. Thanks for enumerating here though!

py/visdom/server.py
JackUrb: Thanks for clarifying! I should probably be using this for my other CLI tools as well then 😅

JackUrb (approved):
Thanks for clarifying on this @da-h - happy to have this in the codebase!

PR #844, opened 2022-03-21
Action to build js files automatically

da-h:
## Description
This PR moves the requirement to commit build files from the contributor to a dedicated github action.

## Motivation and Context
The goal of this PR is to ensure consistency across builds on the one hand, and to increase security of the deployment on the other hand.

## How Has This Been Tested?
See the action result in [my own test branch](https://github.com/da-h/visdom/actions/runs/2014654689)

## Notes
The new action triggers on any pushes of the `js/**` files. Alternatively, it would be reasonable to either
- trigger on pushes onto `master` only
- trigger on closing of merge requests (I see no direct possibility to trigger on "merged" merge requests)

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/update-js-build-files.yml
+++ b/.github/workflows/update-js-build-files.yml
@@ -1,0 +1,29 @@
+name: Update Static JS Files
+
+on:
+  push:
+    paths:
+        - 'js/**'
+
+jobs:
+  update-static-js-files:
+    runs-on: ubuntu-latest
+    name: "Update Static JS-Files"
+    steps:
+      - uses: actions/checkout@v3
+        # > if the push needs to trigger other workflows, use a repo scoped Personal Acces Token
+        # see: https://stackoverflow.com/questions/57921401/push-to-origin-from-github-action/58393457#58393457
+        # with:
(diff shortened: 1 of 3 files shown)


JackUrb (approved):
This is a great quality-of-life improvement, thanks for contributing to the dev-ops here @da-h! Only possible improvement I can see is having the action visibly fail if `npm install` or `npm run build` were to fail, but I'm happy to pull this in as-is.

da-h:
Quick question on the suggestion for improvement: what do you mean exactly? Only resource I can find is the [use of `${{ failure() }}` in an action to save build failure logs](https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-runner-context). I am somewhat at a loss here. :sweat_smile:

JackUrb:
To some degree I'm not sure if it would already happen (that the action would display a failure) if you were to make a mistake. Can be tested by putting a package that doesn't exist into the `package.json` (to test `npm install`) and then making an unparsable syntax error in the js code (to test `npm run build`).

da-h:
Indeed, this is the case.
I just double checked this for both mentioned scenarios:

image

PR #846, opened 2022-03-21
Interactive Smoothing of Line Plots

### LINKED ISSUE #18: Feature Request: Improved line plotting functions with interactive smoothing
I think it would be good if it could support better line plotting like with tensorboards writer.add_summary summary.scalar:

1) append single data points at a time
2) interactive smoothing slider for line plots ( e.g. with https://plot.ly/python/slider-widget/ )
3) transparent view of non-smoothed data 
4) error envelopes ( https://plot.ly/python/continuous-error-bars )
5) logging of relative/wall time simultaneous to global steps in x-axis

da-h:
## Description
This PR adds interactive smoothing, similar in spirit to how tensorboard allows noisy line plots to be shown smoothed.

## Motivation and Context
This is a long-awaited feature on my part, and i'm sure the broader public will be interested as well. (Possibly closes #18 as well).
This PR adds a button to the top-right corner of every `PlotPane` that shows at least one set of lines.

**Technical notes**:
- With #837 `js/Pane` adds another button in the top-right corner. This PR rewrites this part by adding the property `barwidgets` in order to also allow multiple buttons to be shown.
- The smoothing algorithm used is savitgzy-golay. This PR uses the node module [savitzky-golay-generalized](https://github.com/mljs/savitzky-golay-generalized) (MIT licensed), thus the requirement has been added to the `package.json`.

**Note**:
This PR is a draft as it builds upon #837, which is not merged yet.

## How Has This Been Tested?
Just tested using some of the `demo.py` line plots.

record

## Types of changes

## Checklist:


da-h:
Small note on the side:
To let the smoothed and non-smoothed lines to share the same color one currently needs to specify the color beforehand.

It would be nice if this could be done automatically, however, the actual used color cannot be extracted at the time needed as far as I can tell. Is there any possibility to let the PlotPane query the color of the "default" line plots?

da-h:
Last push changes the following:
- changed the selection text from "Selected" to "Smoothing"
- adapted the color of the `~`-button to match the other button colors

JackUrb:
Awesome to see a PR coming in and attacking one of our earliest feature requests!

> To let the smoothed and non-smoothed lines to share the same color one currently needs to specify the color beforehand.

So you're saying that, unless you provide a value for the `markerColor`, we can't extract a color? That is to say when `marker.d.color` is `null`, plotly is pulling default colors only _then_?

From experimentation, I believe that plotly determines this based on the index of the trace into the list of traces, and determines the color based on that index mod 10 into d3's base colors:
https://github.com/plotly/plotly.js/blob/master/src/components/color/attributes.js

da-h:
Was exactly what I was looking for, thank you! :muscle: 
* I have now adjusted the coloring so that the color matches correctly both in the case of given colors and in the case of standard colors by plotly. (It has been solved by just adding empty objects to the list until it contains a factor of 10 objects. That way plotly uses the same coloring for the smoothed versions of the lines).
* Also, in the latest push, I have removed the legend entries for the smoothed lines.

JackUrb (approved):
Padding was a really clever solution - _ideally_ we'd be able to pull the length of plotly.js's default color list to get the pad modulus value, but short of going that far I think this solution is really nice.

Thanks for tackling yet another feature, and delivering a high-quality PR for it as well!

PR #848, opened 2022-04-14
Fix demo.py by correcting an argument-assert of vis.heatmap

da-h:
## Description
This is just a small bugfix in `demo.py`.

In #834, i have added the `update` flag for histograms. However, I also introduced an error in `demo.py`, because I have added a wrong argument check right before pushing the feature. :facepalm: 
This PR solves the issue by replacing the check for `"reset"` with the correct value `"replace"`.

*A side note*: this (`replace`) is also specified in the documentation and matches the naming in `vis.line`.

## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -1746,7 +1746,7 @@ class Visdom(object):
         - `opts.nancolor`: if not None, color for plotting nan
                            (`string`; default = `None`)
         """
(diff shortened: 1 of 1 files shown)


JackUrb (approved):
Thanks for the identification, and more importantly, the fix 👍

PR #835, opened 2022-03-02
Apply webpack best Practice for better Development

da-h:
## Motivation & Description
The current `webpack.config.js` does not enable source maps by default, making it rather hard to debug the JavaScript code.
I took this opportunity to enable source maps by also applying the [current "best practice" for webpack](https://webpack.js.org/guides/production/).

**Note**:
I have removed the `npm buckbuild` script from `package.json` as it appears nowhere else in the project and is just an alias for `npm build`

## How Has This Been Tested?
I used it already for debugging. Functionality-wise it should not alter anything besides source maps in the browser.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] For JavaScript changes, I have re-generated the minified JavaScript code.
  -> Technically the binary differs due to source maps added. I was unsure if I shall also commit the new compiled file.
 
**PS**: In case you are interested in a release-bot that also compiles the files accordingly after merging a PR, I can offer some simple github actions/workflows that could tackle that.

PR #849, opened 2022-04-14
Update most npm dependencies (Babel, Webpack, Eslint, Prettier, ...)

da-h:
## Description
In order to bring visdom up to date and to promote new developments, this PR is about updating the dependencies of the project.

## Motivation and Context
Following the suggestions in #837 I have rewritten most of the react components to match the newer hook-based react templates. *(PR coming soon).*
I felt that splitting up the "meta" part from actual implementations is cleaner and easier to review, thus this separate PR.

This PR solves most of the *dependency warnings* when doing a fresh install, i.e. `rm -rf node_modules package-lock.json && npm install`.

**In more detail**:
* Upgraded Babel 6 to **Babel 7** by follwing the [Migration Guide](https://babeljs.io/docs/en/v7-migration)
* Upgraded Webpack 4 to **Webpack 5** by follwing the [Migration Guide](https://webpack.js.org/migrate/5/)
* Upgraded Eslint 5 to **Eslint 7**
* Upgraded Husky 1 to **Husky 7** & lint-staged 8 to **lint-staged 12**
* Upgraded Prettier 4 to **Prettier 8** & applied the new prettier flavor to all `*.js` files
   (Note: This mainly just adds brackets in some places)
* Upgraded react-devtools 3 to **react-devtool 4**
* Upgraded **all other Modules** using `npm update --save`
* Fix `threesjs` to last known working version `r105`.
  Threesjs does not use SemVer. [There has been several api changes in threesjs](https://github.com/mrdoob/three.js/wiki/Migration-Guide) & the version has been updated automatically without any change of the respective `EmbeddingPane`-code.
* Removing absolete `webpack.config.js` (as this has been replaced by `webpack.prod.js` and `webpack.dev.js` in #835.

## How Has This Been Tested?
With every PR, i test all touched functions by pressing all buttons and switches. Also I typically use each feature for some days before I publish the PR. Nevertheless, visdom would certainly benefit from a test suite for just such changes.
Since i contribute to the project only as a hobby and i don't come from the web corner, the node-ecosystem is quite new to me. In case of UI changes, a ui regression test-framework sure would certainly be helpful.
**Is there possibly a clear choice, such as a widely used package, that is particularly obvious for a project like visdom that uses react?**

## Note
Especially considering the upcoming "hook-based react"-PR, a new branch could be considered for this PR until some unit-tests are in place. On the other hand, the subsequent code-changes are far-reaching enough that one would not want to repeat them in case there are further changes on the main branch in the meantime. What is your opinion on this?

## Types of changes
As far as I can tell, no features have been changed.

## Checklist:
  `npm lint` does throw some errors, which will however solved in the follow-up PR. Compilation & running the demo however works.


JackUrb:
Wow I just went through a dependency update for the Mephisto project recently - these are herculean efforts at times so big thanks @da-h for getting this together. 

> Is there possibly a clear choice, such as a widely used package, that is particularly obvious for a project like visdom that uses react?

Agreed that it has been difficult to make frontend changes without this sort of tooling set up. As such this is a good question. @pringshia do you have a suggestion here? I could certainly open an issue with anything you suggest.

> On the other hand, the subsequent code-changes are far-reaching enough that one would not want to repeat them in case there are further changes on the main branch in the meantime. 

At the moment @da-h I think you're the most prolific visdom contributor, I'd be happy to defer to whichever approach you're more comfortable with. I tend to err towards large feature branches in projects without many incoming contributions however, it makes it easier to track the whole thing and prevents interim potentially broken states. That being said, if you've run through all of the plot options in the demo and tried some switches, that's about as much as we were doing on the project getting this far.

pringshia:
Thanks for these changes! Your contributions are much appreciated.

With regards to testing, given the highly interactive and visual nature of visdom, I would suggest end-to-end integration testing using something like [Cypress](https://www.cypress.io/).

For more simple React tests, I would suggest `react-testing-library` with `jest` as the test-runner, but I don't think these will be as valuable of a return-on-investment as Cypress tests would be.

da-h:
I am glad to see that this first real test of the CI passed.

One thing is a bit confusing: because I have forced-pushed the branch, the result is a bit hidden due to the fact that the static-js-bot has issued a commit after the workflows are started.

I think there is a simple fix for that: by **setting the js-bot to only update static/js files on pushes to master**.
The bot would trigger after the PR has been accepted, build errors are catched using the cypress ci tests.
Probably this is cleaner anyways in case of multiple parallel PRs.
On the other hand, if a branch is created first and the PR next (which is the used case), the "update static/js files" commit is the last one already, and thus, the action should indeed be visible.

Let me know what you think.

JackUrb:
I imagine we'd want the cypress tests to run on the branch _after_ the `update static/js files` commit, as those changes may change the outcome of the test, correct?

da-h:
The test results should not change, because the workflow compiles new `static/js files` anyway as a first step. Thus, the outcome should not change at all. My suggestion was purely cosmetic:

Currently, the test result is shown in the list of commits as above using a green checkmark.
image

However, if the auto-generated commit were not there, and the commit on which the tests are started is the last one in the list, the state of the workflows are presented in a clearer way.
image


**As I said, it's a purely cosmetic and probably also not worth the hassle.
Sorry the confusion. :)**

JackUrb:
Ah okay sorry - I missed the step where the test autogenerates the new static/js files. I think the visibility is actually an important thing to bring up (as we've had issues with it in other repos). In theory, as the cypress test requires this generation, we'd already be testing it as part of this run. As such I'd be in favor of running the generation on merge to master.

da-h:
Ok, thx the feedback. I'll create a new PR for that change.
(Should be a one-liner actually).

da-h:
Removed the `update static/js files` commit to match the new workflow.

JackUrb (approved):
With this now tested, I'm happy to pull it in. Thanks for going through the additional legwork to make these types of changes easier to execute in the future @da-h!

PR #853, opened 2022-05-04
Add 'change-properties' Cypress Test

da-h:
## Description
This PR adds a simple cypress test that tests the function to change any property used in any `PlotPane`. (I.e. the function from #837).

## Motivation and Context
During the rewrite of most react components using the hook based react-api, I did encounter unexpected changes in this function, thus the suggestion to add a test for this.

Note:
This PR will fail, as the test does not exist on the `master` branch, yet. (#854 solves this).

## How Has This Been Tested?
I've tested this PR already [on my own fork](https://github.com/da-h/visdom/pull/9), including also #854 .

## Screenshots:
The test uses the `plot_line_basic` example:
plot_line_basic

and changes some properties, resulting in:
change-properties

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
      Added a new visual regression test
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #854, opened 2022-05-04
Enable PRs That Change Cypress Visual Regression Tests

da-h:
## Description
This PR fixes a small issue with the current cypress testing workflow related to PRs that modify these tests.

## Motivation and Context
The visual regression tests in `cypress/integration/screenshot.js` take screenshots of the state *before* the code changes of any PR. In case a new test is added in such a PR, this procedure will always result in a failed test, as on the base branch the new test cases are not present yet.

The solution is simply to take the screenshots on the base branch, but using the new test setup.

Note: Certainly, this procedure can fail in case a breaking API-change occurs in the cypress-package and a PR includes an `npm update` that includes this API-change. However, this should happen much less often than new tests being created.

## How Has This Been Tested?
I tested the change using the PR #853 [in my fork](https://github.com/da-h/visdom/pull/9).

## Types of changes

## Checklist:

      No code changed.

### DIFF
--- a/.github/workflows/process-changes.yml
+++ b/.github/workflows/process-changes.yml
@@ -49,9 +49,20 @@ jobs:
 
       - name: "Checkout Repository"
         uses: actions/checkout@v3
+
+      - name: "Save current Head-Ref as PR-branch"
+        shell: bash
+        run: |
+          git checkout -B PR-HEAD
+
       - uses: ./.github/actions/prepare
         with:
           usebasebranch: true
+
+      - name: "Checkout Tests from Head-Ref"
+        shell: bash
+        run: |
+          git checkout PR-HEAD -- ./cypress
 
       - name: Cypress test:init
         uses: cypress-io/github-action@v2


JackUrb:
Thanks for catching this edge case and moving forward with a fix! One question I have - what do we do for PRs that we expect the regression tests to fail, for instance if we make a deliberate change to the interface?

da-h:
Merging even though a test failed should be possible as long as the option "[Require status checks to pass before merging.](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule)" is not turned on.

Additionally, in the current workflow definition, it is more visible whether the functional test passes successfully while the visual test fails. By inspecting the logs of the failed test, you can nail down which tests fail exactly / whether only the expected tests fail. Also, each test-run saves all screenshots that are made during testing.

However, it would probably be cleanest if the PR-creator could somehow define which sub-tests do not cause errors because they are known to change visually / fail.

PR #856, opened 2022-05-29
[Big change] Migration to React Hooks

da-h:
This PR migrates most React components of this project to the newer [hook based react style](https://reactjs.org/docs/hooks-intro.html).

## Description
This PR applies three main changes to `js` files of the code base:
1. **Migrating most React components** (all but `EmbeddingsPane` and `App`) **to hook based react functions.**
2. **Solving some minor bugs on the way**:
    - added missing `tbody` tags in `PropertyLists`. (The browser console did mention this in its warnings).
    - added missing `key`-attributes in `Pane`. (The browser console did mention this in its warnings).
    - with the new react style the changing-properties-feature stopped working consistently in `Pane` and in `PropertiesPane`.  
      Specifying to call `stopPropagation` only in `PropertiesPane` solved the issue. (However, I do not really understand why. Any idea? :sweat_smile:  )
3. **Code normalization.**
    - *Removing unneeded functions.*  
       I.e. `getWindowSize` and `getContentSize` have not been used in `Pane` & `resize` has not been used in both, `Pane` and `PropertiesPane`.
    - *Rearranging all code blocks in the touched files so that all files have a similar structure in this order*:
        - state variables
        - public events (overwritable by props)
        - private events (not overwritable by props)
        - (other, e.g. the calculations of Image position and scale in `ImagePane`)
        - rendering
    - Especially in the beginning of my work for this PR, I've found myself a bit lost in some quite long functions. Thus, I've *added code comment headings* for easier navigation in the larger files (see `ImagePane`).

**Notes**:
* The third point in the list is actually optional to accomplish the goal of this PR and rather a personal taste of mine to put more structure into some of the longer files. To make this step optional, all changes for this step are in separate commits (those named `js: code normalize X`).  
    Feel free to remove or adapt in case something does not meet the code style of the project.
* I also appreciate any suggestions for improvement, especially considering that this change is rather huge and i'm quite new to react and react hooks.

## Motivation and Context
This change has been suggested in a [previous discussion](https://github.com/fossasia/visdom/pull/837#discussion_r822121979), and I am also sure that future changes would benefit of a more up-to-date code base.

## How Has This Been Tested?
No additional testing. As no functionality should have been changed the cypress tests should succeed.

## Types of changes
- [x] Refactor

## Checklist:

### DIFF
--- a/js/ImagePane.js
+++ b/js/ImagePane.js
@@ -7,88 +7,76 @@
  *
  */
 
-import React from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import EventSystem from './EventSystem';
 import Pane from './Pane';
 
 const DEFAULT_HEIGHT = 400;
 const DEFAULT_WIDTH = 300;
 
-class ImagePane extends React.Component {
-  _paneRef = null;
-  _imgRef = null;
-  _natHeight = null;
-  _natWidth = null;
+function ImagePane(props) {
+  const paneRef = useRef();
+  const imgRef = useRef();
+  const [view, setView] = useState({ scale: 1, tx: 0, ty: 0 });
+  const [imgDim, setImgDim] = useState({ width: null, height: 0 });
+  const [selected, setSelected] = useState(props.selected);
+  const [mouseLocation, setMouseLocation] = useState({
+    x: 0,
+    y: 0,
+    visibility: 'hidden',
+  });
+  const [dragStart, setDragStart] = useState({
+    x: 0,
+    y: 0,
+  });
 
-  state = {
-    lastModTime: 0,
-    scale: 1,
-    tx: 0,
-    ty: 0,
-    selected: this.props.selected,
-    mouse_location: { x: 0, y: 0, visibility: 'hidden' },
-  };
+  useEffect(() => {
+    setSelected(props.selected);
+  }, [props.selected]);
+  useEffect(() => {
+    // Reset the image settings when the user resizes the window. Avoid
+    // constantly resetting the zoom level when user has not zoomed.
+    if (Math.abs(view['scale'] - 1) > Number.EPSILON) handleReset();
+  }, [props.width, props.height]);
 
-  drag_start_x = null;
-  drag_start_y = null;
+  // initialize
+  useEffect(() => {
+    const onEvent = (event) => {
+      switch (event.type) {
+        case 'keydown':
+        case 'keypress':
+          event.preventDefault();
+          break;
+        case 'keyup':
+          props.appApi.sendPaneMessage({
+            event_type: 'KeyPress',
+            key: event.key,
+            key_code: event.keyCode,
+          });
+          break;
+        case 'click':
+          props.appApi.sendPaneMessage({
+            event_type: 'Click',
+            image_coord: mouseLocation,
+          });
+          break;
+      }
+    };
 
-  componentWillReceiveProps = (nextProps) => {
-    if (nextProps.selected !== this.props.selected) {
-      this.setState({ selected: nextProps.selected });
-    }
-    if (
-      (nextProps.width != this.props.width ||
-        nextProps.height != this.props.height) &&
-      Math.abs(this.state.scale - 1) > Number.EPSILON
-    ) {
-      // Reset the image settings when the user resizes the window. Avoid
-      // constantly resetting the zoom level when user has not zoomed.
-      this.handleReset();
-    }
-  };
+    EventSystem.subscribe('global.event', onEvent);
+    return function cleanup() {
+      EventSystem.unsubscribe('global.event', onEvent);
+    };
+  }, [mouseLocation]);
 
-  onEvent = (event) => {
-    if (!this.props.isFocused) {
-      return;
-    }
-
-    switch (event.type) {
-      case 'keydown':
-      case 'keypress':
-        event.preventDefault();
-        break;
-      case 'keyup':
-        this.props.appApi.sendPaneMessage({
-          event_type: 'KeyPress',
-          key: event.key,
-          key_code: event.keyCode,
-        });
-        break;
-      case 'click':
-        this.props.appApi.sendPaneMessage({
-          event_type: 'Click',
-          image_coord: this.state.mouse_location,
-        });
-        break;
-    }
-  };
-
-  componentDidMount() {
-    EventSystem.subscribe('global.event', this.onEvent);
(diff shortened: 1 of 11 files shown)


js/ImagePane.js
```diff
@@ -7,88 +7,41 @@
...
-  componentDidMount() {
-    EventSystem.subscribe('global.event', this.onEvent);
-  }
-
-  componentWillUnmount() {
-    EventSystem.unsubscribe('global.event', this.onEvent);
-  }
-
-  handleDownload = () => {
+function ImagePane(props) {
```
JackUrb, line 17: As a _tiny_ nit, a common react pattern when we know that _this_ component uses specific `props` and the downstream component (that we extend to with `{...props}`) uses different ones is to do parameter de-structuring inline, like:
```
function ImagePane({ known, used, values, ...props})
```
This both clarifies values that this component needs as well as prevents them from being passed down.

In this particular class though, there's a lot of stuff that is being used here and still passed down to the pane, so for these I like doing a de-structure right at the top:
```
const {appApi, height, width, other, args, that, are, needed} = props;
```
This still makes it easy to reason about the required prop arguments upfront.

js/util.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import React, { useEffect, useRef } from 'react';
+
+// custom hook to get previous value of a variable
+function usePrevious(value) {
```
JackUrb, line 13: So if I understand this correctly, this will return the value last provided to this function? And then the `useEffect` ends up triggering _after_ the return, updating the value? 

Are you able to have more than one `usePrevious` across your application? (is `React` that good at tracing callsites for `useRef` to hook to?) If so that's pretty cool!

JackUrb (approved):
Alright, sorry this took so long to get back to @da-h. Overall though, this is an amazing change. Honestly I have an easier time parsing through and reading your new take on the code than I do trying to understand the old code, and I wrote most of the original stuff! (albeit years ago now)

> Specifying to call stopPropagation only in PropertiesPane solved the issue. (However, I do not really understand why. Any idea?)

Are you saying that, if you add the `blurStopPropogation` chain to the underlying `PropertyItem`, but then only set it in `PropertiesPane`, that fixes _both_ the issue in `PropertiesPane` _and_ in the regular `Pane`? That's really weird, and would point me to a difference in event propagation behavior by default in the `Pane`. Might remain something of a mystery...

>  As no functionality should have been changed the cypress tests should succeed.

This is a huge accomplishment really, there was a lot of code here to keep right, so the fact that tests are still passing is great.

js/util.js
da-h: **Actually, yes!**

I was curious and tested if one could create a chain of `usePrevious`-calls to get the penultimate value or the value before that.

I.e. the following code-snippet did work.
```
  const contentID1 = props.contentID
  const contentID2 = usePrevious(contentID1);
  const contentID3 = usePrevious(contentID2);
```

As `usePrevious` calls `useRef` and `useEffect` internally, i would expect the function to follow the same semantic rules as those. (I mean that it should not be called in a loop, etc.).

js/ImagePane.js
da-h: Thanks for pointing this out, I did not know that syntax, yet! :)

**I have adapted all Components to include such a de-construction of all prop-values (if possible) at the start of the function.**

I have added that change as a new commit for easier inspection of the changes.
(If you want, I can squash these changes also into the respective previous commits before merging).

da-h:
Hey there!,

this time it was me who was on the road. :)

> > Specifying to call stopPropagation only in PropertiesPane solved the issue. (However, I do not really understand why. Any idea?)

> Are you saying that, if you add the blurStopPropogation chain to the underlying PropertyItem, but then only set it in PropertiesPane, that fixes both the issue in PropertiesPane and in the regular Pane? That's really weird, and would point me to a difference in event propagation behavior by default in the Pane. Might remain something of a mystery...

It seems so. I would have expected that both Types (`PropertiesPane` and `Pane`) would not change the behavior of the respective Child-Elements that actually set values (i.e. the `PropertyItem`s). Maybe there is a special case somewhere in `main.js` checking the parent class that I have not yet seen. However, the current constellation now seems to work consistently, and the comment at that line of code suggests to future contributors that any change of this function possibly needs additional thoughts. :)

da-h:
This is change. The test on the current master branch (without the changes of this PR) has failed. Let's see if I find something.

js/util.js
JackUrb: I'm learning something new every day here too - this is a cool functionality!

da-h:
Ah too bad, thought the "update to master" button would rebase and not create a merge commit.

da-h:
Strange thing that the test fails again, especially because I have tested both this PR and the fix on my github fork and also locally without any failing test.
I will look into that and propose anotherr fix tomorrow.

JackUrb:
> Ah too bad, thought the "update to master" button would rebase instead of creating a merge commit.
> 
> Too keep things clean, I can remove the merge commit tomorrow an rebase the changes. (Unfortunately, i do not find a possibility to do that on the road).

No worries here, I've been able to keep track!

da-h:
Hey, just found the errors:

**short summary**:
- i messed up the cypress initialization by forgetting to set an important flag (see #866)
- there were three little differences to the old implementation that caused the rest of the errors:
    - barwidgets in the panes have been copied on every click, so the cypress error was indeed legitimate here
    - the line smoothing slider did not show the chosen value before the code refactoring due to a typo (i adapted to match the previous behavior for now)
    - the Property Lists have been missing tbody tags. adding those resulted in a slightly different formatting. (i adapted the css file to match the previous formatting as well)

**On the consistency of the tests**:
I have also tested the consistency of the cypress tests on the way by resubmitting the exact same PR multiple times to my fork.
I found that two tests (`misc_video_tensor` and `image_save_jpeg`) did fail once in 50 tests or so. Possibly we should add a short sleep-command there to make this consistent across all testing machines where loading times may differ slightly.

(In the example image below, all branches test the exact same changes. Here, the `image_save_jpeg` has failed in test 4).
Screenshot 2022-09-17 at 10-26-20 Consistency Test (fix hook_based_react) 4_10 · da-h_visdom@d940252

JackUrb:
Another great QoL change for the codebase here, thanks again @da-h!

Issue #861, opened 2022-09-01
refactoring the data model to enable automatical saving


shouldsee:
**Is your feature request related to a problem? Please describe.**
Ex. I'm always frustrated when Visdom crashed and all my data lost. current solutions call `vis.save` to create json snapshots but this sounds expensive. 

**Describe the solution you'd like**
- Automatically saving env data to disk on a incremental basis.. when I call `vis.scatter` to create a plot, the server writes the received data
immediately to disk.
- alternatively, server caches data on a periodically basis.
- open to discuss whether it should be immediate or periodic

**Describe alternatives you've considered**
Not sure what's the best impl yet.

JackUrb:
Hi @shouldsee - you may be interested in the [`log_to_filename`](https://github.com/fossasia/visdom#:~:text=log_to_filename%3A%20If%20not%20none%2C%20log%20all%20plotting%20and%20updating%20events%20to%20the%20given%20file%20(append%20mode)%20so%20that%20they%20can%20be%20replayed%20later%20using%20replay_log%20(default%3A%20None)) argument, which automatically logs all outgoing plots to a file such that the log can be replayed later if the server crashes.

shouldsee:
> Hi @shouldsee - you may be interested in the [`log_to_filename`](https://github.com/fossasia/visdom#visdom-arguments-python-only) argument, which automatically logs all outgoing plots to a file such that the log can be replayed later if the server crashes.

Sounds cool. This is essentially a transaction log stored to the file. I am more interested in the stateful persistent storage.

Manik-Khajuria-5:
This issue has been solved ref https://github.com/fossasia/visdom/pull/1293

PR #862, opened 2022-09-02
[feat] refactor data storage to allow saving on a per-window basis

shouldsee:
## Description

- changed eager mode logic 
- changed data saving and loading logic
    - old logic in `SimpleJsonStorage`
    - per-window logic in `SimpleWindowJsonStorage`
    - using hard-coded MRO to switch between the logic for now.
- added a utility script for testing
- added auto directory creation
- added DEBUG utility printing

## Motivation and Context

#861 . changing towards a per-window caching scheme

## How Has This Been Tested?

- `rm -rf ./test_data/*`
- start server `python server.py -env_path test_data`
- `python test_simple.py` create plots an save
- closed server 
- reopen server
- check `env:test` is correctly loaded

```
tree test_data/
test_data/
├── test
│   ├── jsons
│   │   ├── 0.json
│   │   ├── 1.json
│   │   ├── 2.json
│   │   ├── 3.json
│   │   └── 4.json
│   └── reload.json
└── view

```

## Types of changes

## Checklist:


js/TextPane.js
```diff
@@ -20,7 +20,7 @@ class TextPane extends React.Component {
     switch (e.type) {
       case 'keydown':
       case 'keypress':
-        e.preventDefault();
+//        e.preventDefault();  /// this would prevent entering text into input box
```
JackUrb, line 23: IIRC this was trying to prevent text panes from submitting on enter. Plus, the goal is to send keypress events to the backend, not necessarily to have them render up-front here.

py/visdom/Makefile
```diff
@@ -0,0 +1,21 @@
+clean:
```
JackUrb, line 1: I'm somewhat against adding _yet another_ scripting flow to the visdom codebase. This should either be part of the `package.json` scripts, or we should be having a discussion about the alternate we should be using instead.

py/visdom/__init__.py
```diff
@@ -576,12 +577,12 @@ def on_message(ws, message):
                     try:
                         handler(message)
                     except Exception as e:
+                        import traceback
+                        tbmsg = traceback.format_exc()
                         logger.warn(
-                            'Visdom failed to handle a handler for {}: {}'
-                            ''.format(message, e)
+                            'Visdom failed to handle a handler for {}: {}.'
+                            '\n traceback:{}'.format(message, e, tbmsg)
```
JackUrb, line 584: This is a nice-to-have addition

JackUrb (commented):
Hi @shouldsee. Thanks for putting up a PR. I'll admit, this one is pretty difficult to review for a number of reasons.
- Class naming here is somewhat obtuse. `B`, `NIF`, etc
- Without explanations for these classes both inline and in some kind of overall explanation, it's hard to follow your intent and methodology.
- Overall, these additions should exist outside the `server.py` file, which is already bloated (and part of why we're trying to refactor).
- Addition of `Makefile` is complicating the devX here.
- Last, there are some comment-ed out changes (like in `demo.py`), which we really shouldn't be maintaining in the codebase, alongside conditionals such as `if 1:`

I'm open to the feature change you're proposing here, but I'm definitely not ready to fully review, let alone accept, in the current state. Would love to see it all in a more final state.

js/TextPane.js
shouldsee: Ah I see. I was putting forms into visdom window, and .preventDefault would make the form inputting fairly laggy, hence commented out.

JackUrb: Perhaps there's value in a new `FormPane` component, likely a parent class for the `PropertiesPane`. This would then be able to accomplish what you're attempting in #868, but in a cleaner way

tonypzy:
The server has since been refactored into a module structure (py/visdom/server/), making this PR's diff no longer applicable. The per-window storage concept has good sides but it needs a fresh implementation against the current architecture. Closing after 2+ years of inactivity.

PR #863, opened 2022-09-14
Cypress Click-Event Hotfix

da-h:
Fixes cypress test fails (possibly triggered) due to reappearing windows during re-rendering.

## Description
In the recent MR #856, the initialization of the cypress tests failed unexpectedly.  (See [these error logs](https://github.com/fossasia/visdom/actions/runs/3054697717/jobs/4926953999).  

The test prepares screenshot on the `master` branch, which has been passing successfully in a previous PR test run. I am honestly a bit unsure why this happened, and, I cannot reproduce the error, yet.  

My guess is that during re-rendering due to a change in the react state, cypress has registered the replaced/recreated window as a new object and thus registered two windows instead of only one.
The error message suggests adding a `multiple: True` flag in case multiple clicks should be performed.
I see no problem in enabling this in case such a coincidence happens again, thus this PR.

**Note**:
Also, the auto-formatter has applied some changes to the tests, which I have committed as well.

## Types of changes

## Checklist:

### DIFF
--- a/cypress/integration/screenshots.init.js
+++ b/cypress/integration/screenshots.init.js
@@ -65,7 +65,7 @@ describe(`Take screenshot for PlotPane functions`, () => {
       seed: 43,
     });
     cy.open_env(env1);
-    cy.get('button[title="smooth lines"]').click();
+    cy.get('button[title="smooth lines"]').click({ multiple: true }); // should be single element, concurrency (?) has triggered error here
     cy.get('input[type="range"]').then(($range) => {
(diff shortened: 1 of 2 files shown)


da-h:
**Note**: The test `visual-regression-test-init` did not fail here, although it should have performed the exact same steps as #856. Thus, I believe this was a concurrency problem of some kind as described above.

JackUrb (approved):
Thanks for the quick fix here, hopefully this makes cypress more stable.

PR #870, opened 2022-09-28
release version 0.2

da-h:
## Description
As discussed in #845, this PR tests the workflow implemented in #845 by releasing version 0.2.

Merging this PR after #845 should
1. create a github release
2. create & push a git tag for the release commit
3. pushing the new release 0.2 to pypi.org

PR #872, opened 2022-10-04
Fix pypi workflow

da-h:
## Description
This PR fixes the errors and warnings in the PyPI workflow.

## Motivation and Context
The release workflow of #870 should have deployed visdom to PyPI, however it failed.
* Reason was that the `setup.py`-file was missing the content type for the `long_description` field. Setting that to `text/markdown` solves the error that halted execution.
* This PR also resolves a warning that asked to fix the version of the official pypi-github-action that actually deploys the package.

## How Has This Been Tested?
This PR has been tested by pushing a new release to my own fork.
(See [here](https://github.com/da-h/visdom/actions/runs/3185759853/jobs/5195680037) for the result of that workflow, where the checks of the package succeeded, but only the uploading fails due to missing secrets in my fork.).

## Types of changes

## Checklist:


da-h:
*Note*: Re-Triggering the workflow requires to increase a change in the version-number and with that a re-release, unfortunately. Thus, I increased the PATCH/MICRO version number, being the smallest according to semver and [the hitchikers guide to packaging](https://the-hitchhikers-guide-to-packaging.readthedocs.io/en/latest/specification.html). We could however also signal no real change using a `post` suffix as described in the [PEP440](https://peps.python.org/pep-0440/), which *should* be supported by PyPI.

JackUrb:
I think for now just moving up a micro version is enough. If this becomes a regular occurrence we can look into a retrigger flow of some kind (or, for instance, trying the PyPI release prior to the GH release somehow?). Seems overall like a really minor issue at the moment though.

JackUrb (approved):

JackUrb:
[It worked!](https://pypi.org/project/visdom/) Massive thanks @da-h, this is going to be a significant ease-of-use improvement for the developers experience. No more excuses for `master` to diverge super far from what's publicly available.

da-h:
Hooray! What I relief. :D  :tada: 
Big thanks also for your help & patience! :)

PR #877, opened 2022-10-18
Eslint GitHub workflow

da-h:
## Description
I suggest to add a eslint action to the github workflows.

## Motivation and Context
When working myself on the javascript part of this project I have already used the auto-fix feature of eslint, and also tried to match new code to the suggestions provided by that linter. With all those react-plugins and language extensions it seems to me to be a good feature for new developers.

**Changes**:
I thus suggest to add
* `npm run int:fix` to allow editors automatically and devs manually fix minor mistakes in the proposed code by eslint
* a github-action that checks code for style guide-mistakes, to help reviews to focus more on PR-features rather than a code quality control.
* `eslint-plugin-simple-import-sort` to give the import blocks a nicer look (fixes the order automatically as well)

**Other Changes**:
* the auto-sorter has sorted the `package.json`. I have applied it's changes here as well
* I have also applied all auto-fixes by `npm run lint:fix`
* Lastly, the linter errors that could not be fixed, I have adapted according to the error messages. (Most errors did come from missing accessibility-attributes or too-long lines.)

**Discussion**:
While I do find some linters cumbersome at times (i look at you, python linters), eslint seems to me very well thought out.
Especially in combination with an integration of its language-server into an editor, I personally find it quite usable.

## How Has This Been Tested?
Rather only minor changes, thus function should not be changed at all.
More interestingly, the checks of this PR should also include the proposed action.

## Types of changes
- [x] Github Workflows
- [x] Code Refactor

## Checklist:


da-h:
Removed the renaming `lint` -> `eslint` in latest push.
The requirement of renaming was not given anymore, because i switched from `eslint/github-action` (deprecated) to just calling `npm lint` in `actions/setup-node`.

JackUrb (approved):
+1 for linting being an improvment! Though I'm still a fan of `black` for python formatting.

da-h:
Well, we would also need a linting-tool for the python part.
And I didn't know black, so .... is that something we could try for visdom? :)
I'm very happy to try and learn something new.

Issue #878, opened 2022-10-20
Not able to load latest visdom 0.2.2 module


bhavanap12:
By installing the latest visdom 0.2.2 using pip, loading the module raises an error. 
**ModuleNotFoundError: no module names 'visdom.utils'**

Steps to reproduce
1. pip install visdom==0.2.2
2. python3 -c "import visdom"

**Expected behavior**
Clean load of visdom module

da-h:
Hey there, thanks for pointing this out! :)

Did some refactoring recently, but this one should be easy to fix, see the PR linked above.

da-h:
Hi, could you please check the most recent release of visdom `0.2.3`?

rajnisht7:
Closing as this issue problem doesn't exist anymore.Thank You

PR #879, opened 2022-10-20
Add missing __init__.py files & let setup.py find all importable packages on its own

da-h:
## Description
In the recent refactoring in #858 I've missed to tell `setup.py` about the new python directories.
This results in visdom to be effectively broken when installed from `pip` or `setup.py`.

The solution is quite easy:
- adding all missing `__init__.py` files
- tell `setup.py` about the new directories. I have solved this by using the automatic `setuptools.find_packages` method.

This should also solve #878, who has pointed out the error.

## How Has This Been Tested?
Running locally `python setup.py install --user` works now as expected.

## Note
Maybe it would be best to install visdom using `setup.py` for all tests so this does not happen again?

## Types of changes

## Checklist:

       A nothing has changed really, I added a `.post1` to the version.


JackUrb:
The visual regression test is failing as we need to update that to no longer reference `server.py`. We should do a check to see if that string exists anywhere in the codebase before merging.

da-h:
Thx for looking into this. I have submitted a PR to fix this error.

JackUrb (approved):

PR #880, opened 2022-10-20
github-actions: adapt cypress test to new server code

da-h:
## Description
This PR adapts the cypress test workflow to the new server code (for future reference: #858).
Note that it removes the additional checks introduced in #858 (commit 1a34bb6b) to have the test work across the different versions for the server (before and after refactoring).

(The test failed here because of a missing argument for `test`. It should have been `if test -f "py/visdom/server.py"` instead of `if test "py/visdom/server.py"`. Nevertheless, the extra test-case here is not required anymore.)

## How Has This Been Tested?
Tested on my own fork, see [here](https://github.com/da-h/visdom/pull/25)

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/process-changes.yml
+++ b/.github/workflows/process-changes.yml
@@ -69,7 +69,7 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           install: false
-          start: python3 py/visdom/server -port 8098 -env_path /tmp
+          start: visdom -port 8098 -env_path /tmp
           wait-on: 'http://localhost:8098'
           spec: cypress/integration/*.init.js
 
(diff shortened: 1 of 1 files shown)


da-h:
Added a commit to let all tests use the `visdom` command instead of the "local" call `python3 py/visdom/server`.
* The former is what users actually would call.  
   The benefit is that using this version for actual testing would automatically detect errors like in #878.
* The latter is a bit problematic, because the new file structure uses absolute imports of the `setup.py` installed package `visdom`. I.e. running `python3 py/visdom/server` will not work if the code has not been installed.  
  Note that this would not have been a problem for a "single-file" script.

JackUrb:
Fair point - the single file script was able to avoid install, but too unwieldy to continue to maintain. I think the tradeoff is worthwhile.

JackUrb (approved):

Issue #875, opened 2022-10-14
Create `env_path` if it does not exist or give instructions what to do.


da-h:
**Is your feature request related to a problem? Please describe.**
In case one specifies a env-path that does not exist, for instance, `visdom -env_path /tmp/doesnotexist`, visdom exits with a `FileNotFoundError`-Exception.
```
> $ visdom -env_path /tmp/blub                                                                                                            [±master ●]
Checking for scripts.
It's Alive!
Traceback (most recent call last):
  File "/home/dah/.pyenv/versions/3.8.13/bin/visdom", line 33, in <module>
    sys.exit(load_entry_point('visdom', 'console_scripts', 'visdom')())
  File "/home/dah/Projects/visdom/py/visdom/server.py", line 2102, in download_scripts_and_run
    main()
  File "/home/dah/Projects/visdom/py/visdom/server.py", line 2093, in main
    start_server(port=FLAGS.port, hostname=FLAGS.hostname, base_url=base_url,
  File "/home/dah/Projects/visdom/py/visdom/server.py", line 1960, in start_server
    app = Application(port=port, base_url=base_url, env_path=env_path,
  File "/home/dah/Projects/visdom/py/visdom/server.py", line 206, in __init__
    self.state = self.load_state()
  File "/home/dah/Projects/visdom/py/visdom/server.py", line 299, in load_state
    env_jsons = [i for i in os.listdir(env_path) if '.json' in i]
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/blub'                                 
```

I suggest to let visdom try to create the specified folder first.
This problem occurs also when some users run the project in an environment, where the `DEFAULT_ENV_PATH` (`$USER/.visdom`) does not exist. (See #874)

**Describe the solution you'd like**
I see two easy solutions here:
First: try to create the folder first.
I think this is the better approach, as the default env-folder will not exist for first-time users, giving them possibly the impression that the code does not work.

**Describe alternatives you've considered**
The alternative is to give the user the explicit instruction to create the folder before using it, effectively catching Pythons `FileNotFoundError` Exception.

JackUrb:
Automatically creating the folder is definitely the right call. I'm surprised we overlooked it the first time around here.

Issue #874, opened 2022-10-11
cannot run "python -m visdom.server"


Dracilem:
The error is:

/usr/local/lib/python3.7/dist-packages/visdom/server.py:38: DeprecationWarning: zmq.eventloop.ioloop is deprecated in pyzmq 17. pyzmq now works with default tornado and asyncio eventloops.
  ioloop.install()  # Needs to happen before any tornado imports!
Checking for scripts.
Downloading scripts, this may take a little while
It's Alive!
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.7/dist-packages/visdom/server.py", line 2106, in <module>
    download_scripts_and_run()
  File "/usr/local/lib/python3.7/dist-packages/visdom/server.py", line 2102, in download_scripts_and_run
    main()
  File "/usr/local/lib/python3.7/dist-packages/visdom/server.py", line 2098, in main
    eager_data_loading=FLAGS.eager_data_loading)
  File "/usr/local/lib/python3.7/dist-packages/visdom/server.py", line 1963, in start_server
    eager_data_loading=eager_data_loading)
  File "/usr/local/lib/python3.7/dist-packages/visdom/server.py", line 206, in __init__
    self.state = self.load_state()
  File "/usr/local/lib/python3.7/dist-packages/visdom/server.py", line 299, in load_state
    env_jsons = [i for i in os.listdir(env_path) if '.json' in i]
FileNotFoundError: [Errno 2] No such file or directory: '/root/.visdom/'

I searched a lot, but no solution is related.

da-h:
Hey there,
could you both check if you run the most recent version, `0.2.1`?
- The errors show different line numbers here than where these lines currently are in the code.
- Also, the current implementation should have tried to create the required paths before their use.

Ctozzzzz:
I have the same problem. I've checked the version of visdom. It's 0.2.1, but it still couldn't work.

da-h:
PS: I just saw that conda-forge does not use the most recent version, yet.

See:
* https://anaconda.org/conda-forge/visdom
* https://github.com/conda-forge/visdom-feedstock/pull/14

da-h:
**Long-Term Solution**:
[...]

PR #839, opened 2022-03-04
Fix Relative Env Path Errors

da-h:
## Description
Just a small one, fixing #568 & thus enabling to start the visdom server locally, i.e. by using
```sh
visdom -env_path .
```

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] For JavaScript changes, I have re-generated the minified JavaScript code.

Issue #613, opened 2019-04-26
vis.svg(svg_file=xxx) not working


SsnL:
Try `vis.svg` with the this svg file:
https://www.dropbox.com/s/exev3zmdn6ckeyj/test.svg?dl=0
which looks like 
![Screenshot 2019-04-26 12 51 42](https://user-images.githubusercontent.com/5674597/56823512-0d21b380-6822-11e9-9697-c981c59d691c.png)

Observe in web UI:
![Screenshot 2019-04-26 12 52 02](https://user-images.githubusercontent.com/5674597/56823534-23c80a80-6822-11e9-961f-a941b13e6c06.png)
with a lot of errors in js console like
![Screenshot 2019-04-26 12 52 14](https://user-images.githubusercontent.com/5674597/56823544-2c204580-6822-11e9-9795-0e90b5b307f0.png)

JackUrb:
Hm I'm not quite sure why this would be broken if `matplot` is working as they rely on the same backend, but I'll see what I can find out. Thanks for flagging!

mraniket404:
Hi @maintainers 👋

I noticed this issue has been inactive for a while. 
I would like to work on this issue. Please assign it to me if it's still open.

Thanks!

rajnisht7:
Closing this issue as it doesn't exist anymore. Thank You.

PR #881, opened 2022-10-20
Fix "no such directory error" on startup

da-h:
## Description
This PR solves the "no such directory error" when starting visdom with an `-env_path` that does not exist. For new users this is even the case for just calling `visdom` without any arguments.

**Background**:
In trying to solve one issue, I have to admit, I created another one in PR #839.
Reason for this shortcoming is probably that the function `ensure_dir_exists` has previously created the directory for the parent of the given path, i.e. expecting a proper file-path, but, it has been used in two variants in the code:
1. first for layouts (in the function `load_layouts`), where it has been called on a specific `layout.json`-file
2. for arbitrary `env_paths`, where the function did only work when the path given by the user would have ended with a path-delimiter (`/`). (what i wanted to solve in #839).

**This PR solves this issue by**
1. letting the function work specifically on directories, as the name suggests. by using `os.path.abspath` the function would also normalize any trailing path-delimiter (`/`)
2. ensuring that all uses (only two in the code-base) of that function call it with a directories
3. on the way I say that the function has been called in `load_layouts` even if the path already existed, thus I (re)moved `ensure_dir_exists` there
4. i have also added the working directory to the logger-output so that users know where their data is stored

## Motivation and Context
See issues #875 and #874 for details. (This issue should solve both of these.)

## How Has This Been Tested?
The following variants are working with this PR in place:
- Startup:
   - `visdom` (creates `$HOME/.visdom` if it does not exist)
   - `visdom -env_path /tmp/doesnotexist` (creates directory)
   - `visdom -env_path /tmp/doesnotexist/` (creates directory also with a trailing `/`)
   - `visdom -env_path .` (enabled by the change of #613 )
- Layout Saving: (as this has been changed in the PR as well)
   - start visdom → select any env and any view → open "manage views" window (the folder on the right of the menu bar) → check that "fork" and "delete" work as expected

## Types of changes

## Checklist:

  Minor version increase to `0.2.3`.

# Note
The checks will probably fail, till #880 and #879 is merged.


da-h:
Rebased onto master.

PR #896, opened 2022-12-06
Cypress Tests for -use_frontend_client_polling variant of visdom

da-h:
## Description
Ensuring functionality also using the `-use_frontend_client_polling` flag.

## Motivation and Context
Regarding the recent proxy-users and an upcoming code simplification of the server socket handlers, I suggest to add some testing of the polling-feature.
I have explicitly only added the "functional tests" to be tested against the `-use_frontend_client_polling` flag, which I though should be sufficient.

**In addition, a few minor changes were necessary in the following test cases:**
- the ImagePane tests `image_callback` and `image_callback2` failed now and then for the poller-implementation, but not for the websocket implementation. This was due to two windows being loaded in different orders in both implementation. (The image-panes load slower when using long polling). This PR solves this by adapting all cypress selectors for image panes to use a "pane with an Image" instead of the "first" pane in the list of panes.
- The test for the view modal did not work due to a similar problem: the image pane loaded slower for the poller-implementation. For sake of consistency, I've added `wait(500)` commands in between the respective pane-creation commands.
- The `text_callback` test failed randomly as well. This is probably due to too a missing order-enforcement of the events when passing incoming events to the callback. (Not sure here. Is this correct?). This PR increases the delay in-between callback keystrokes to `200ms` (from previously `50ms`) to solve the issue of inconsistent tests.

## How Has This Been Tested?
Testing using this very PR-action, which should already include the test.
Also, to test for consistency, I've triggered the exact same action onto 10 branches, see for instance [here](https://github.com/da-h/visdom/pull/22)

## Types of changes

- [x] Test cases

## Checklist:

### DIFF
--- a/.github/workflows/process-changes.yml
+++ b/.github/workflows/process-changes.yml
@@ -137,7 +137,7 @@ jobs:
           path: cypress/screenshots
 
   funcitonal-test:
-    name: "Functional Test"
+    name: "Functional Test (Websocket)"
     runs-on: ubuntu-latest
     needs: install-and-build
     steps:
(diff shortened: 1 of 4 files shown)


JackUrb (approved):
Happy to have more tests here

PR #898, opened 2023-01-07
Deduplicate Socket-Handler Code

da-h:
## Description
This PR merges duplicated code in the six classes from the file `py/visdom/server/handers/socket_handlers.py`:
- `VisSocketHandler`
- `VisSocketWrapper`
- `SocketHandler`
- `SocketWrapper`
- `SocketWrap`
- `VisSocketWrap`

reducing also the total file size by about 29%.

**Notes**:
- To be sure that no functionality is missing in the refactored code, **this PR should not be merged before #896** that also tests for the polling variant of visdom.
- I probably did a lousy job at naming the classes:
  - First, I could not use the prefix `Base` as the code uses it already. Instead I used the prefix `All`.
  - Second, lacking a better unifying naming scheme for the `*Handler` and `*Wrapper` classes, I named the base class `*HandlerOrWrapper`.
  - Do you maybe have a better idea for a cleaner naming scheme?
- This also fixes a bug that probably no one has ever noticed: WebSocket clients could use the command `pop_embeddings_pane`, polling-based socket clients could not use the command. This PR fixes this bug implicitly.

## Motivation and Context
Some duplicated code in the four classes has been quite entangled. This PR aims to reuse the respective code segments to make code changes more consistent.

## How Has This Been Tested?
`-`

`-`

## Types of changes
- [x] code refactor or cleanup (changes to existing code for improved readability or performance)

## Checklist:

### DIFF
--- a/py/visdom/server/handlers/socket_handlers.py
+++ b/py/visdom/server/handlers/socket_handlers.py
@@ -14,22 +14,11 @@ the data_model itself.
 """
 
 import copy
-import getpass
-import hashlib
 import json
-import jsonpatch
 import logging
-import math
 import os
 import time
-from collections import OrderedDict
-
-try:
-    # for after python 3.8
-    from collections.abc import Mapping, Sequence
-except ImportError:
-    # for python 3.7 and below
-    from collections import Mapping, Sequence
+import types
 
 import tornado.ioloop
 import tornado.escape
@@ -52,9 +41,7 @@ from visdom.server.defaults import MAX_SOCKET_WAIT
 # basehandler
 # TODO abstract out any direct references to the app where possible from
 # all handlers. Can instead provide accessor functions on the state?
-# TODO abstract socket interaction logic such that both the regular
-# sockets and the poll-based wrappers are using as much shared code as
-# possible. Try to standardize the code between the client-server and
+# TODO Try to standardize the code between the client-server and
 # visdom-server socket edges.
 
 
@@ -74,7 +61,11 @@ from visdom.server.defaults import MAX_SOCKET_WAIT
 #   Write access is limited to data and view organization (i.e. layout settings, env removal and env saving)
 
 
-class VisSocketHandler(BaseWebSocketHandler):
+class AnySocketHandlerOrWrapper(BaseWebSocketHandler):
+    def __init__(self, *args, **kwargs):
+        self.polling = False
+        super().__init__(*args, **kwargs)
+
     def initialize(self, app):
         self.state = app.state
         self.subs = app.subs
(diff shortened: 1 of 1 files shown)


da-h:
**Suggestions**
1. I propose a further unification by allowing any client to be both `Socket`s (i.e., reading data access) **and** `VisSockets` (i.e., writing data access).
   As both classes are function-wise completely orthogonal, the feature should be easily implementable in a backward-compatible way, and the change would simplify the code quite a bit.
2. Currently, the REST-API and the socket-API provide mostly orthogonal functions/commands to any client. For instance, the REST-API provides `win_exists`, `fork_env`, etc., while the socket-API provides `save_layouts`, `delete_env`, etc.
   I suggest merging the REST-API endpoints with the socket endpoints; both could offer the same commands to any client capable of handling these. Also, an easier-to-use endpoint structure could pave the way for further client implementations.


What do you think?

JackUrb:
> Currently, the REST-API and the socket-API provide mostly orthogonal functions/commands to any client. 

This is an astute observation, and I agree that further unification would be possible. Perhaps this could be improved by a `VisAPI` base class, which could then be extended for versions using REST vs socket-based communication in either direction, and then people could initialize `Visdom` with configuration options that assist setup. 

```python
class VisAPI(...):
  ...
  
  @staticmethod
  def from_config(...) -> "VisAPI":
    ...
```

da-h:
>This is an astute observation, and I agree that further unification would be possible. Perhaps this could be improved by a VisAPI base class, which could then be extended for versions using REST vs socket-based communication in either direction

I completely agree here!

> and then people could initialize Visdom with configuration options that assist setup.

I am not sure if I understand it correctly. Do you mean: there could be a handshake between server and client that contains what commands they both agree with? So to speak a *rolling* API-version?

JackUrb:
> there could be a handshake between server and client that contains what commands they both agree with? So to speak a rolling API-version?

Ah not quite, that's more complicated than what I imagine. Just that the `Visdom` object can have different instances of the API abstraction to allow for all the weird network settings that exist out there, and so long as they adhere to the API (and either REST or sockets) then the server can register them

da-h:
> Then connecting clients to the server could be either, so long as they're using the agreed VisAPI.

One further question here: What would be the benefit of such an predefined agreement instead of just allowing any client to open any socket-type at any time?

JackUrb:
> > Then connecting clients to the server could be either, so long as they're using the agreed VisAPI.
> 
> One further question here: What would be the benefit of such an predefined agreement instead of just allowing any client to open any socket-type at any time? (The latter seems easier to implement & to manage, however, I am also open to implement the former if it is better.)

It's mostly just about establishing some common API layer, then we don't really need to care which one someone is opening.

da-h:
Got it. Thanks for clarifying. :)

da-h:
(Rebased onto master).

JackUrb (approved):
This at least cuts down on some of the deduplication here, though I'd say my original namings for these classes weren't great to begin with. At least changes now won't be missed between the two, though if things get complex enough refactoring the API and making it connection-method-agnostic by default will probably be the way to go. No need to do that anytime soon though!

PR #900, opened 2023-01-08
[Big change] Migrate main.js to Functional React

da-h:
## Description
This PR follows up on #856 and migrates the remaining component (`App`) to functional React. As a side effect, this PR also fixes all the linting errors related to JavaScript files.

## Motivation and Context
The goal is again to modernize the code base to enable better long-term maintainability.

For me, this one has been the most challenging PR so far and it is my third attempt at solving the task.
This is probably due to
1. **my personal requirement of trying to keep the PR as reviewable as possible**.
   To accomplish this I have tried to keep all code blocks and functions in the same respective position. Unfortunately, the changes required for the migration are still numerous.
2. **some very specific timing-requirements of redraws for the application to work properly**. 
   In the previous code, there have been some comments noticing "non-conventional" React style and I think that these and some other design decisions did not fit well with  "hook based react/functional react" anymore.
   Solving these timing / redraw issues would have required a complete restructuring of many parts of the component, which would have been practically a complete rewrite of the code. This would have clashed with the former requirement.

Thus, I went with a `1:1` translation that is a bit easier to review (the first commit). The PR can be compared with the previous code in a side-by-side manner.
The drawback is that the many tricks from the old code (i.e. the "non-conventional" React parts) needed to be retained/reimplemented for the new style. (See the small commits after the huge first one).

**Future follow-up PR ideas**:
After this PR, tidying up the code base further should be possible in smaller PRs again.
For instance, I'd suggest to
- extract the API-related functions to an encapsulated API-component
- merge the variables `envID` and `envIDs` as they share the same function (one for compare views, the other for single view mode)
- Simplifying the pane-store structure.
  Especially variables such as `consistent_pane_copy` duplicates code and makes it hard to understand some functions. A comment in the code suggests to use redux. Is this still the "way to go" for applications such as visdom?

**A final note on the two additional test changes**:
- The new implementation fixes a bug related to window resize resetting (triggered by double-clicking the lower-right pane corner).
  As the cypress test did ignore that bug explicitly, this PR adapts the test to the now correct expected values.
- During work on this PR I found another inconsistency in the cypress tests, probably due to the new implementation.
  In more detail, closing an environment requires waiting for an animation to end (and an `x` to appear). With the new implementation, the animation is probably triggered slightly delayed, thus I've added a `100ms` delay in the respective test.

## How Has This Been Tested?
As I've said, this is my third attempt to finish this PR while making all tests succeed.
The new implementation may create new bugs that have not been present before, however, with the testing in place I am confident that almost all features work just as before the PR.

Just to be sure, I've also 10-fold triggered the PR against all tests without any errors on my fork.

`-`

## Types of changes

## Checklist:
    (Maybe not yet? I am confident in the changes I've made, still I'd test the new version for at least a week or two before releasing it to the public. ;) )

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -393,6 +393,9 @@ function App() {
     }
   };
 
+  // we need to update the socket-callback so that we have an up-to date state
+  if (_socket.current) _socket.current.onmessage = _handleMessage;
+
   // close server connection
   const disconnect = () => {
     _socket.current.close();
--- a/js/main.js
+++ b/js/main.js
@@ -744,18 +744,18 @@ function App() {
     });
   };
 
-  const updateToLayout = (layoutID) => {
+  const updateToLayout = (newLayoutID) => {
     setSelection((prev) => ({
       ...prev,
-      layoutID: layoutID,
+      layoutID: newLayoutID,
     }));
     // TODO this is very non-conventional react, someday it shall be fixed but
     // for now it's important to fix relayout grossness
-    selection.layoutID = layoutID;
+    selection.layoutID = newLayoutID;
     if (selection.layoutID !== DEFAULT_LAYOUT) {
-      relayout();
-      relayout();
-      relayout();
+      callbacks.current.push('relayout');
+      callbacks.current.push('relayout');
+      callbacks.current.push('relayout');
     }
   };
 
@@ -1076,17 +1076,11 @@ function App() {
       filter={filterString}
       onFilterChange={(ev) => {
         setFilterString(ev.target.value);
-        // TODO remove this once relayout is moved to a post-state
-        // update kind of thing
-        relayout();
-        relayout();
+        callbacks.current.push('relayout');
       }}
       onFilterClear={() => {
         setFilterString('');
-        // TODO remove this once relayout is moved to a post-state
-        // update kind of thing
-        relayout();
-        relayout();
+        callbacks.current.push('relayout');
       }}
     />
   );
--- a/js/main.js
+++ b/js/main.js
@@ -696,9 +696,6 @@ function App() {
       ...prev,
       panes: newPanes,
     }));
-    // TODO this is very non-conventional react, someday it shall be fixed but
-    // for now it's important to fix relayout grossness
-    storeData.panes = newPanes;
     updateLayout(newLayout);
   };
 
--- a/js/main.js
+++ b/js/main.js
@@ -813,12 +813,13 @@ function App() {
    *
    * @param data Data to be sent to backend.
    */
-  const sendPaneMessage = (data) => {
-    if (focusedPaneID === null || sessionInfo.readonly) {
+  const sendPaneMessage = (data, target = null) => {
+    if (!target) target = focusedPaneID;
+    if (target === null || sessionInfo.readonly) {
       return;
     }
     let finalData = {
-      target: focusedPaneID,
+      target: target,
       eid: selection.envID,
     };
     $.extend(finalData, data);
--- a/js/main.js
+++ b/js/main.js
@@ -254,11 +254,29 @@ function App() {
     if (_socket.current) {
       return;
     }
+
+    const _onConnect = () => setConnected(true);
+    const _onDisconnect = () => {
+      // check if is mounted. error can appear on unmounted component
+      if (mounted.current) {
+        callbacks.current.push(() => {
+          _socket.current = null;
+        });
+        setConnected(false);
+      }
+    };
+
     // eslint-disable-next-line no-undef
     if (USE_POLLING) {
-      _socket.current = new Poller(this);
+      _socket.current = new Poller(
+        correctPathname,
+        _handleMessage,
+        _onConnect,
+        _onDisconnect
+      );
       return;
     }
+
     var url = window.location;
     var ws_protocol = null;
     if (url.protocol == 'https:') {
@@ -271,21 +289,8 @@ function App() {
     );
 
     socket.onmessage = _handleMessage;
-
-    socket.onopen = () => {
-      setConnected(true);
-    };
-
-    socket.onerror = socket.onclose = () => {
-      // check if is mounted. error can appear on unmounted component
-      if (mounted.current) {
-        setConnected(false);
-        callbacks.current.push(() => {
-          socket.current = null;
-        });
-      }
-    };
-
+    socket.onopen = _onConnect;
+    socket.onerror = socket.onclose = _onDisconnect;
     _socket.current = socket;
   };
 
--- a/js/main.js
+++ b/js/main.js
@@ -14,7 +14,7 @@
 import 'fetch';
 import 'rc-tree-select/assets/index.css';
 
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import ReactDOM from 'react-dom';
 import ReactResizeDetector from 'react-resize-detector';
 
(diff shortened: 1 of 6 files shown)


js/main.js
```diff
@@ -61,78 +61,80 @@ if (ACTIVE_ENV !== '') {
...
+
+  // internal variables
+  const mounted = useRef(false);
+  const [connected, setConnected] = useState(false);
+  const [sessionInfo, setSessionInfo] = useState({ id: null, readonly: false });
+  const [resizeClickHappened, setResizeClickHappened] = useState(false);
+  const windowSize = useRef({
+    // width: 1280,
+    // cols: 100,
+  });
```
JackUrb: It's a bit odd to initialize this ref with empty values, as it's unclear where these will eventually come from.

js/main.js
```diff
@@ -61,78 +61,80 @@ if (ACTIVE_ENV !== '') {
...
+
+  // user-changeable
+  const [showEnvModal, setShowEnvModal] = useState(false);
+  const [showViewModal, setShowViewModal] = useState(false);
+  const [focusedPaneID, setFocusedPaneID] = useState(null);
+  const [selection, setSelection] = useState({
     envID: use_env,
     envIDs: use_envs,
     layoutID: DEFAULT_LAYOUT,
-    // Bad form... make a copy of the global var we generated in python.
```
JackUrb, line 77: I still think this isn't great form, but alas. Not to resolve here, but maintain the note

js/main.js
```diff
@@ -154,61 +157,78 @@ class App extends React.Component {
...
   };
 
-  processBatchedPanes = () => {
-    let newPanes = Object.assign({}, this.state.panes);
-    let newLayout = this.state.layout.slice();
+  // run processing on queue
+  const processBatchedPanes = () => {
+    // wait until app is mounted
+    if (!mounted.current) {
+      // console.log('not processing');
```
JackUrb: can drop log

js/main.js
```diff
@@ -154,61 +157,78 @@ class App extends React.Component {
...
 
-    this._pendingPanes.forEach((pane) => {
-      this.processPane(pane, newPanes, newLayout);
+    _pendingPanes.current.forEach((pane) => {
+      processPane(pane, newPanes, newLayout);
     });
 
-    this._pendingPanes = [];
-    this._timeoutID = null;
+    _pendingPanes.current = [];
```
JackUrb: In theory this code still has a race condition that may drop pane updates. Not sure if `React` introduces any kind of async that could be unstable. We should do
```
pendingPanes = _pendingPanes.current
_pendingPanes.current = []
penddingPanes.forEach(...
```
_just in-case_

js/main.js
```diff
@@ -154,61 +157,78 @@ class App extends React.Component {
...
+    }));
   };
 
-  processPane = (newPane, newPanes, newLayout) => {
+  // process single pane
+  const processPane = (newPane, newPanes, newLayout) => {
     let exists = newPane.id in newPanes;
     newPanes[newPane.id] = newPane;
 
+    consistent_pane_copy.current[newPane.id] = deepcopy(newPane);
```
JackUrb, line 202: does it change behavior that this deepcopy is made before modifications happening below in the `else`?

JackUrb (commented):
Wow @da-h, you've outdone yourself. This is a *huge* change and I'm sure one to be appreciated by anyone who has wanted to make changes to Visdom but been intimidated by the oudated absolute mess it used to be.

> A comment in the code suggests to use redux. Is this still the "way to go" for applications such as visdom?

It's an approach, but prop drilling and state management is something that really differs from frontend developer to frontend developer. Not using one tends to be more explicit but verbose.

> `consistent_pane_copy `

I'll be honest, I've forgotten what the purpose of this initially was

JackUrb:
>  trying to keep the PR as reviewable as possible

Thanks for taking the extra effort for this - it definitely made it easier for me to check over what was going on. All-in-all your "Future follow-up PR ideas" are exactly what I'd imagine the remaining vision is to shore up the frontend code into a truly clean state. 

> Maybe not yet? I am confident in the changes I've made, still I'd test the new version for at least a week or two before releasing it to the public. ;) 

Honestly with all your DevX, stability, and clarity improvements I think that v1.0.0 could be on the horizon :)

js/main.js
da-h: Very good point. I am pretty sure I came across this bug for an unpublished speed test I once wrote.
I will definitively include this.

js/main.js
da-h: I moved that line up here, because there has been the a duplicate operation at the start of the condition (right after)
```
if (!exists)
```
the other is in the else branch, however at the end there.

Functionality-wise, there should be no difference in the order, as `newPane` is not changed in this function (`processPane`) at all. Just the layout (`newLayout`) gets an update, but is not affected by `consistent_pane_copy`.

js/main.js
da-h: Adapted & amended first commit. (I found it cleaner that way)

js/main.js
da-h: Adapted & amended first commit. (I found it cleaner that way)

js/main.js
da-h: Adapted & amended first commit. (I found it cleaner that way)

js/main.js
da-h: added a new commit for this change

js/main.js
```diff
@@ -355,94 +372,99 @@ class App extends React.Component {
...
         break;
       default:
         console.error('unrecognized command', cmd);
     }
   };
 
-  disconnect = () => {
-    this._socket.close();
+  // we need to update the socket-callback so that we have an up-to date state
+  if (_socket.current) _socket.current.onmessage = _handleMessage;
```
da-h, line 395: > > `consistent_pane_copy `
> 
> I'll be honest, I've forgotten what the purpose of this initially was

If I understood correctly, the variable exists, because all functions called by `socket.onhandler` or below it in the calling hierarchy use the state of the initialization of the socket itself.
On the other hand, all other functions exist in subsequent calls of the outer react function and thus have updated states.
To tackle this,
- the variable `consistent_pane_copy` is the one variable that ensures to always contain the correct state of the server, implicitly telling the newer render-functions-instances about all changes
- `_checkWindow` for instance applies changes always to both, the current state pane store and the `consistent_pane_copy` to ensure both are synced in the "long term"
- the old code had sometimes overwritten state variables using operations such as `state.panes = newPanes` to update variables in the short term (i.e. in case of race conditions)

Migrating that part in a `1:1` way was the hardest part and I could be completely wrong about what the variable actually does.
However, I've added here an update of the socket to always use the `_handleMessage` from the current render-call. This one line has fixed several problems I had during this PR.

da-h: Interestingly, in the new implementation I could remove the variable completely without any failing tests.
(See #901).

da-h:
Rebased onto master (including the polling-tests now).

This did not work properly, so I also adapted the polling class to match the new code in a new commit.

da-h:
Polling-based initial connection is a bit delayed, thus the "manual server reconnect" cypress test had to be adapted to wait a bit for the initial "online" status to appear. I used the opportunity to also add multiple fast user-reconnects to the test.

da-h:
> Honestly with all your DevX, stability, and clarity improvements I think that v1.0.0 could be on the horizon :)

:100: x :heart:
Sure! I'd suggest to publish the major release after the proposed changes from here and #898 are done & in-place.
This should provide a clean code base for anyone to start new improvements on.

JackUrb (approved):
Alright sorry for the delay here, I think we should pull this in!

da-h:
No stress, good things take time. :)

PR #901, opened 2023-01-14
Replaced Window Update Check by 'versioning' (previously used 'hashing')

da-h:
## Description
This PR 
- removes the variable `consistent_pane_copy` from `js/main.js`
- **api-change**: implements version-numbers for window content, s.t. a client knows if it is up-to-date or needs to re-request the window content
- **api-change**: removes the api endpoint `/win_hash` (as it is not used anymore in the project)
- additionally, moves the pane-update to the `processPane` method. this is required to save some requests where the whole window content would have be resent (see notes below).

## Motivation and Context
(See discussion below).

## How Has This Been Tested?
(Just the cypress tests).

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,6 @@ vis._send({'data': [trace], 'layout': layout, 'win': 'mywin'})
 - [`vis.delete_env`](#visdelete_env) : delete an environment by env_id
 - [`vis.win_exists`](#viswin_exists) : check if a window already exists by id
 - [`vis.get_env_list`](#visget_env_list) : get a list of all of the environments on your server
-- [`vis.win_hash`](#viswin_hash): get md5 hash of window's contents
 - [`vis.get_window_data`](#visget_window_data): get current data for a window
 - [`vis.check_connection`](#vischeck_connection): check if the server is connected
 - [`vis.replay_log`](#visreplay_log): replay the actions from the provided log file
@@ -753,13 +752,6 @@ Optional arguments:
 #### vis.get_env_list
 
 This function returns a list of all of the environments on the server at the time of calling. It takes no arguments.
-
-#### vis.win_hash
-
-This function returns md5 hash of the contents of a window `win` if it exists on the server. Returns None otherwise.
-
-Optional arguments:
-- `env` : Environment to search for the window in. Default is `None`.
 
 
 #### vis.get_window_data
--- a/js/main.js
+++ b/js/main.js
@@ -101,7 +101,6 @@ function App() {
   );
 
   // non-triggering state variables
-  const consistent_pane_copy = useRef({});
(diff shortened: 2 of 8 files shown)


da-h:
That's interesting: the variable is required for react 17. I've tested it on react 16 and 18 successfully.
As 18 gives some very good changes, like automatic batched processing, I'll suggest to look into this once/if visdom runs under react 18.

da-h:
With #908 in-place, this PR succeeds all tests now & is thus now ready for review.

JackUrb (commented):
Now that I read this through though, I remember what the `consistent_pane_copy` was for. It served the following purpose: When a window **patch** comes in, apply the patch to the consistent copy and see if the hash is the same. If it's not, then it's possible that a patch was missed (perhaps offline at the time) and as such the computed window might be wrong. If it is, then apply the patch to the live copy to retain any local changes (zoom, legend options, etc). 

This, in retrospect, is a _really inefficient_ way to resolve versioning. Instead, I suggest we update the server and here to just have an incrementing version number attached to the patches (instead of `finalHash`), and if the version you get in a patch isn't the immediate next version, request the full pane (`postForEnv`). This prevents the need to be hashing on both sides.

da-h:
I see!!
I would have not thought about this. Thanks for clarifying. :+1: 

I'll update this the next days to your proposed solution.

da-h:
I've just implemented the window-message versioning, to replace `consistent_pane_copy`.
With the background-information of how this mechanism worked, I now also understood the reason for the `setTimeout` inside of the `_checkWindow` (now called `updateWindow`) function:
- in case a pane-update is found for a pane that is not yet processed, we wait for some time until the update can be applied
- then, if the time passes and no such pane is found, we (re-)request the pane using `postForEnv`
- this, however can lead to wrong orders of window updates: imagine there are two updates waiting for a pane to be saved in `storeData` using such timers, say update1 and update2 (that builds upon the changes in update1). In case the pane is saved, just when update1 is re-queued using a new timer and update2 checks for the existence of the pane, the patch should lead to a wrong state, or possibly even to an error.

Using the idea of specifying the `version` of each pane state, we can circumvent the timers:
- we move the actual update to the `processPane`-function and remember the last version that we know is waiting to be processed
- if a new update comes in, we first check if the pane exists in `storeData` and matches our expectation of the previous version. Or, If it does not exist or the version does not match, we check also if there is a update waiting to be processed, with our expected version of the pane state.
- In total, I expect this implementation to be a bit snappier.

da-h:
**Just to be sure**:

JackUrb (approved):
This fix is awesome - I think it's going to have a more significant performance win associated with it than you expect for high-data or long-running applications.

PR #902, opened 2023-01-26
Upgrade from React 16 to React 17 & npm update --save

da-h:
## Description
Upgrading from react 16 to react 17 including all dependencies that did not work anymore with react 17.
No code changes required.

Note: removing also the package `react-select` as it is not used in the project anymore.

## Motivation and Context
[React 17 can be applied safely & directly without any further changes.](https://reactjs.org/blog/2020/10/20/react-v17.html)
React 18 on the other hand requires more changes to be made:
For this project, this is due to some changes the react-team made to the order of events and renders to be executed internally.
This fixes some minor bugs in visdom, that have been expected in some test cases. (The tests did mention the bugs in the code as a comment). With react 18 these bugs are easily fixable. There will be another PR for the React 18 update.

## How Has This Been Tested?
`-`

Note, the javascript-linter is expected to fail here until #900 is merged.

## Types of changes
- [x] Package updates
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactor or cleanup (changes to existing code for improved readability or performance)

## Checklist:
- [ ] I adapted the version number under `py/visdom/VERSION` according to [Semantic Versioning](https://semver.org/)
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #908, opened 2023-02-27
Fix ImagePane tests

da-h:
## Description
This PR fixes the failing `ImagePane` cypress tests.

## Motivation and Context
Tests started to fail after #900 & #902 were merged, although both were succeeding when tested separately.
- One problem was that in the previous version, the click-positions were always off by one pixel. The new version has fixed this. I went the extra few steps to also check for correct pixel positions in the test, which has not been done before.
- Second, the image dragging cypress tests did not work as expected. After some tinkering, I best fixed the test using the `@4tw/cypress-drag-drop` cypress plugin. I also added to test the combination of zooming + panning.

## How Has This Been Tested?
- Locally with cypress against the current master + this PR
- on my fork on top of #898 
- on my fork on top of #901 

## Types of changes

- [x] Test changes (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/cypress/integration/image.js
+++ b/cypress/integration/image.js
@@ -165,19 +165,21 @@ describe('Image Pane', () => {
       .should('have.attr', 'ry', 30);
   });
 
+  let click1 = [12, 34];
+  let click2 = [45, 67];
   it('image_callback', () => {
     cy.run('image_callback', { asyncrun: true });
     cy.get(img_selector)
       .parents(win_selector)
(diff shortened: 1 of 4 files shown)


da-h:
LGTM now.
It seems it has been a timing issue with the polling variant.
I've basically added just a short delay in the test before the second click in the callback-test.

JackUrb (approved):
Thanks for the fix here @da-h

PR #914, opened 2023-03-23
Merge variables envID and envIDs

da-h:
## Description
Just a small proposal for simplification.
In the react component (`js/main.js`), there is a variable called `envID` and another one called `envIDs`.
As `envID == envIDs[0]` if any env is selected, this PR suggests to remove the duplicate variable `envID`.

## Motivation and Context
The javascript code distinguishes the case of a single selected environment and multiple ones.
However, if multiple ones are selected, some functions effectively use just the first environment of the list, instead of merging the environments somehow. For instance, forking environments or views is also possible for multiple envs selected, in both cases it uses always the name of the first environment. (I guess this is minor a bug).

I know that the proposed change does not solve this at all, still I suggest to simplify the code at this point.
If you have an idea how to improve this PR, please let me know. :)
For instance, I think it would be reasonable to just disable the Fork- and View-Buttons when multiple envs are selected for now as their behavior is currently not well-defined.

## How Has This Been Tested?
Just a short local test.

## Types of changes

## Checklist:

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -42,20 +42,16 @@ const GridLayout = WidthProvider(ReactGridLayout);
 const sortLayout = ReactGridLayout.utils.sortLayoutItemsByRowCol;
 const getLayoutItem = ReactGridLayout.utils.getLayoutItem;
 
-var use_env = null;
 var use_envs = null;
 if (ACTIVE_ENV !== '') {
   if (ACTIVE_ENV.indexOf('+') > -1) {
     // Compare case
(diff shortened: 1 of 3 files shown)


JackUrb (approved):
In retrospect it was a little silly we didn't have this setup initially. Thanks for the fix!

da-h:
Refactoring something like this is easier than writing it new from scratch, when the function set of the project was growing, and it was unclear where the project was heading, I guess.

PR #918, opened 2023-04-24
Extract js-api methods into seperate file

da-h:
## Description
This PR refactors server-related functions in the client code by moving them into the `js/api/ApiProvider.js` component.

Also:
- Enhanced `sendPaneMessage` and `sendEmbeddingPop` by adding `targetId` and `envID` as arguments, making the code more versatile for pane-functions that trigger server messages when not selected. However, no changes in user functionality have been made.
- Replaced all "raw" `sendSocketMessage` calls with specialized versions for improved code readability and maintainability.
- Renamed communication functions for better semantics and consistency:
    - `sendX`: Functions send data or queries *to the server*. Normalized naming to `sendNounVerb` for an improved autocompletion experience.
    - `onX`: Functions that are called when the server sends something *to the client*, aligning with event naming conventions.

## Motivation and Context
The aim was to enhance code readability and reduce the lines of code in `main.js`.
Using `useContext` has simplified the use of API functions in child components. Previously, these functions needed to be forwarded in the JSX calls, but now they can be accessed globally.

## How Has This Been Tested?
`-`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Code refactor or cleanup (changes to existing code for improved readability or performance)

## Checklist:

- [ ] I adapted the version number under `py/visdom/VERSION` according to [Semantic Versioning](https://semver.org/)
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #920, opened 2023-04-24
More consistent env-select test with polling enabled

da-h:
## Description
This PR fixes the failing test "*env selection works*" (in `./cypress/integration/basic.js`) for the polling-option of visdom (i.e. using the argument `-use_frontend_client_polling`).

Reason for the error was simply that upon initialization of the client (i.e. when opening `http://localhost:8097`), two calls to query the initial env `main` are implicitly sent.
The failed test then proceeded to check if un-selecting and re-selecting the env works correctly.
The error appeared because the polling variant which is slower than the websocket variant, has responded with the second of the two initial queries when the test already cleared the selection box and thus expected the selection to be empty.

Note that the error appears only if the polling is too slow in its initial setup, thus, the error had appeared only somewhat randomly.

This can be easily solved, by letting the test wait until the page loaded correctly (~1sec).

## Motivation and Context
See #917 and #918, where the error appeared.

## How Has This Been Tested?

10-fold consistency using github actions.

## Types of changes

## Checklist:

### DIFF
--- a/cypress/integration/basic.js
+++ b/cypress/integration/basic.js
@@ -27,7 +27,7 @@ describe('Test Setup', () => {
   })
 
   it('env selection works', () => {
-    cy.visit('/')
+    cy.visit('/').wait(1000)
     cy.get('.rc-tree-select [title="main"]').should('exist')
     cy.get('.rc-tree-select').contains('main').trigger('mouseover').wait(100);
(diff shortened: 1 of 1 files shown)


JackUrb (approved):

JackUrb:
These bugs can be tricky to find, thanks for the discovery and solve!

PR #936, opened 2024-05-30
Fix disabled text and number property inputs by removing keypress preventDefault() calls interfering from unrelated panes.

### LINKED ISSUE #935: Text and number input properties not working
**Bug Description**
When using a text or number property, upon trying to change the initial value, the event that's fired only contains the old value.

**Reproduction Steps**
1. Create a properties window in any environment with a text input i.e. 
    [
        {'type': 'text', 'name': 'Test', 'value': 'Testing...'},
    ]

2. Register a callback on that window to print every event. 

3. Try to type another letter at the end of the text box that says 'Testing...' 

4. Observe the 'PropertyUpdate' event that is printed. event['value'] is still 'Testing...'. 

**Expected behavior**
The event['value'] should be 'Testing...<typed letter>'

tripp528:
## Description

Commented out some preventDefault calls on keypress events that were being called from panes that weren't the target for the event.

## Motivation and Context

Text and number property inputs were not functioning when a separate embeddings, text, or image pane was present and had a keypress callback handler registered to it. 

Fixes #935

### DIFF
--- a/js/panes/EmbeddingsPane.js
+++ b/js/panes/EmbeddingsPane.js
@@ -30,7 +30,7 @@ class EmbeddingsPane extends React.Component {
     switch (e.type) {
       case 'keydown':
       case 'keypress':
-        e.preventDefault();
+        // e.preventDefault();
         break;
       case 'keyup':
         if (this.props.isFocused)
--- a/js/panes/ImagePane.js
+++ b/js/panes/ImagePane.js
@@ -159,7 +159,7 @@ function ImagePane(props) {
       switch (event.type) {
         case 'keydown':
         case 'keypress':
-          event.preventDefault();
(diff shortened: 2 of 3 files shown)


tripp528:
I'm not sure what those preventDefault() calls were meant for exactly, but this doesn't seem to have introduced any issues thus far. A cleaner fix might be to check if e.g. the ImagePane was the target before calling preventDefault.

tonypzy:
Since the previous PR has been pending for a long time, I created a new PR (https://github.com/fossasia/visdom/pull/1265) to fix the issue with disabled text and number property inputs.

vedansh-5:
In light of PR #1265 closing this PR. 
Thanks @tripp528 for your contribution.

PR #958, opened 2026-02-10
Enable Custom x/y axis values for heatmap

### LINKED ISSUE #937: Include axes lists in heatmap
Nearly 100% of the time when I generate heatmaps, the x and y axes do not have integer values. Instead, they are equispaced lists defined by the scenario, and it is annoying to infer the list value from the figure pixel index when the figure doesn't include appropriately scaled axes.

I would like to see two optional arguments to visdom's heatmap(), specifically, I'm hoping the user would be provide optional inputs like
```python
'x':horizontal_axis_vector 
'y':vertical_axis_vector 
```

So that an example use would be 
```python
import visdom
import numpy as np
viz = Visdom()
horizontal_axis_vector= np.arange(-5,6)[np.newaxis]
vertical_axis_vector = (np.arange(0,10)*.2)[np.newaxis]
h = np.square(vertical_axis_vector.T @ horizontal_axis_vector)
viz.heatmap(h, opts={'x':horizontal_axis_vector, 'y':vertical_axis_vector})  
```

The input-output behavior would be similar to matplotlib's imshow() or matlab's imagesc()

kirandharmar867-maker:
Summary:
This PR adds support for explicit x-and y-axis values in visdom.heatmap(), allowing users to provide
properly Scaled axes instead of replying on inferred pixel indices.

Changes:
*Added optional opts["x"] for explicit x-axis values
*Added optional opts["y"] for explicit y-axis values
*validated axis length against the input heatmap shape
*preserved backward compatibility with existing columnnames / rownames

Motivation:
Currently, visdom.heatmap() always infers axis values from pixel indices, which makes it difficult to interpret heatmaps generated from non-integer or non-uniform Grids.

This change enables users to pass meaningful axis values directly, aligning visdom behavior with common
plotting tools like plotly and matplotlib.

Related issue:
Closes #937 

Checklist:
*Change is backward compatible
*No API breakage
*Minimal and scoped to heatmap()
*CI passes input Shapes
*Validates input Shapes

## Summary by Sourcery

Allow visdom heatmap plots to accept explicit x and y axis values while preserving existing behavior.

New Features:
- Add support for explicit x-axis values via opts['x'] in visdom.heatmap.
- Add support for explicit y-axis values via opts['y'] in visdom.heatmap.

Enhancements:
- Validate custom x and y axis lengths against the input heatmap shape to prevent mismatches.

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -1831,6 +1831,8 @@ class Visdom(object):
         - `opts.xmax`    : clip maximum value (`number`; default = `X:max()`)
         - `opts.columnnames`: `list` containing x-axis labels
         - `opts.rownames`: `list` containing y-axis labels
+        - `opts.x`: list containing x-axis labels
+        - `opts.y`: list containing y-axis labels
         - `opts.nancolor`: if not None, color for plotting nan
                            (`string`; default = `None`)
         """
@@ -1866,6 +1868,12 @@ class Visdom(object):
         _title2str(opts)
         _assert_opts(opts)
 
+        if "x" in opts and isinstance(opts["x"], np.ndarray):
+            opts["x"] = opts["x"].tolist()
+    
+        if "y" in opts and isinstance(opts["y"], np.ndarray):
+            opts["y"] = opts["y"].tolist()
+
         if opts.get("columnnames") is not None:
             assert (
                 len(opts["columnnames"]) == X.shape[1]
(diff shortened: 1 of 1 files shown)


kirandharmar867-maker:
> A few issues to address before this can be merged:
> 
> 1. **numpy serialisation** — `opts["x"]` and `opts["y"]` are likely numpy arrays. Passing them directly into the data dict will cause JSON serialisation errors. Convert them with `.tolist()`, following the same pattern already used for `z` (`X.tolist()`).
> 2. **Fallback with `None`** — `opts.get("x", opts.get("columnnames"))` silently falls back to `columnnames` even if the caller explicitly passed `opts["x"] = None`. Use an explicit `if "x" in opts:` check instead to respect intentional `None` values.
> 3. **Grammar in assertion message** — "number of row in X" should read "number of rows in X", consistent with the style of the existing `rownames` assertion above.
> 4. **Docstring** — The method docstring (around line 1827–1836) lists `opts.columnnames` and `opts.rownames` but does not mention the newly supported `opts.x` and `opts.y`. Please add entries for both.

Hii @norbusan
Thanks for the detailed review! I’ve addressed the points you mentioned: converted `opts["x"]` and `opts["y"]` numpy arrays using `.tolist()`, fixed the fallback logic to respect explicit `None`, corrected the grammar in the assertion message, and updated the docstring to include `opts.x` and `opts.y`.  

I also learned a lot while working on this fix and really enjoyed the process. Please let me know if there’s anything else I should adjust.
Please let me know if there’s anything else I should adjust.
Thank You.

PR #960, opened 2026-02-19
feat: add --download_only flag and visdom-download CLI for offline setups

### LINKED ISSUE #942: quickly and manually download all the required packages by the following method
In my settings, my linux server is offline and my local machine is online. so, i download packages by local machine and copy to linux.

Users can quickly and manually download all the required packages by the following method:
1. Download Visdom and comment out main() in run_server.py as follows:
```
def download_scripts_and_run():
    download_scripts()
    # main()
```
2. Install Visdom in an editable way, pip install -e visdom.tar.gz
3. Open the corresponding visdom code in VSCode and manually execute Python run_Server. py

By performing the above operations, it is possible to achieve full access to JavaScript, CSS, fonts, etc. without the need for manual intervention

pankajbaid567:
## Description

Adds a dedicated mechanism to download all required static assets (JS, CSS, fonts) **without** starting the Tornado server. Users on air-gapped/offline machines currently have to manually hack `run_server.py` (comment out `main()`) just to download dependencies on an online machine — this change makes that a first-class feature.

**Three ways to use:**
- `visdom --download_only` — new flag on the existing CLI
- `visdom-download` — new dedicated console-script entry point
- `python -m visdom.server --download_only` — module invocation

**Files changed:**
- `py/visdom/server/run_server.py` — added `--download_only` arg to the parser and a new `download_only()` function
- `setup.py` — added `visdom-download` console-script entry point
- `py/visdom/server/__main__.py` — routes `--download_only` to `download_only()` when invoked as a module

## Motivation and Context

Fixes #942. Users with offline Linux servers need to download Visdom's JS/CSS/font dependencies on an online machine and then copy them over. The current workaround requires manually editing `run_server.py` to comment out `main()`. This change provides a clean, supported way to do that.

## How Has This Been Tested?

- All three modified files verified with `ast.parse` — no syntax errors.
- Manual verification: confirmed `--download_only` flag triggers `download_scripts()` and exits without starting the Tornado server (no "It's Alive!" message).
- Confirmed that the normal `visdom` command still works as before (backward compatible).

## Types of changes

## Checklist:

## Summary by Sourcery

Introduce a dedicated offline asset download mode for Visdom without starting the server.

New Features:
- Add a --download_only flag to the Visdom server CLI to download static assets and exit without running the server.
- Expose a new visdom-download console script entry point for downloading Visdom client assets in isolation.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -62,7 +62,7 @@ runs:
 
       # load artifacts from previous runs
       - name: "Load built js-files"
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v7
         with:
           name: pr-build
           path: ./py/visdom/static/js/
--- a/.github/workflows/process-changes.yml
+++ b/.github/workflows/process-changes.yml
@@ -62,7 +62,7 @@ jobs:
         if: steps.checkout.outputs.jsfileschanged > 0
 
       - name: "Save built js-files"
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: pr-build
           if-no-files-found: error
@@ -104,7 +104,7 @@ jobs:
           wait-on: 'http://localhost:8098'
           spec: cypress/integration/*.init.js
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: cypress-init-screenshots
           path: cypress/screenshots_init
@@ -119,7 +119,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: ./.github/actions/prepare
 
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v7
         with:
           name: cypress-init-screenshots
           path: cypress/screenshots_init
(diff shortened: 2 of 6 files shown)


norbusan:
A few things worth addressing:

1. **Double download risk**: `download_scripts_and_run()` (run_server.py:234) already calls `download_scripts()` before `main()`. If `main()` also calls `download_scripts()` when it detects `--download_only`, assets get downloaded twice when using the `visdom` console script with that flag. Consider having `download_scripts_and_run()` drop its own `download_scripts()` call and let `main()` be the single place that decides whether and when to download, or restructure so the download happens in exactly one place regardless of entry point.

2. **`download_only()` wrapper**: This function exists only to call `main()` via a private `_download_only` kwarg. It would be cleaner for the `visdom-download` entry point to call a properly documented public function rather than tunnelling through an internal flag.

3. **Test coverage**: Is there a way to add a smoke test that verifies the flag exits cleanly without starting the server? Even a simple check that `"It's Alive!"` is absent from stdout would prevent regressions.

marcoag (changes requested):
Thanks for the PR.

A few things to address before we can merge:

1. The diff includes GitHub Actions artifact version bumps (`upload-artifact` v3→v4, `download-artifact` v3→v7) — these are already in master. Please rebase your branch to clean this up.

2. When running `visdom --download_only`, `download_scripts()` is called twice. The `visdom` entry point (`download_scripts_and_run`) always calls `download_scripts()` first, then calls `main()` — which also calls `download_scripts()` when it sees the `--download_only` flag.

mariobehling (changes requested):
Please address outstanding feedback.

vedansh-5:
Thanks @pankajbaid567 for your contribution. Closing this PR due to inactivity, this change is being followed in #1268

PR #974, opened 2026-02-23
fix(scatter): relax markercolor palette size constraint

### LINKED ISSUE #357: MarkerColor dimension restrictions are too strict
If one wanted to use the same color scheme for multiple plots, they'd be unable to do so. Instead of bounding between 1 and the number of plots, it should simply make sure that the indices given are valid given the markercolor size (ie max(marker_index) <= marker_dim)

devs6186:
## Description

`_markerColorCheck` previously required the marker-color palette to have **exactly** `L` entries (where `L = max(Y)`), making it impossible to reuse a shared colour palette across multiple scatter plots when the palette was larger than any single plot's label count.

This PR:
- Relaxes the palette-size assertion from `== L` to `>= L`, so a larger shared palette is accepted as long as every label index in `Y` is within bounds.
- Fixes a pre-existing bug in the error message: the second format argument was `X.shape[1]` (the coordinate dimensionality, 2 or 3) instead of `X.shape[0]` (the number of data points).
- Adds an explicit `Y.max() <= mc.shape[0]` guard in the palette path with a clear, actionable error message.

The per-point colour path (`mc.shape[0] == X.shape[0]`) is completely unchanged.

## Motivation and Context

Fixes #357. Users want to define one colour scheme and pass it to several scatter plots; the strict `== L` check forced them to slice the palette for every call.

## How Has This Been Tested?

- `python -m py_compile py/visdom/__init__.py` — no syntax errors.
- Manually verified that an `mc` array of size `> max(Y)` no longer raises, and that an `mc` array smaller than `max(Y)` still raises with a clear message.
- Existing scatter + markercolor behaviour remains valid (the change is a pure relaxation).
- `python -m black --check py` — passes.

N/A — internal validation logic change only.

## Types of changes

## Checklist:

## Summary by Sourcery

Relax scatter marker-color palette validation and add support for a configurable base URL when running behind a proxy.

New Features:
- Support a configurable VISDOM_BASE_URL so the frontend uses the correct base path when served behind a proxy.

Bug Fixes:
- Relax scatter marker-color palette size constraints to allow palettes larger than the maximum label while still validating label indices and fixing the error message dimensions.

Enhancements:
- Propagate base URL configuration from the Python server into the rendered HTML and JavaScript API, ensuring consistent URL resolution across views.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,6 +40,7 @@
     }
   },
   "globals": {
-    "Plotly": "readonly"
+    "Plotly": "readonly",
+    "VISDOM_BASE_URL": "readonly"
   }
 }
--- a/js/api/ApiProvider.js
+++ b/js/api/ApiProvider.js
@@ -14,9 +14,22 @@ const ApiProvider = ({ children }) => {
   // helper functions //
   // ---------------- //
 
-  // Normalize window.location by removing specific path segments
-  // and ensuring the pathname ends with a '/'
+  // Use server-configured base URL when behind a proxy (e.g. /path1); otherwise
+  // derive from pathname so static and socket URLs use the correct path.
   const correctPathname = () => {
+    if (
+      typeof VISDOM_BASE_URL !== 'undefined' &&
+      VISDOM_BASE_URL &&
+      VISDOM_BASE_URL !== '/'
+    ) {
+      return VISDOM_BASE_URL.slice(-1) === '/' ? VISDOM_BASE_URL : VISDOM_BASE_URL + '/';
+    }
+    if (
+      typeof VISDOM_BASE_URL !== 'undefined' &&
+      (VISDOM_BASE_URL === '/' || VISDOM_BASE_URL === '')
+    ) {
+      return '/';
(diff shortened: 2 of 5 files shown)


marcoag (changes requested):
The markercolor palette fix is the right call, but this PR mixes three unrelated changes:

1. The scatter `_markerColorCheck` fix (the actual goal of this PR)
2. The `VISDOM_BASE_URL` proxy fix — this is already covered by #978
3. Black formatting changes across `__init__.py` (ternary reformatting, parenthesised expressions) that are unrelated to the markercolor logic

Please modify with only the `_markerColorCheck` change. The proxy code should stay in #978, and the formatting changes should be reverted to keep the diff focused.

mariobehling:
Thanks for the contribution!

A process note.

mariobehling (changes requested):
Please address change requests.

ArnavBallinCode (commented):
Thank you for your contribution, closing this pr since it has been inactive since a month

PR #977, opened 2026-02-23
feat(scatter): support float Y labels in scatter plots

### LINKED ISSUE #651: Float numbers as labels at scatter plot
I would like to plot a scatter plot with labels that are floats and not int values. I think many data scientists have the same problem and it would be great if you Y data could also be float.
The visdom experienced an exception while running: AssertionError('labels should be integers')
If you use float values as y values for a scatter plot  from the example demo.py https://github.com/facebookresearch/visdom/blob/master/example/demo.py
I tried to do this workaround
    for i in range(0, len(data["labels"])):
        data["labels"] = int(data["labels"] * 100000)
    #Y = np.random.rand(100)
    old_scatter = viz.scatter(

        Y = data["labels"],
        X = [list(a) for a in zip(data["top"], data["middle"])],

which resulted in a The visdom experienced an exception while running: TypeError("cannot convert the series to <class 'int'>")

**Additional context**

devs6186:
## Description

`scatter()` previously hard-asserted that `Y` must contain integers ≥ 1, raising `AssertionError('labels should be integers')` for any float label array. This prevented common use-cases such as colouring scatter points by a continuous metric (dice score, F1, loss, etc.).

This PR extends `scatter()` to accept arbitrary numeric `Y` arrays:

**Integer fast-path (unchanged):** if all `Y` values are already whole numbers and `Y.min() >= 1`, the existing code runs exactly as before.

**Float / arbitrary path (new):**
1. `label_values = np.unique(Y)` captures the sorted original values.
2. `Y` is remapped to sequential 1-based indices via `np.searchsorted`, so that `markercolor`, `linecolor`, `dash`, and all palette lookups (which expect integer indices) continue to work correctly.
3. `labels = np.arange(1, K+1)` where `K = len(label_values)`.
4. Trace names show the real label value (`"0.1"`, `"0.5"`) instead of the internal index (`"1"`, `"2"`).

`label_values` is `None` on the integer path and when `Y` is not provided, so the trace-name fallback remains `str(k)` — fully backward-compatible.

## Motivation and Context

Fixes #651. Users want to use float-valued data (e.g. confidence scores, dice coefficients) directly as scatter group labels without manually converting them to integers.

## How Has This Been Tested?

- `python -m py_compile py/visdom/__init__.py` — no syntax errors.
- `python -m black --check py` — clean.
- Manually verified:
  - Integer labels `[1, 2, 3]`: trace names "1", "2", "3" (unchanged). ✓
  - Non-contiguous integer labels `[1, 3, 5]`: trace names "1", "3", "5" (unchanged via `str(k)` path). ✓
  - Float labels `[0.1, 0.5, 0.9]`: remapped to indices 1, 2, 3; trace names "0.1", "0.5", "0.9". ✓
  - Zero-origin integer labels `[0, 1, 2]`: remapped to 1, 2, 3; trace names "0", "1", "2". ✓
  - Single-trace `Y=None` + `name=`: unchanged behaviour. ✓

N/A — internal label-mapping change; legend text update visible in Plotly.

## Types of changes

## Checklist:

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -60,9 +60,7 @@ try:
         perplexity = (
             50
             if num_entities >= 150
-            else num_entities // 3
-            if num_entities >= 21
-            else 7
+            else num_entities // 3 if num_entities >= 21 else 7
         )
         Y = bhtsne.run_bh_tsne(
             X, initial_dims=X.shape[1], perplexity=perplexity, verbose=True
@@ -179,9 +177,11 @@ def _axisformat(xy, opts):
         return {
             "type": opts.get(xy + "type"),
             "title": opts.get(xy + "label"),
-            "range": [opts.get(xy + "tickmin"), opts.get(xy + "tickmax")]
-            if has_ticks
-            else None,
+            "range": (
+                [opts.get(xy + "tickmin"), opts.get(xy + "tickmax")]
+                if has_ticks
+                else None
+            ),
             "tickvals": opts.get(xy + "tickvals"),
             "ticktext": opts.get(xy + "ticklabels"),
             "dtick": opts.get(xy + "tickstep"),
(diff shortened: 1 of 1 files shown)


mariobehling:
Thanks for the contribution!

A process note.

mariobehling:
No, sourcery-ai upgrades are not required to contribute here. Sourcery will reset every week. So, do not worry.

Manik-Khajuria-5 (changes requested):
@devs6186 Can u please fix linitng error and also address copilot reviews

Jayantparashar10:
Hey @devs6186, thanks for the contribution and for identifying this gap — float Y label support is a real missing feature. This work has been picked up and extended in #1277  which addresses the review concerns raised here.

PR #1018, opened 2026-03-11
Fix shape mismatch in visdom.line when Y has shape (N,1)

KushDutta23:
Flatten Y when it has shape (N,1) to prevent incorrect reshaping that causes X to be tiled incorrectly and results in a shape mismatch error.

## Types of changes

## Checklist:

## Summary by Sourcery

Bug Fixes:
- Normalize Y with shape (N,1) to a 1D array in visdom.line so that X tiling and shape checks no longer fail for single-column data.


KushDutta23:
Thanks for the clarification! I see the issue with the ordering now — reshaping before tiling makes sense. I’ll take a look at #1064 and pick up another issue.

PR #1031, opened 2026-03-13
fix(ci): replace deprecated set-output in PyPI workflow

Celestialglitch:
## Description

Updates the PyPI deployment workflow to replace the deprecated `::set-output` command with the modern `$GITHUB_OUTPUT` environment file approach.

Changed:
```yaml
echo "::set-output name=TAG_NAME::$(cat py/visdom/VERSION)"
echo "::set-output name=TAG_EXISTS::$(git tag | grep v$(cat py/visdom/VERSION) | wc -l)"
```

To:
```yaml
echo "TAG_NAME=$(cat py/visdom/VERSION)" >> $GITHUB_OUTPUT
echo "TAG_EXISTS=$(git tag | grep v$(cat py/visdom/VERSION) | wc -l)" >> $GITHUB_OUTPUT
```

Also applies Prettier formatting according to the project's `.prettierrc`.

## Motivation and Context

GitHub deprecated the `::set-output` command in October 2022, with enforcement starting June 1, 2023. Workflows still using this syntax can fail during execution.

This change migrates the workflow to the recommended `$GITHUB_OUTPUT` mechanism to ensure the PyPI release pipeline continues to work correctly.

Reference: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

## How Has This Been Tested?

- Verified YAML syntax
- Confirmed output variables are used correctly in subsequent steps
- Compared the change with GitHub's official migration guidance

Note: Full validation requires triggering the release workflow, which happens when the VERSION file is updated.

N/A

## Types of changes

## Checklist:

## Summary by Sourcery

Update the PyPI deployment GitHub Actions workflow to use the recommended output mechanism and tidy formatting.

CI:
- Replace deprecated ::set-output commands with $GITHUB_OUTPUT in the PyPI release workflow to maintain compatibility with current GitHub Actions.
- Reformat the pypi.yml workflow file to align with the project's YAML and quoting style.


Manik-Khajuria-5 (changes requested):
@Celestialglitch There are several cosmetic changes included along with the functional fix.  grep -Fx will be a much safer choice than just grep.Can u please go through the PR again  and address the copilot reveiws.

vedansh-5 (commented):
Thanks @Celestialglitch for the contribution. The migration looks good overall.

One improvement worth making is the current tag existence check. Since `grep` treats the version string as a regex and the match is not anchored, it may lead to false positives. For example, if we are checking for `v1.2.3`, it could also match tags like `backup-v1.2.3`.

It would be safer to use an exact tag lookup via `git tag --list`.

Saksham-Sirohi:
@vedansh-5 @Manik-Khajuria-5 I have changed the base branch for the PR

.github/workflows/pypi.yml
```diff
@@ -21,8 +21,8 @@ jobs:
           git tag --list
       - name: Retrieve version
         run: |
-          echo "::set-output name=TAG_NAME::$(cat py/visdom/VERSION)"
-          echo "::set-output name=TAG_EXISTS::$(git tag | grep v$(cat py/visdom/VERSION) | wc -l)"
+          echo "TAG_NAME=$(cat py/visdom/VERSION)" >> $GITHUB_OUTPUT
+          echo "TAG_EXISTS=$(git tag | grep v$(cat py/visdom/VERSION) | wc -l)" >> $GITHUB_OUTPUT
```
rajnisht7, line 25: @Celestialglitch the code looks but kindly use `"${GITHUB_OUTPUT}"` instead of `$GITHUB_OUTPUT` as good practice

tonypzy:
Could you explain the necessity of bumping `style-loader` from `3.3.1` to `4.0.0`? I'd suggest reverting this specific change to avoid potential Node.js version conflicts. Our CI currently runs on Node 16, whereas `style-loader@4.0.0` explicitly requires Node.js >= 18.12.0. Keeping it at `3.3.1` would be the safest approach here.
```
tonypzy@Mac visdom % npm view style-loader@4.0.0 engines
{ node: '>= 18.12.0' }
tonypzy@Mac visdom % npm view style-loader@3.3.1 engines
{ node: '>= 12.13.0' }
```

Manik-Khajuria-5:
@Celestialglitch Thanks for contribution . We are closing this PR as it does not align with the intent

Issue #688, opened 2019-11-13
Invalid reshaping of Y for lineplots sharing the same X-axis


antonmeskildsen:
**Bug Description**
When plotting a line (via visdom.line) I get an error of incompatible shapes for X and Y when using inputs Y of size Nx1 and X of size N. Specifically, I localized the problem to line 1688 of __init__.py where Y is reshaped to size 1xN if only one line is used (i.e. Y originally has shape Nx1). This activates the if condition on line 1691 which makes X tiled to shape NxN. This causes a problem on line 1694 since X and Y now don't have the same shape. This is where the exception occurs and crashes my program!

I can 'fix' the error by changing line 1688 to: `Y = Y.reshape(Y.shape[0])`. Of course, I am not sure whether this is actually necessary or whether my usage of Visdom.line is the problem. I have observed the same code to be working a few months ago (August) without any changes.

A final note. The first time I use the plot function where Y has size 1x1 it does work correctly. I suspect this is due to line 1692 where X is tiled if X and Y have different ndims. 

**Reproduction Steps**
Enter steps to reproduce the behavior:
1. Create vector X of size N with simple increasing values, e.g. by range(N).
2. Create vector Y of size Nx1 with random values, e.g. by numpy.random.uniform(size=(N, 1))
3. Plot line in visdom using the Visdom.line method. The update parameter is set to 'replace'.

Example code:
`data = np.vstack([m[self.metric] for m in self.loggers]).T
        if self.x_metric == 'auto':
            x = np.linspace(1, data.shape[0], data.shape[0])
        else:
            x = np.vstack([m[self.x_metric] for m in self.loggers]).T

        self.win = self.vis.line(
            Y=data,
            X=x,
            win=self.win,
            update='replace' if self.win else None,
            opts=dict(
                legend=self.legend,
                markers=True,
                xlabel=self.units,
                ylabel=self.metric,
                title=self.metric,
            )
        )`

**Expected behavior**
To get a single line as output without errors.

**Client logs:**
For issues that make it to the point of reaching the frontend in a browser, please include the javascript logs from that browser. In Chrome, javascript logs can be found via View -> Developer -> JavaScript Console.

**Server logs:**
`Checking for scripts.
It's Alive!
INFO:root:Application Started
You can navigate to http://localhost:8095
INFO:tornado.access:304 GET / (::1) 1237.00ms
INFO:tornado.access:101 GET /socket (::1) 1.00ms
INFO:root:Opened new socket from ip: ::1
[...]

Issue #761, opened 2020-10-06
Plotting for Y.ndim == 2 with Y.shape[1] == 1 leads to inconsistent shapes


tdavchev:
https://github.com/facebookresearch/visdom/blob/7b1203a0aa43a72a6e3dc2271ddba41f9cc960ce/py/visdom/__init__.py#L1692

The reference code above combined with an X.ndim == 1 leads to wrong tiling of X
e.g. 

Consider Y with shape Mx1 and X with shape M. After reverting Y on the referenced line above we get Y with shape 1xM and X with shape M, then
https://github.com/facebookresearch/visdom/blob/7b1203a0aa43a72a6e3dc2271ddba41f9cc960ce/py/visdom/__init__.py#L1696 leads to tiling X to MxM which is mismatching Y.

Issue is overcome if I do not transpose Y if it's second shape is 1. This bug is related to a workaround of:
https://github.com/facebookresearch/visdom/commit/40c2a3ef0e4898177dd3bcad1063e8a31cb28fb1

Krish-bhardwaj:
I am interested to fix this issue. I request you to allow me to fix it. can you assign me this issue

PR #1064, opened 2026-03-19
fix-line-nx1-y

shanooo773:
What changed: 
 Visdom.line() now treats Y shaped (N,1) as a single line when X is (N,).
Why: 
 fixes crash/mismatch when plotting a single line with shared x-axis.
Issue: 
 references #688.

## Summary by Sourcery

Bug Fixes:
- Treat Y arrays shaped (N,1) as a single 1D line when X is (N,) to prevent crashes and shape mismatch errors in visdom.line with a shared x-axis.


marcoag (changes requested):
The fix looks correct. A few things before merging:

1. Please update the PR description to use `Fixes #688` and `Fixes #761` so both issues are closed on merge.
2. A unit test for the (N,1) case would help prevent regressions.
3. Minor: `Y[:, 0]` on `np.matrix` inputs stays 2D — `np.asarray(Y)[:, 0]` would be safer, though `np.matrix` is deprecated so this is low priority.

mariobehling (changes requested):
Please address the change requests.

sumedhaagh:
Hi! I noticed there are some requested changes (like adding a unit test and minor fixes).  
If there’s no active work ongoing, I’d be happy to help complete this PR.

PR #1142, opened 2026-04-01
Fix lazy env JSON path resolution

### LINKED ISSUE #1141: Lazy environment loading uses incorrect path <env>/.json instead of <env>.json
Lazy environment loading appears to construct the on-disk path incorrectly in two places, causing Visdom to look for:

`<env_path>/<eid>/.json`

instead of the file layout used elsewhere in the codebase:

`<env_path>/<eid>.json`

This can prevent environments that are not already loaded in memory from being loaded from disk correctly.

## Affected code

The issue appears in these locations:

- `py/visdom/utils/server_utils.py` in `compare_envs`
- `py/visdom/utils/server_utils.py` in `load_env`

Current code uses:

```python
p = os.path.join(env_path, eid.strip(), ".json")
```

which resolves to a directory-style path such as:

```text
~/.visdom/main/.json
```

## Why this looks incorrect

Other parts of the codebase consistently use flat JSON files named `<env>.json`:

- `serialize_env` writes to:
```python
os.path.join(env_path, "{0}.json".format(env_id))
```

- environment deletion also assumes:
```python
os.path.join(self.env_path, "{0}.json".format(msg["eid"]))
```

- app startup scans `env_path` for `*.json` files directly

So the lazy-load path shape seems inconsistent with how environments are actually stored.

## Expected behavior

When a saved environment is not already loaded in memory, Visdom should attempt to load it from:

```text
<env_path>/<eid>.json
```

## Actual behavior

Visdom appears to try loading from:

```text
<env_path>/<eid>/.json
```

which does not match the storage format used elsewhere.

## Suggested fix

Replace:

```python
os.path.join(env_path, eid.strip(), ".json")
```

with something like:

```python
os.path.join(env_path, "{}.json".format(eid.strip()))
```

in both `compare_envs` and `load_env`.

SxxAq:
## Summary

Fix lazy environment loading to use the same `<env>.json` file layout used elsewhere in Visdom.

## Changes
- update `compare_envs` to load envs from `os.path.join(env_path, "{}.json".format(eid.strip()))`
- update `load_env` to use the same path construction

## Why
Environments are serialized, deleted, and discovered as flat JSON files like `<env>.json`, but these two lazy-load paths were incorrectly constructing `<env>/.json`.

- closes: #1141

## Summary by Sourcery

Bug Fixes:
- Correct lazy loading paths in `compare_envs` and `load_env` to read environments from `<env>.json` files instead of `<env>/.json`.

### DIFF
--- a/py/visdom/utils/server_utils.py
+++ b/py/visdom/utils/server_utils.py
@@ -244,7 +244,7 @@ def compare_envs(state, eids, socket, env_path=DEFAULT_ENV_PATH):
         if eid in state:
             envs[eid] = state.get(eid)
         elif env_path is not None:
-            p = os.path.join(env_path, eid.strip(), ".json")
+            p = os.path.join(env_path, "{}.json".format(eid.strip()))
             if os.path.exists(p):
                 with open(p, "r") as fn:
                     env = tornado.escape.json_decode(fn.read())
@@ -382,7 +382,7 @@ def load_env(state, eid, socket, env_path=DEFAULT_ENV_PATH):
     if eid in state:
         env = state.get(eid)
     elif env_path is not None:
-        p = os.path.join(env_path, eid.strip(), ".json")
+        p = os.path.join(env_path, "{}.json".format(eid.strip()))
         if os.path.exists(p):
             with open(p, "r") as fn:
                 env = tornado.escape.json_decode(fn.read())


Manik-Khajuria-5 (approved):
@SxxAq I checked the codebase and changes LGTM ! Thanks for contribution

Manik-Khajuria-5:
@SxxAq Thanks for contibution We already have this implemented

PR #1147, opened 2026-04-01
Fix layout message payload mismatch and command typo in ApiProvider

Nikita2106-05:
## Issue
There is a mismatch between the payload sent from ApiProvider and what `onLayoutMessage` expects.

ApiProvider was sending:
{ cmd: ..., update: ... }

But main.js expects:
{ data: ..., update: ... }

This caused `data` to be undefined at runtime, breaking layout updates.

## Fix
- Renamed `cmd` → `data` in layout message handler
- Fixed typo `commmand` → `command` in both layout and window handlers
- Replaced `==` with strict equality `===`

## Result
- Layout updates now correctly pass data to `parseLayoutsFromServer`
- Prevents silent bugs due to typo in command field
- Improves consistency and reliability of message handling

## Summary by Sourcery

Fix message payload shape and command field handling for window and layout messages in ApiProvider to restore correct layout updates and improve reliability.

Bug Fixes:
- Align layout message payload with onLayoutMessage expectations so layout data is correctly passed through.
- Correct the command field typo in window and layout message handling to prevent silent failures in update detection.
- Use strict equality for command comparisons to avoid unexpected truthiness issues in message routing.

### DIFF
--- a/js/api/ApiProvider.js
+++ b/js/api/ApiProvider.js
@@ -109,14 +109,14 @@ const ApiProvider = ({ children }) => {
           readonly: cmd.readonly,
         }));
         break;
-      case 'pane':
-      case 'window':
-      case 'window_update':
-        apiHandlers.current.onWindowMessage({
-          cmd: cmd,
-          update: cmd.commmand == 'window_update',
-        });
-        break;
+     case 'pane':
+case 'window':
+case 'window_update':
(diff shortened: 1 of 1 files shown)


rajnisht7:
@Nikita2106-05 thank you for your contribution, we will be moving forward with #1149, there are many other open issues which you can look into. Thank you again for your contribution.

PR #1148, opened 2026-04-01
Fix lazy environment loading path (use <env>.json instead of directory path)

KushDutta23:
### Fix incorrect environment file path in lazy loading

Lazy environment loading constructs an incorrect file path:

    <env_path>/<eid>/.json

However, environments are stored as flat JSON files:

    <env_path>/<eid>.json

### Changes
- Updated path construction in:
  - compare_envs
  - load_env

### Impact
- Fixes failure to load environments not already in memory
- Aligns lazy loading logic with serialize_env and other parts of the codebase

This restores correct behavior when loading saved environments from disk.

## Summary by Sourcery

Bug Fixes:
- Correct environment file path construction in lazy loading helpers so they load <env>.json files instead of treating the environment ID as a directory.

### DIFF
--- a/py/visdom/utils/server_utils.py
+++ b/py/visdom/utils/server_utils.py
@@ -244,7 +244,7 @@ def compare_envs(state, eids, socket, env_path=DEFAULT_ENV_PATH):
         if eid in state:
             envs[eid] = state.get(eid)
         elif env_path is not None:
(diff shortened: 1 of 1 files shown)


mariobehling:
Please open an issue first and explain the problem in the issue. Then open a PR and link it to the issue.

PR #1151, opened 2026-04-01
fix: correct typo preventing window_update and layout_update

nightcityblade:
Fixes #1126

The update flag in `ApiProvider.js` was computed using `cmd.commmand` (3 m's) instead of `cmd.command`, causing `window_update` and `layout_update` incremental updates to never be applied.

**Changes:**
- Line 117: `cmd.commmand` → `cmd.command` (window_update)
- Line 130: `cmd.commmand` → `cmd.command` (layout_update)

## Summary by Sourcery

Fix incremental update handling for window and layout messages in the API provider.

Bug Fixes:
- Correct the update flag computation for window_update messages so they are applied incrementally.
- Correct the update flag computation for layout_update messages so they are applied incrementally.

### DIFF
--- a/js/api/ApiProvider.js
+++ b/js/api/ApiProvider.js
@@ -114,7 +114,7 @@ const ApiProvider = ({ children }) => {
       case 'window_update':
         apiHandlers.current.onWindowMessage({
           cmd: cmd,
-          update: cmd.command == 'window_update',
+          update: cmd.command === 'window_update',
         });
         break;
       case 'reload':
@@ -127,7 +127,7 @@ const ApiProvider = ({ children }) => {
       case 'layout_update':
         apiHandlers.current.onLayoutMessage({
           cmd: cmd.data,
-          update: cmd.command == 'layout_update',
+          update: cmd.command === 'layout_update',
         });
         break;
       case 'env_update':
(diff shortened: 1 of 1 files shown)


nightcityblade:
Good point from Sourcery! Updated both comparisons to use strict equality (`===`) instead of loose equality (`==`). Thanks!

mariobehling:
Thanks for the contribution!

A process note.

nightcityblade:
Thanks for the note. I understand the repository preference, but I do not have Copilot enabled on this account, so I cannot trigger the automatic Copilot review from my side. The code change itself is already updated to address the earlier review feedback, so if there is any additional manual feedback on the patch I’m happy to address it promptly.

Manik-Khajuria-5:
@nightcityblade
many test are failing and no screenshot or video is being provided can u please go through it once again

nightcityblade:
I dug into the failing CI. The failures are real, and they are not caused by missing screenshots alone. The visual regression run is crashing in app code with  inside the layout message handling path in , which is why the websocket / functional jobs cascade red too. I’m treating this as a genuine follow-up item and will revisit the message-shape handling before pushing anything else here.

nightcityblade:
Correction to my last note because shell quoting mangled two file references: the failing visual regression job is throwing `TypeError: Cannot read properties of undefined (reading 'length')`, and the stack points into the layout message handling path in `main.js`. So there is a real runtime regression to fix here, not just missing screenshot evidence.

Manik-Khajuria-5 (changes requested):
@nightcityblade Can u please fix failing test

Manik-Khajuria-5:
@nightcityblade Due to inactivity we are closing this PR in favour of #1149 Thank u for ur contribution

PR #1152, opened 2026-04-01
Fix/typo in embeddingpane.js

rajnisht7:
## Description
Fixed two bugs in `EmbeddingsPane.js`:
- inside onEvent event was used instead of e
- in onGoBack `this.props.env` was mentioned instead of `this.props.envID`

## Motivation and Context
- first bug causing incorrect key event
- second bug was breadking `onGoBack`

## How Has This Been Tested?
tested locally with events 
verified that `sendEmbeddingPop` now receives proper `envID`

## Types of changes

## Checklist:

## Summary by Sourcery

Fix incorrect event handling and environment ID usage in the embeddings pane.

Bug Fixes:
- Correct key press handling in the embeddings pane by using the actual event parameter.
- Pass the correct environment identifier to the go-back handler in the embeddings pane.


Manik-Khajuria-5 (approved):
@rajnisht7 Though changes were good but please make sure to write proper description for PR so it would be easy to figure out what PR os trying to solve and make sure to add video 

Image

Jayantparashar10 (approved):
Tested locally. Both fixes are legit bugs:

-  `onKeyPress` uses e as the parameter but the original code was calling `event.key` `event.key` Code event here falls back to `window.event` which is deprecated and often undefined. Verified in browser DevTools: after the fix, key presses correctly send the actual key value over WebSocket.                                                                                                                                                                                          
-   `this.props.env` doesn't exist on this component the prop is envID. So onGoBack was passing undefined to `sendEmbeddingPop`. Fix is correct.
                                                                                                                                                                                                                                             Good catch on both. LGTM.         


<img width="1567" height="414" alt="Image" src="https://github.com/user-attachments/assets/32360c07-5f69-479d-bfaf-850e1768c4e9" />

vedansh-5 (approved):

Saksham-Sirohi (approved):

PR #1154, opened 2026-04-01
origin fix-typo-length

sumedhaagh:
### Description

Replaced a TODO comment in `ApiProvider.js` with a warning log to handle the case when the socket connection is not initialized.

### Changes

* Added `console.warn` to notify when `_socket.current` is not available
* Removed the TODO comment

### Type of Change

* Code refactor / cleanup

This improves debugging without affecting existing functionality.

### DIFF
--- a/js/api/ApiProvider.js
+++ b/js/api/ApiProvider.js
@@ -36,7 +36,7 @@ const ApiProvider = ({ children }) => {
   // Send a low-level message to the server
   const sendSocketMessage = (data) => {
     if (!_socket.current) {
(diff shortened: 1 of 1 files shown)


sumedhaagh:
Added a warning log for the case when the socket connection is not initialized in ApiProvider. This replaces the existing TODO comment and improvess  debugging without affecting functionality.

sumedhaagh:
Closing this PR as it contains unrelated changes. Created a clean PR with only the intended fix.

PR #1221, opened 2026-04-19
fix zero division in quiver normalization for zero magnitude vectors

rajnisht7:
## Description
this PR fixes runtime issue in the quiver  normalization  where division by zero can occur when all vectors have zero magnitudes
it also  handles non-finite values  like NaN, Inf in the normalization

## Motivation and Context
in current implementation if the input vector field contains only zero vectors, the computed magnitude becomes zero, resulting in runtime errors.

## How Has This Been Tested?
tested it by simulating the normalisation logic, also with zero and non-finite values

## Types of changes

## Checklist:

## Summary by Sourcery

Bug Fixes:
- Avoid division-by-zero and non-finite operations during quiver vector normalization by skipping normalization when the computed magnitude is invalid or non-positive and issuing a warning.


py/visdom/__init__.py
```diff
@@ -2179,11 +2179,36 @@ def quiver(self, X, Y, gridX=None, gridY=None, win=None, env=None, opts=None):
...
+
+            if finite_mag.size == 0:
+                warnings.warn(
+                    "Skipping quiver normalization: all magnitudes are non-finite (NaN or Inf)",
+                    RuntimeWarning,
+                )
+            else:
+                max_mag = finite_mag.max()
+
+                if max_mag <= 0:
```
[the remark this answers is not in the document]
rajnisht7: added

Manik-Khajuria-5:
@rajnisht7 Please priovide a screenshot or video for This PR

Manik-Khajuria-5:
@rajnisht7 This is way too much code for a simple division-by-zero guard.

py/visdom/__init__.py
```diff
@@ -2179,11 +2179,39 @@ def quiver(self, X, Y, gridX=None, gridY=None, win=None, env=None, opts=None):
...
-            X = X / (magnitude / opts["normalize"])
-            Y = Y / (magnitude / opts["normalize"])
+                isinstance(opts["normalize"], numbers.Number)
+                and opts["normalize"] > 0
+                and np.isfinite(opts["normalize"])
+            ), "opts.normalize should be a finite positive number"
+            magnitude = np.sqrt(np.add(np.multiply(X, X), np.multiply(Y, Y)))
+            finite_mask = np.isfinite(magnitude)
+
+            if not np.any(finite_mask):
```
Manik-Khajuria-5: @rajnisht7 If all input vectors are NaN/Inf, the real problem is upstream the caller is passing garbage data. A warnings.warn that silently skips normalization hides a data
   bug from the user. The original assert + crash behavior was arguably better because it made bad input obvious.

Manik-Khajuria-5:
@rajnisht7  I think u must work on this issue again . Thanks

py/visdom/__init__.py
```diff
@@ -2179,11 +2179,39 @@ def quiver(self, X, Y, gridX=None, gridY=None, win=None, env=None, opts=None):
         # normalize vectors to unit length:
         if opts.get("normalize", False):
             assert (
-                isinstance(opts["normalize"], numbers.Number) and opts["normalize"] > 0
-            ), "opts.normalize should be positive number"
-            magnitude = np.sqrt(np.add(np.multiply(X, X), np.multiply(Y, Y))).max()
```
Manik-Khajuria-5, line 2357: @rajnisht7 Can u please just check one more approach like if we just add magnitude > 0 this might  fix the issue

Manik-Khajuria-5 (changes requested):
@rajnisht7 Thanks for contribution Here is what I get when I tried rendering the issue on dev branch There is no crash can u please specify which issue it solves and provide proper video/screenshot impact of changes of ur PR on visdom Thanks again for contribution 

Image

rajnisht7:
> @rajnisht7 Thanks for contribution Here is what I get when I tried rendering the issue on dev branch There is no crash can u please specify which issue it solves and provide proper video/screenshot impact of changes of ur PR on visdom Thanks again for contribution
> Image

try these three cases:

case 1 : all zero vectors that is
X = np.zeros((3, 3))
Y = np.zeros((3, 3))

case 2: one inf vector  that is
X = np.array([[1., 2.], [3., np.inf]])
Y = np.array([[1., 0.5], [0.5, 0.]])

case 3: one nan vector
X = np.array([[1., 2.], [3., np.nan]])
Y = np.array([[1., 0.5], [0.5, 0.]])

py/visdom/__init__.py
rajnisht7: adding magnitude>0 will only fix all zeros case

Manik-Khajuria-5 (approved):
@rajnisht7 Nice work ! LGTM read all changes and verified 

Image

Jayantparashar10 (approved):
LGTM — fix correctly handles all edge cases (zero vectors, NaN, Inf inputs) without crashing, verified locally with assertions and visual output.


Screenshot 2026-05-21 at 11 56 39 PM

vedansh-5 (approved):
The changes looks fine to me - but I couldn't reproduce the runtime error as mentioned - the result and logs were similar in both dev and feature branch.

``` 
>>> X = np.zeros((3, 3))
>>> Y = np.zeros((3, 3))
>>> vis.quiver(X=X, Y=Y, opts=dict(normalize=True))
'window_3f3cbcc2a076cc'
>>> X = np.array([[1., 2.], [3., np.inf]])
>>> Y = np.array([[1., 0.5], [0.5, 0.]])
>>> vis.quiver(X=X, Y=Y, opts=dict(normalize=True))
'window_3f3cbccd198df4'
>>> X = np.array([[1., 2.], [3., np.nan]])
>>> Y = np.array([[1., 0.5], [0.5, 0.]])
>>> vis.quiver(X=X, Y=Y, opts=dict(normalize=True))
'window_3f3cbcd7f3cff2'
```

Saksham-Sirohi (approved):

PR #979, opened 2026-02-24
Fix WebSocketClosedError on writing to closed sockets (#934)

pankajbaid567:
Fix WebSocketClosedError when writing to disconnected clients (#934)

## Description
This Pull Request prevents the server from crashing when broadcasting to closed WebSocket connections.

When sending data to clients via WebSockets, occasional disconnects leave stale websocket connections on the server side until they are clean. During this window, writing a message to these closed sockets raises a `tornado.websocket.WebSocketClosedError`, leading to an `Uncaught exception POST /events (127.0.0.1)` on the server and crashing the broadcast event. This fixes [#934](https://github.com/fosdem/visdom/issues/934).

**Updates based on PR Feedback:**
- Abstracted the `try...except tornado.websocket.WebSocketClosedError` logic into a single centralized helper function `safe_write_message(socket, msg, context="")` in `server_utils.py`.
- Applied `safe_write_message` across all broadcasting functions in `server_utils.py` and across sockets in `socket_handlers.py`.
- Added a `logging.debug` statement within the exception handler to improve operational visibility without crashing the server.
- Ensured that `write_message` is the only operation inside the try/except block so no other critical errors (like `KeyError` or `AttributeError`) are silently masked.

## Motivation and Context
This ignores the error for closed sockets, allowing the loop to continue broadcasting to other healthy websocket clients without crashing the server.

## How Has This Been Tested?
Ran `demo.py` from the examples to verify that normal plotting and updates work as expected, and simulated ungraceful websocket closures to confirm it resolves the `WebSocketClosedError` without crashing the `POST /events` endpoint.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Code refactor or cleanup (changes to existing code for improved readability or performance)

## Checklist:
- [ ] I adapted the version number under `py/visdom/VERSION` according to [Semantic Versioning](https://semver.org/)
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1244, opened 2026-04-24
Fix WebSocketClosedError when sending to closed socket

### LINKED ISSUE #934: ERROR:tornado.application:Uncaught exception POST /events (127.0.0.1)
I have recently encountered an issue while using Visdom for data visualization, specifically receiving frequent WebSocketClosedError errors when attempting to send data to the client via WebSocket. The problem seems to arise when attempting to write a message to the WebSocket connection, only to find that the connection has already been closed. The detailed error trace is as follows：

image
Given this context, I have several inquiries seeking your assistance:

Are there known scenarios or configuration issues that could lead to unexpected closure of WebSocket connections during data transmission?
Are there any recommended best practices or configuration adjustments to enhance the stability of WebSocket connections?

sumedhaagh:
Fixes #934

This PR prevents crashes when attempting to send messages over a closed WebSocket.

Changes:
- Checks if the socket is alive before sending data
- Avoids WebSocketClosedError
- Falls back to HTTP when socket is not available

This improves stability when WebSocket connections close unexpectedly.

## Summary by Sourcery

Bug Fixes:
- Prevent crashes caused by sending over a closed or dead WebSocket by detecting socket liveness and disabling socket use when necessary.

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -742,6 +742,10 @@ class Visdom(object):
         if not self.send:
             return msg, endpoint
 
+        if self.use_socket and not self.socket_alive:
+            logger.warn("Socket not alive, switching to HTTP")
+            self.use_socket = False
+
         if "win" in msg and msg["win"] is None and create:
             msg["win"] = "window_" + get_rand_id()
 
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -743,8 +743,7 @@ class Visdom(object):
             return msg, endpoint
 
         if self.use_socket and not self.socket_alive:
-            logger.warn("Socket not alive, switching to HTTP")
-            self.use_socket = False
+            logger.warning("Socket not alive, falling back to HTTP for this request")
 
         if "win" in msg and msg["win"] is None and create:
             msg["win"] = "window_" + get_rand_id()


Jayantparashar10 (changes requested):
@sumedhaagh Thanks for tackling #934! The fix prevents the `WebSocketClosedError` crash nicely, but there are a few things worth addressing:

- Permanent fallback: Toggling `self.use_socket = False` globally means all future sends switch to HTTP permanently, even if the socket reconnects later. Consider falling back only for the current request, or adding a reset mechanism when `socket_alive` becomes `True` again.
- Thread safety: The `socket_alive` check followed by the flag toggle is not atomic — if `Visdom` is used from multiple threads, this could lead to race conditions.
- Deprecated logger: Use `logger.warning(...)` instead of `logger.warn(...)`, which has been deprecated since Python 3.2.

Overall the intent is solid — happy to approve once these are cleaned up!

sumedhaagh:
@Jayantparashar10 Thanks for the review! I’ve updated the implementation to avoid permanently disabling `self.use_socket` and changed `logger.warn` to `logger.warning`. The fallback is now handled only for the current request instead of globally switching future requests to HTTP.

Jayantparashar10:
@sumedhaagh  Thanks for the contribution. After reviewing the crash flow for issue #934 , the `WebSocketClosedError` is raised server-side in `broadcast()` in `server_utils.py` when the server tries to push to a disconnected browser tab it's not in the Python client `_send()` path. The `_send()` method always sends via HTTP POST regardless of `socket_alive`, so this check doesn't change any behaviour and the server crash still occurs. PR #979 addresses the correct location. Closing this one. Again thanks for the contribution.

PR #1289, opened 2026-05-10
fix: allow shared markercolor palettes larger than max(Y) in scatter plots

### LINKED ISSUE #357: MarkerColor dimension restrictions are too strict

### LINKED ISSUE #761: Plotting for Y.ndim == 2 with Y.shape[1] == 1 leads to inconsistent shapes

### LINKED ISSUE #762: Bug raises when encountering x = [1.0, 2] and y = [[0.78], [1]]
**Bug Description**
I want to draw two points with each affiliating with one entry denoted as an array of length 1.

Line 1687 of __init__.py will cause side effects on line 1692.
And there comes the shape assertion bug on line 1695.

ArnavBallinCode:
## Description

This PR fixes overly strict palette-size validation in `_markerColorCheck` for scatter plot marker colors.

Previously, Visdom required the palette size to exactly match `max(Y)`:

```python
mc.shape[0] == L
```

This prevented reuse of larger shared palettes across scatter plots with fewer categories.

Example:
- Shared palette size = 5
- Scatter plot categories (`max(Y)`) = 2

The plot incorrectly raised an `AssertionError` even though the palette contained sufficient colors.

This PR relaxes the validation to:

```python
mc.shape[0] >= L
```

allowing scatter plots to accept any palette containing at least the required number of category colors.

Additionally:
- fixes an incorrect error-message dimension reference (`X.shape[1]` → `X.shape[0]`)
- adds an explicit guard for undersized palettes to prevent cryptic `IndexError` failures

---

## Motivation and Context

This change improves scatter plot usability and enables reuse of global/shared color palettes across multiple plots.

Before this fix, users had to manually slice palettes for every scatter call:

```python
markercolor = palette[:max(Y)]
```

This was repetitive, error-prone, and made consistent visualization pipelines harder to maintain.

With this fix:
- larger reusable palettes now work correctly
- category colors remain consistent across plots
- invalid palette configurations produce clearer errors

Fixes #357
Fixes #761 
Fixes #762 

---

## How Has This Been Tested?

Tested locally using a custom Visdom script.

### Test cases covered

1. Shared palette larger than `max(Y)`
   - Palette size = 5
   - Scatter plots using 2, 3, and 4 categories
   - All plots rendered successfully after the fix

2. Exact palette match
   - Existing valid behavior remains unchanged

3. Undersized palette
   - Verified explicit assertion is raised before indexing
   - Confirmed clear actionable error message

4. Error message validation
   - Verified `X.shape[0]` (number of points) is now reported correctly instead of `X.shape[1]`

### Manual verification

Ran:
```bash
python -m visdom.server
python tests/markercolor_bug_demo.py
```

Verified:
- multiple scatter plots successfully reuse the same shared palette
- category colors remain consistent across plots
- no regressions in scatter rendering behavior

---

Screenshot 2026-05-10 at 11 26 42 PM

Shows successful rendering of:
- Dataset A (2 categories)
- Dataset B (3 categories)
- Dataset C (4 categories)

all using the same shared 5-color palette.

---

## Types of changes

---

## Checklist:

## Summary by Sourcery

Relax scatter marker color palette size validation to allow palettes with at least as many colors as categories and improve error handling for invalid configurations.

Bug Fixes:
- Allow marker color palettes larger than the number of categories in scatter plots instead of requiring an exact size match.
- Correct the marker color validation error message to reference the number of points using X.shape[0] instead of X.shape[1].
- Add a guard that raises a clear assertion when the marker color palette is smaller than max(Y) to prevent obscure index errors.


Jayantparashar10 (changes requested):
>assert Y.max() <= mc.shape[0]

The assert `assert (Y.max() <= mc.shape[0])` can never fire — the first assert already requires `mc.shape[0] >= L` where `L = int(Y.max())`, so by the time execution reaches this block, `Y.max() <= mc.shape[0]` is already guaranteed. You can remove it the first assert handles undersized palettes with a clearer message anyway.

Manik-Khajuria-5 (changes requested):
@ArnavBallinCode Can u please fix the pylinter ci failure

Manik-Khajuria-5 (approved):
@ArnavBallinCode Nice works Have read whole changes and verified  LGTM! 

The validation now accepts as many colors as user wants 

Image

vedansh-5 (approved):
LGTM!

Image

Saksham-Sirohi (approved):

PR #1297, opened 2026-05-15
perf: optimize update_packet to avoid full deepcopy overhead

tonypzy:
## Description
This PR optimizes the performance of `update_packet` in `web_handlers.py` by avoiding a full `copy.deepcopy` of the entire window packet. Instead of deep-copying the entire `p` dictionary (which includes large immutable structures like UI configurations and layouts), it uses a **shallow copy** (`p.copy()`) to capture all top-level keys, and limits `deepcopy` to only the nested mutable fields (`content` and `old_content`) that are modified in-place during updates.

## Motivation and Context

This addresses the existing `TODO` comment in `web_handlers.py` regarding `make_patch` not being high performance.

#### Functional Correctness:
Execute `example/demo.py` to verify that all the panes are rendered correctly.

image

## Types of changes

## Checklist:


tonypzy:
@vedansh-5 @rajnisht7 @Manik-Khajuria-5 @Jayantparashar10 Please review.

ArnavBallinCode (commented):
The optimization looks good overall, but I have one concern regarding maintainability/correctness.

`MUTABLE_KEYS` creates an implicit dependency on the behavior of `UpdateHandler.update()`. Previously, `jsonpatch.make_patch()` operated on a deepcopy of the entire packet, so any future field mutation was automatically captured. With a hardcoded list, if `UpdateHandler.update()` starts modifying a new key later and `MUTABLE_KEYS` is not updated accordingly, the generated patch could silently miss changes.

Would it make sense to derive these keys dynamically, or at least centralize/document this coupling more explicitly?

tonypzy:
> The optimization looks good overall, but I have one concern regarding maintainability/correctness.
> 
> `MUTABLE_KEYS` creates an implicit dependency on the behavior of `UpdateHandler.update()`. Previously, `jsonpatch.make_patch()` operated on a deepcopy of the entire packet, so any future field mutation was automatically captured. With a hardcoded list, if `UpdateHandler.update()` starts modifying a new key later and `MUTABLE_KEYS` is not updated accordingly, the generated patch could silently miss changes.
> 
> Would it make sense to derive these keys dynamically, or at least centralize/document this coupling more explicitly?

Thanks for your pointing this out.

1. I think dynamic derivation will revert to a full `deepcopy(p)`, which is exactly what this optimization aims to avoid.
2. I fully agreed to document to let other developers understand this coupling, I have added the comment above `MUTABLE_KEYS` in `update_packet`, as well as in `UpdateHandler.update()` and `update_window()`. Any updates to `p` and its sub-fields are entirely driven by these two methods.

Manik-Khajuria-5 (changes requested):
@tonypzy Overall intent of PR was good But I also have same concern as Arnav right now the keys are hardcoded . either we have to get them dynamically or go with deep copy approach

tonypzy:
> @tonypzy Overall intent of PR was good But I also have same concern as Arnav right now the keys are hardcoded . either we have to get them dynamically or go with deep copy approach

I took your advice and used a shallow `copy()` instead of deepcopying everything. Since shallow copy allows `old_p` and `p` to share references for top-level keys, we no longer need to maintain the `MUTABLE_KEYS` list. I only applied `deepcopy()` to `content` and `old_content` because of their nested structures. I believe this addresses the issue you mentioned.

tonypzy:
> The optimization looks good overall, but I have one concern regarding maintainability/correctness.
> 
> `MUTABLE_KEYS` creates an implicit dependency on the behavior of `UpdateHandler.update()`. Previously, `jsonpatch.make_patch()` operated on a deepcopy of the entire packet, so any future field mutation was automatically captured. With a hardcoded list, if `UpdateHandler.update()` starts modifying a new key later and `MUTABLE_KEYS` is not updated accordingly, the generated patch could silently miss changes.
> 
> Would it make sense to derive these keys dynamically, or at least centralize/document this coupling more explicitly?

I took your advice and used a shallow `copy()` instead of deepcopying everything. Since shallow copy allows `old_p` and `p` to share references for top-level keys, we no longer need to maintain the `MUTABLE_KEYS` list. I only applied `deepcopy()` to `content` and `old_content` because of their nested structures. I believe this addresses the issue you mentioned.

vedansh-5 (approved):
You have revised the hardcoded keys part, LGTM! Thanks @tonypzy

Manik-Khajuria-5 (approved):
LGTM !

ArnavBallinCode (approved):
LGTM

rajnisht7 (approved):

PR #1336, opened 2026-05-27
fix: reduce redundant Plotly redraws during realtime updates

### LINKED ISSUE #1333: Improve high-frequency Plotly rendering stability for realtime updates
## Problem

While reviewing realtime visualization behavior, I noticed that high-frequency plot updates may lead to rendering instability in some scenarios.

Existing reports already mention issues such as:

* flickering heatmaps
* unstable redraw behavior
* browser rendering slowdown
* React tree failures caused by Plotly exceptions

This becomes more visible during:

* rapid tensor streaming
* continuously updating heatmaps
* long-running realtime dashboards
* multiple simultaneously updating panes

---

## Proposed Improvement

I would like to propose improvements focused on realtime rendering stability and update efficiency.

Possible areas:

### 1. Render scheduling/coalescing

* batch rapid updates
* reduce redundant renders
* avoid unnecessary full redraws

### 2. Smarter Plotly update strategy

Evaluate when to use:

* `Plotly.react`
* `extendTraces`
* partial updates
* restyle operations

instead of always triggering expensive redraw paths.

### 3. Error isolation

Introduce pane-level error boundaries so a single Plotly exception does not destabilize larger parts of the UI.

### 4. Immutable update handling

Avoid mutating incoming plot state directly where possible to improve rendering consistency.

---

## Potentially Relevant Areas

* `PlotPane.js`
* `main.js`
* realtime update handlers
* pane rendering pipeline

---

## Expected Benefits

* smoother realtime visualization
* reduced flickering
* improved browser responsiveness
* more stable long-running dashboards
* lower rendering overhead

---

## Additional Notes

I’m currently profiling update behavior locally under rapid streaming workloads to better understand where redraw bottlenecks and instability originate.

Ridanshi:
Fixes #1333.

High-frequency streaming updates could trigger overlapping `Plotly.react` calls because the render effect ran on every React render with no frame-level coalescing. Under rapid update workloads this causes redundant redraws within the same browser paint cycle, wasting CPU and causing visible flicker.

## Changes

**`js/panes/PlotPane.js`**

- Add a `requestAnimationFrame` coalescing guard around the `Plotly.react` call: cancel any pending frame before scheduling a new one. Bursts of updates now collapse into a single draw per browser paint, with no change to the existing render logic.
- Guard `smooth_d.marker` and `smooth_d.marker.line` against `undefined` before setting `marker.line.color` in the smoothing path (crash on traces with no marker object).

**`js/api/ApiProvider.js`**

- Fix triple-typo `cmd.commmand` → use `update: true` directly inside `case 'window_update'` and `case 'layout_update'`. The switch already guarantees the command matches, so the comparison was always false and the version-aware update path never ran.

## Testing

- `npm run lint -- js/panes/PlotPane.js js/api/ApiProvider.js` — no errors
- `npm run build` — succeeds, only the pre-existing bundle-size warnings


rajnisht7 (changes requested):
@Ridanshi this PR have some unnecessary changes which are not at all required, the code structure doesnt follow good practices, also this PR targets multiple changes, kindly make sure one PR targets one issue only. Thank You

rajnisht7:
@Ridanshi kindly address copilot reviews as well

rajnisht7:
@Ridanshi kindly resolve the conflicts also kindly have a look at the logic implemented

Ridanshi:
Thank you for the feedback.

I reviewed the implementation again after resolving the merge conflicts and specifically re-evaluated the logic to ensure the changes remain focused on issue #1333.

The current PR is limited to:

* requestAnimationFrame-based coalescing to avoid redundant Plotly redraws during rapid update bursts
* a guard in the smoothing path to prevent crashes on traces without marker definitions
* simplifying the `window_update` / `layout_update` handling in `ApiProvider.js`

I also verified that the changes do not alter the existing Plotly update flow, zoom/pan preservation behavior, or component lifecycle handling. Lint and build checks continue to pass.

Please let me know if there are any specific parts of the implementation you would like me to revisit further.

Manik-Khajuria-5 (changes requested):
@Ridanshi Thanks for contribution . The original code had a typo cmd.commmand that is triple 'm' so update was always false, meaning incremental updates never worked. But hardcoding true is the wrong fix It will treat every message to full replacement.

Manik-Khajuria-5 (changes requested):
@Ridanshi This PR is scheduled to be close within 2 3 days in case changes are not made Again Thanks for contribution

PR #1352, opened 2026-05-28
fix: atomic env writes and optional periodic autosave

### LINKED ISSUE #1332: Add incremental per-window persistence and autosave support for improved crash recovery
## Problem

Currently environment persistence appears to rely primarily on monolithic environment-level serialization.

For large environments or long-running sessions, this may create several limitations:

* expensive full-environment saves
* increased crash recovery risk
* higher write overhead
* inefficient persistence for frequently updated windows
* difficulty recovering partially corrupted saves

This becomes especially important during long-running ML experiments where Visdom may remain active for many hours.

---

## Proposed Improvement

I would like to propose introducing a more incremental persistence architecture.

Possible direction:

### 1. Per-window persistence

Instead of serializing entire environments repeatedly:

* persist windows independently
* update only modified windows
* reduce save overhead

### 2. Autosave support

Add configurable autosave modes:

* periodic autosave
* write-through autosave
* graceful shutdown save

### 3. Atomic persistence operations

Improve corruption resistance using:

* temporary files
* atomic renames
* recovery-safe writes

### 4. Storage abstraction layer

Potentially introduce:

* `EnvStore`
* `WindowStore`

to separate persistence logic from runtime state management.

---

## Potentially Relevant Areas

* `server_utils.py`
* `app.py`
* environment serialization flow
* save/load handlers

---

## Expected Benefits

* improved crash recovery
* reduced save overhead
* better scalability for large environments
* safer long-running experiment visualization
* easier future persistence extensions

---

## Additional Notes

I am currently reviewing the persistence architecture and migration compatibility considerations before experimenting with implementation approaches.

Ridanshi:
Closes #1332.

## Problem

If the server process is interrupted (crash, OOM, SIGKILL) while writing an environment JSON, the file can be left partially written. On the next startup, `load_state` / `LazyEnvData` would then read a truncated or invalid JSON, silently losing that environment's state.

There is also no way to limit how much work is lost between an unclean exit and the previous explicit save or server shutdown.

## Changes

**`py/visdom/utils/server_utils.py`**

- Add `atomic_write_json(path, data)`: writes JSON to `path + ".tmp"` first, then calls `os.replace()` to rename it into place. On POSIX this rename is atomic; on Windows it is as safe as the platform allows. If the rename fails the temp file is cleaned up and the original is left untouched.
- Update `serialize_env()` to use `atomic_write_json()` for both the normal write path and the long-filename hash-fallback path. The observable output format is unchanged.

**`py/visdom/server/run_server.py`**

- Add `-autosave_interval` CLI argument (float, default `0` = disabled).
- When `autosave_interval > 0`, start a `PeriodicCallback` that calls `serialize_all()` every N seconds. Combined with the existing `atexit` flush on clean shutdown, state loss on an unclean exit is bounded to at most one autosave interval.

No behaviour changes when `-autosave_interval` is not passed.

## Testing

- `python -m pytest test/test_persistence.py test/test_caption.py -v` — 18 tests, all pass
- New tests cover: correct data written, no `.tmp` file left on success, original file preserved when `os.replace` fails, temp file cleaned up on failure, legacy JSON format preserved, only requested envs saved, `.tmp` leftover on disk does not shadow a good env file, `autosave_interval` parameter present with default 0.

## Compatibility

- Existing env JSON files are read and written in the same format.
- `LazyEnvData` and the load path are untouched.
- `autosave_interval=0` (the default) leaves all existing behaviour identical.

### DIFF
--- a/py/visdom/server/run_server.py
+++ b/py/visdom/server/run_server.py
@@ -65,6 +65,7 @@ def start_server(
     use_frontend_client_polling=False,
     bind_local=False,
     eager_data_loading=False,
+    autosave_interval=0,
 ):
     logging.info("Server started")
     app = Application(
@@ -84,6 +85,14 @@ def start_server(
     logging.info(f"Working directory: {os.path.abspath(env_path)}")
 
     atexit.register(serialize_all, app.state, env_path=env_path)
+
+    if autosave_interval > 0 and env_path is not None:
+        autosave_cb = ioloop.PeriodicCallback(
+            lambda: serialize_all(app.state, env_path=env_path),
+            int(autosave_interval * 1000),
+        )
+        autosave_cb.start()
+        logging.info(f"Autosave enabled every {autosave_interval}s")
 
     if "HOSTNAME" in os.environ and hostname == DEFAULT_HOSTNAME:
         hostname = os.environ["HOSTNAME"]
@@ -171,6 +180,13 @@ def main(print_func=None):
         action="store_true",
         help="Load data from filesystem when starting server (and not lazily upon first request).",
     )
+    parser.add_argument(
+        "-autosave_interval",
+        metavar="autosave_interval",
+        type=float,
+        default=0,
+        help="Periodically save all environments every N seconds (0 = disabled).",
+    )
     FLAGS = parser.parse_args()
 
     # Process base_url
@@ -256,6 +272,7 @@ def main(print_func=None):
         use_frontend_client_polling=FLAGS.use_frontend_client_polling,
         bind_local=FLAGS.bind_local,
         eager_data_loading=FLAGS.eager_data_loading,
+        autosave_interval=FLAGS.autosave_interval,
     )
 
 
--- a/py/visdom/utils/server_utils.py
+++ b/py/visdom/utils/server_utils.py
@@ -118,6 +118,26 @@ class LazyEnvData(Mapping):
         return len(self._raw_dict)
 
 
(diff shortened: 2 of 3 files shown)


Manik-Khajuria-5 (changes requested):
@Ridanshi Thanks for contribution . Autosave interval is default set to 0 which means it is disabled so the safety feature most users need is silently off unless they discover it . also what if someone passes 0.001 as interval which becomes a 1ms save loop and thrashes the server. I don't think we should give user a feat to save interval in cli arg

Ridanshi:
Thank you for the feedback.

That's a fair point. My intention was to keep the change backward-compatible by making autosave opt-in, but I agree that a disabled-by-default safety feature may provide limited practical benefit.

I also understand the concern about exposing the interval as a CLI argument, since extremely small values could lead to excessive serialization activity.

I'll revisit the approach and look into simplifying the implementation. One option would be to keep the atomic write improvements while reconsidering how (or whether) autosave should be exposed and configured.

Would you prefer that I:

* remove the autosave functionality entirely and keep the PR focused on atomic persistence safety, or
* rework autosave so it follows a different integration approach within the existing architecture?

I'm happy to adjust the implementation accordingly.

vedansh-5:
Thanks @Ridanshi for your contribution. Currently we do not need the autosave feature - additionally we do not accept important logic changes from first time contributors. Please choose another issue from the list of issues. We are closing this for now.

PR #1353, opened 2026-05-28
fix: guard broadcasts against closed sockets and bound polling queue

### LINKED ISSUE #1331: Improve WebSocket transport reliability with reconnect-safe state synchronization and bounded event queues
## Problem

While studying the Visdom realtime transport pipeline, I noticed that the current WebSocket/event delivery architecture may become unstable under long-running or high-frequency update workloads.

The current implementation appears to maintain shared mutable subscription state across browser clients and Python event sources, while also supporting both polling and WebSocket-based synchronization paths.

Under sustained realtime updates, this can potentially lead to:

* stale subscriptions
* duplicate event delivery
* dropped updates during reconnects
* inconsistent state replay after temporary disconnects
* unbounded event queue growth for slow clients
* browser instability under rapid update workloads

There are also existing reports related to realtime rendering instability and socket-related failures that may be connected to transport-level synchronization behavior.

---

## Proposed Improvement

I would like to propose a transport reliability improvement focused on:

### 1. Reconnect-safe synchronization

* replay missing events after reconnect
* maintain per-client session state
* improve env/window consistency after browser refresh

### 2. Bounded queue/backpressure handling

* prevent unlimited queue accumulation
* drop/coalesce outdated updates safely
* protect slow clients from overwhelming memory growth

### 3. Heartbeat/connection lifecycle improvements

* detect stale/disconnected clients
* cleanup dead subscriptions properly
* reduce ghost subscribers

### 4. Unified transport abstraction

Currently polling and WebSocket paths appear partially duplicated.

A unified event delivery abstraction could improve:

* maintainability
* transport consistency
* testing reliability

---

## Potentially Relevant Areas

* `socket_handlers.py`
* `ApiProvider.js`
* `Legacy.js`
* `server_utils.py`
* websocket subscription/event handling flow

---

## Expected Benefits

* improved realtime stability
* better long-running training session reliability
* lower memory growth risk
* smoother browser reconnect behavior
* more scalable multi-client visualization

---

## Additional Notes

I am currently studying the existing transport/event flow in detail and would be interested in contributing improvements in this area if maintainers think this direction would be valuable.

Ridanshi:
Closes #1331.

## Problem

Three distinct reliability issues in the websocket/polling transport layer:

**1. Unguarded `write_message()` in broadcast functions**

`broadcast()`, `broadcast_envs()`, and `send_to_sources()` iterate over subscriber dicts and call `write_message()` without catching `WebSocketClosedError`. If a client disconnects between the time it is selected for broadcast and the time `write_message()` executes, the exception propagates uncaught to the HTTP handler that triggered the broadcast (e.g. `PostHandler.post()`), causing that request to fail with a 500. The Python Visdom client then sees an HTTP error rather than a successful plot call.

**2. Unbounded polling message queue**

`AnySocketWrapper.messages` was an unbounded `deque()`. A polling client that stops polling but does not disconnect properly (network pause, browser tab backgrounded) accumulates messages until the 15-second monitor evicts it. Under high-frequency update workloads this can grow significantly.

Additionally `get_messages()` replaced `self.messages` with a plain `list` (`self.messages = []`), which would discard the `maxlen` constraint on every poll cycle — making a bounded deque useless after the first drain.

**3. Spurious layout broadcast on every new subscriber connection**

`SocketHandlerOrWrapper.initialize()` called `self.broadcast_layouts()` (to *all* subscribers) every time a new subscriber connected. For WebSocket connections this sent an unnecessary layout update to every already-connected browser. For polling connections it caused the new client to receive layouts twice: once from `initialize()` and once from `open()`, which already calls `self.broadcast_layouts([self])` to send layouts to the new subscriber only.

## Changes

**`py/visdom/server/handlers/socket_handlers.py`**

- Add `MAX_POLLING_QUEUE = 200` constant
- `AnySocketWrapper.initialize()`: `deque(maxlen=MAX_POLLING_QUEUE)` instead of unbounded `deque()`
- `AnySocketWrapper.get_messages()`: drain with `list()` + `.clear()` instead of reassigning `self.messages`, preserving the bounded deque
- `SocketHandlerOrWrapper.initialize()`: remove the `self.broadcast_layouts()` call; `open()` already handles sending layouts to the connecting subscriber
- `VisSocketHandlerOrWrapper.on_close()` and `SocketHandlerOrWrapper.on_close()`: use `dict.pop(sid, None)` directly instead of an O(n) value scan followed by O(1) key removal

**`py/visdom/utils/server_utils.py`**

- Add `import tornado.websocket`
- `broadcast()`: iterate over a snapshot (`list(self.subs)`), add `self.subs.get(s)` null check, and catch `WebSocketClosedError` per subscriber
- `broadcast_envs()`: iterate over a snapshot, encode message once, catch `WebSocketClosedError` per subscriber
- `send_to_sources()`: same pattern

## Testing

```
python -m pytest test/test_websocket.py test/test_caption.py -v
```

19 tests, all pass. New tests cover:
- Bounded queue capacity and oldest-message eviction
- `get_messages()` preserves the bounded deque after draining
- `on_close()` correct key removal and tolerance for already-removed sid
- `broadcast()` and `broadcast_envs()` do not raise on a closed socket and continue delivering to healthy subscribers

## Compatibility

- No message format changes
- No API changes
- Polling and WebSocket paths behave identically to before except for the three defects fixed above
- `MAX_POLLING_QUEUE = 200` is a conservative limit; a polling client receiving 200 queued messages in a single poll cycle was already effectively stalled

### DIFF
--- a/py/visdom/server/handlers/socket_handlers.py
+++ b/py/visdom/server/handlers/socket_handlers.py
@@ -37,6 +37,10 @@ from visdom.utils.server_utils import (
 )
 from visdom.server.defaults import MAX_SOCKET_WAIT
 
+# Maximum number of messages buffered per polling client.
+# Oldest messages are dropped automatically once the limit is reached.
+MAX_POLLING_QUEUE = 200
+
 
 # TODO move the logic that actually parses environments and layouts to
 # new classes in the data_model folder.
@@ -193,7 +197,7 @@ class AnySocketWrapper(AnySocketHandlerOrWrapper):
     def initialize(self, app):
         super().initialize(app)
 
-        self.messages = deque()
+        self.messages = deque(maxlen=MAX_POLLING_QUEUE)
         self.last_read_time = time.time()
         self.open()
         try:
@@ -230,8 +234,8 @@ class AnySocketWrapper(AnySocketHandlerOrWrapper):
 
     def get_messages(self):
         to_send = []
-        messages = self.messages
-        self.messages = []
+        messages = list(self.messages)
+        self.messages.clear()
         for message in messages:
             if isinstance(message, dict):
                 # Not all messages are being formatted the same way (JSON)
@@ -257,8 +261,7 @@ class VisSocketHandlerOrWrapper(AnySocketHandlerOrWrapper):
         self.write_message(json.dumps({"command": "alive", "data": "vis_alive"}))
 
     def on_close(self):
-        if self in list(self.sources.values()):
-            self.sources.pop(self.sid, None)
+        self.sources.pop(self.sid, None)
 
     def on_message(self, message):
         msg = tornado.escape.json_decode(tornado.escape.to_basestring(message))
@@ -323,11 +326,9 @@ class SocketHandlerOrWrapper(AnySocketHandlerOrWrapper):
 
     def initialize(self, app):
         super().initialize(app)
-        self.broadcast_layouts()
 
     def on_close(self):
-        if self in list(self.subs.values()):
-            self.subs.pop(self.sid, None)
+        self.subs.pop(self.sid, None)
 
 
 class SocketHandler(SocketHandlerOrWrapper):
--- a/py/visdom/utils/server_utils.py
+++ b/py/visdom/utils/server_utils.py
@@ -24,6 +24,7 @@ import time
 import re
 import errno
 import tornado.escape
+import tornado.websocket
 from collections import OrderedDict
 
 MAX_ENV_NAME_LEN = 25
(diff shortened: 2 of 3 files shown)


Manik-Khajuria-5 (changes requested):
@Ridanshi Thanks for contribution , These change removes multiple safety guards that were previously protecting connection handling also acc to me max polling queue must not be hardcoded ,Kindly go through the PR once again

Manik-Khajuria-5 (changes requested):
@Ridanshi Can u please remove the test from PR as we are not accepting test right now in visdom Thanks for contribution

Manik-Khajuria-5 (changes requested):
@Ridanshi This PR is scheduled to be closed within 2 3 days in case no changes have been implemneted as suggested within 2 3 days Again thanks for contribution

PR #1368, opened 2026-06-03
fix: warn and clamp float images with values slightly above 1.0

### LINKED ISSUE #602: visdom.images()  sometimes draw all black images
**Bug Description**
I use the mean=[0.485, 0.456, 0.406],std=[0.229, 0.224, 0.225] to  normalize the image.
Then I do some translates on these images. Finally, I restore these images to the normal state and use the visdom.images() to show the images.
The problem is that sometimes these images are all black, I print the image matrix and all values are [0-1],which are right.
Strangely,sometimes these images can draw normally.

**Screenshots**

This is the case:
image

vedansh-5:
## Description
Introduces `_float_img_to_uint8(img)` to handle float image conversion to `uint8`:
- `max <= 1.0`: Scales by 255.
- `1.0 < max <= 1.0 + tol` (where `tol = eps**0.5`): Warns, clamps to `[0, 1]`, and scales by 255.
- `max > 1.0 + tol`: Clips to `[0, 255]` and casts directly.

## Motivation and Context
Fixes #602. Floating point rounding errors (e.g. after normalization transforms) can cause maximum pixel values to end up slightly above `1.0` (like `1.000001`). The previous code skipped scaling in this case and cast directly to `uint8`, resulting in pixel values of `1` (rendering as solid black).

## How has this been tested
- Tested locally 

Before:
image

After
image

## Types of changes
## Checklist:


Manik-Khajuria-5 (approved):
@vedansh-5 LGTM!

py/visdom/__init__.py
```diff
@@ -509,6 +509,24 @@ def _decode_binary_arrays(obj):
...
+    if max_val <= 1.0 + tol:
+        warnings.warn(
+            "Image has float values slightly above 1.0 "
+            "(max={:.6f}). Values will be clamped to "
+            "[0, 1] and scaled to [0, 255].".format(max_val),
+            UserWarning,
+            stacklevel=2,
+        )
+        return np.uint8(np.clip(img, 0.0, 1.0) * 255.0)
+    return np.uint8(img)
```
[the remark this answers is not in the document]
vedansh-5: incorr

py/visdom/__init__.py
```diff
@@ -509,6 +509,24 @@ def _decode_binary_arrays(obj):
     return obj
 
 
+def _float_img_to_uint8(img):
+    """Convert a float image array to uint8."""
+    tol = np.finfo(img.dtype).eps ** 0.5
```
tonypzy: It might be necessary to set up a lower bound which is unrelated with `dtype` concerning `float32` and `float64`.

vedansh-5: Thanks for pointing this out - I have added a floor to the tolerance - the code now handles single point precision and double point precision safely. PTAL @tonypzy

tonypzy (approved):
LGTM

Jayantparashar10 (approved):
LGTM

Saksham-Sirohi (approved):

PR #1434, opened 2026-06-09
Fix plot history tests and ensure local visdom source is used during testing

### LINKED ISSUE #1431: Unit tests fail due to UpdateHandler.update() signature mismatch
### Description

While running the Visdom test suite locally, several unit tests fail because the invocation of `UpdateHandler.update()` in the tests does not match the current implementation signature.

### Current Implementation

`UpdateHandler.update()` in `py/visdom/server/handlers/web_handlers.py` requires:

```python
UpdateHandler.update(
    p,
    args,
    max_text_lines,
    max_old_content,
    max_image_history,
)
```

### Test Invocation

The affected tests call:

```python
UpdateHandler.update(p, args)
```

which results in:

```text
TypeError: UpdateHandler.update() missing 3 required positional arguments:
'max_text_lines',
'max_old_content',
'max_image_history'
```

### Steps to Reproduce

1. Clone the repository.
2. Install dependencies.
3. Run:

```bash
python -m unittest discover tests
```

4. Observe failures originating from tests that invoke `UpdateHandler.update()` with an outdated argument list.

### Expected Behavior

The test suite should execute without signature-related failures.

### Actual Behavior

Multiple tests fail with `TypeError` because the test calls do not match the current method signature.

### Additional Notes

The issue appears to have been introduced after additional parameters were added to `UpdateHandler.update()` without corresponding updates to the affected test cases.

Debajeet-1411:
Fixes #1431

## Summary

This PR fixes test suite failures caused by two issues:

1. `test_plot_history.py` was invoking `UpdateHandler.update()` using an outdated 2-argument call signature, while the current implementation requires additional history-cap arguments.
2. Tests could accidentally import a globally installed `visdom` package instead of the repository source tree, leading to version mismatches and import errors.

## Changes

* Ensure test files load the local source tree (`py/`) before importing `visdom`.
* Add a small helper wrapper in `test_plot_history.py` that supplies the required `UpdateHandler.update()` arguments.
* Remove dependency on `DEFAULT_MAX_*` imports that may not exist in older installed versions.
* Update affected assertions to use the corresponding test constants.

## Verification

Tested with both the system Python interpreter and the project virtual environment:

```bash
python -m unittest discover tests
```

Result:

```text
Ran 26 tests
OK
```

No production code was modified; changes are limited to the test suite.


Manik-Khajuria-5 (changes requested):
@Debajeet-1411 Can u Please provide Proper video for the issue as well as the solution u have introduced

Debajeet-1411:
@Manik-Khajuria-5 Thanks for reviewing.

The issue was caused by two factors:

1. `test_plot_history.py` was still calling `UpdateHandler.update(p, args)`, while the current implementation requires additional cap arguments (`max_text_lines`, `max_old_content`, and `max_image_history`).

   To address this, I added a small helper wrapper that supplies these arguments and updated the affected test call sites.


2. The tests could import a globally installed `visdom` package instead of the local repository source tree. When this happened, the test suite depended on the installed package version rather than the code under test. In particular, some versions do not expose the `DEFAULT_MAX_*` constants used by the tests, resulting in import failures and environment-dependent behavior.

   To address this, I added:

   ```python
   _REPO_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
   sys.path.insert(0, os.path.join(_REPO_ROOT, "py"))
   ```
   so the tests always run against the repository source tree.


I have attached the screenrecording showing:

* the failing behavior,
* the code changes introduced,
* and the test suite passing after the fix (`Ran 26 tests, OK`).

https://github.com/user-attachments/assets/84dd4f3a-5419-415b-a1a8-80ee975b29ac

Manik-Khajuria-5 (commented):
@Debajeet-1411 Thanks for contributioon But our main focus is on version release will review this PR after that If u could solve issue related to curr vulnerability in codebase it would be helpful

rajnisht7:
@Debajeet-1411, This PR is on hold as we are currently focusing on upcoming maintenance release.

Debajeet-1411:
Got it. Thanks for the review and the update 👍

Manik-Khajuria-5 (changes requested):
@Debajeet-1411 Thanks for contribution Do we have this issue right now ? We have already removed these files from test section

Debajeet-1411:
@Manik-Khajuria-5 Thanks for following up and for reviewing this again.

I pulled the latest upstream/dev and verified the current state of the repository. The Python test files that this PR modified (test_plot_history.py and test_json_patch.py) are no longer present in the current codebase, 

The required argument passing for UpdateHandler.update() is natively handled within UpdateHandler.update_packet(), and project validation has successfully transitioned to Cypress E2E tests and live demo scripts as established in PR #1442.
This PR is no longer relevant. I'll go ahead and close it

PR #1444, opened 2026-06-12
fix: remove unused imports and unreachable auth guard in server handlers

Jayantparashar10:
## Description

Remove unused imports and an unreachable auth guard from three server-side Python files.

1. `socket_handlers.py`: Remove import types — the module is imported but never referenced in the file body.

3. `web_handlers.py`:

-   Remove from collections import `OrderedDict` and the Mapping/Sequence try/except import block — none of these names appear in the file body the logic that used them lives in `server_utils.py`.
-   Remove the manual if `self.login_enabled` and` not self.current_user `guard inside `UpdateHandler.post` — it is unreachable because the `@check_auth` decorator already performs the same check and short-circuits before the method body runs. The guard also returned HTTP 400, inconsistent with the decorator's correct 401.

4. `server_utils.py`:

-   Narrow the defaults import to only `DEFAULT_ENV_PATH`, removing `LAYOUT_FILE, DEFAULT_BASE_URL, DEFAULT_HOSTNAME, and DEFAULT_PORT` which are imported but never used in this file.
-   Remove`warn_once` from the `shared_utils` import — it is called from `app.py` directly, not from `server_utils.py`.

## Motivation and Context
Dead imports and unreachable code add maintenance noise and make the codebase harder to navigate. The unreachable auth guard in `UpdateHandler.post` was also silently inconsistent — it returned 400 where every other auth failure returns 401.

## How Has This Been Tested?

1. Started the server with `python -m visdom.server`
2. The full example demo was then run against with `python demo.py -port 8097`
3. All visualizations rendered without errors at `http://localhost:8097`
5. No import errors, no handler crashes, no regressions in any demo component.

## Types of changes

## Checklist:


tonypzy (approved):
LGTM

Manik-Khajuria-5 (changes requested):
@Jayantparashar10 I don't think status code must be removed Can u please go through it once

Jayantparashar10:
> @Jayantparashar10 I don't think status code must be removed Can u please go through it once

The `set_status(400)` is not removed — it stays in the `@check_auth decorator` at `server_utils.py`. The decorator runs before the function body, so when `login_enabled=True` and `current_user=None`, it already returns 400 before the `post()` method body is ever entered. The removed lines are unreachable — the exact same condition is checked and handled one level up

rajnisht7 (approved):

Manik-Khajuria-5 (changes requested):
@Jayantparashar10 Still curious whether we should remove status code form post handler or not @vedansh-5 @Saksham-Sirohi What do u think

Jayantparashar10:
> @Jayantparashar10 Still curious whether we should remove status code form post handler or not @vedansh-5 @Saksham-Sirohi What do u think

`check_auth` at `server_utils.py` already does exactly:
```python3

 if handler.login_enabled and not handler.current_user:
      handler.set_status(401)
      return
```
So the duplicate check inside `UpdateHandler.post()`:
```python3
  if self.login_enabled and not self.current_user:
      self.set_status(400)
      return
```

is completely redundant — `@check_auth` runs first and blocks the request before post() body even executes. It also sets 400 instead of 401 which is actually the wrong status code for unauthorized access — another reason to remove it.

Manik-Khajuria-5 (approved):

vedansh-5 (approved):

Saksham-Sirohi (approved):

PR #1466, opened 2026-06-15
chore: Upgrades react-resize-detector and refactors pane wrapper

vedansh-5:
## Description
Upgrades `react-resize-detector` from `7.1.2` to `9.1.1` and refactors the pane wrapping in `js/main.js` to use the modern `useResizeDetector` hook pattern.

## Motivation and Context
- Bumps an outdated dependency.
- Replaces the legacy component pattern (`ReactDOM.findDOMNode` inside `<ReactResizeDetector>`) with the recommended hook pattern to prevent warnings in React Strict Mode.
- Avoids mounting issues where pane dimensions could be `undefined` or `NaN` initially by providing layout-derived defaults.

## How Has This Been Tested?
- Verified dynamic resizing by dragging pane panel corners and resizing the browser window, ensuring no layout bleeding/overlapping.
- Ran Cypress integration tests locally to ensure clean initial loads and UI navigation.

## Types of changes

- [ x ] Code refactor or cleanup (changes to existing code for improved readability or performance)

## Checklist:

- [ x ] My code follows the code style of this project.


tonypzy (commented):
Thanks for your contribution. I found that you claim to `use the modern useResizeDetector hook pattern` and `replace the legacy component pattern`, but the legacy one has not been changed yet. So I think refactoring is needed.

vedansh-5:
> Thanks for your contribution. I found that you claim to `use the modern useResizeDetector hook pattern` and `replace the legacy component pattern`, but the legacy one has not been changed yet. So I think refactoring is needed.

Thanks for pointing it out - it was lost in the rebase - (https://github.com/fossasia/visdom/pull/1466/changes/e0c488d4a92cd3dd4adf359b0ec16efdfcc5d4a4 - after this commit). 
I will add it back.

Jayantparashar10 (changes requested):
@vedansh-5 The `package.json` in this PR is downgrading other packages. Please rebase it with the dev branch to fix this.

js/main.js
```diff
@@ -770,29 +807,22 @@ const App = () => {
       let filter = getValidFilter(filterString);
       let isVisible = pane.title.match(filter);
 
-      const PANE_TITLE_BAR_HEIGHT = 14;
```
tonypzy, line 773: The previous implementation subtracted 14px (`PANE_TITLE_BAR_HEIGHT`) from the pane content height to account for the title bar. The refactored `PaneWrapper` drops this completely, deducting only 2px instead. I believe we should reconsider calculation of pixels, and please upload pane screenshots to confirm that.

tonypzy: I am just concerned that it will overflow

Manik-Khajuria-5 (changes requested):
@vedansh-5 Previuously there was key Id in component section but now it is missing PLease check it

js/main.js
vedansh-5: You are completely right, and I've restored the PANE_TITLE_BAR_HEIGHT subtraction inside the refactored PaneWrapper. Here;s why I did so-

In the old implementation even though we specified height={h2p(panelayout.h) - 14} in JSX, the ReactResizeDetector component cloned Comp under the hood and overwrote the height prop with the full measured height of the grid item.
Because of this overwrite, the old code actually rendered visualizations using the full window height, causing the bottom 14px to be clipped by CSS .content { overflow: hidden }. The repository's visual regression baseline screenshots were taken in this clipped state.
When I refactored to v9 using useResizeDetector(), the height prop was no longer overwritten and successfully took the correct height (which subtracted 14px). This visual difference compared to the old clipped baseline screenshots caused the visual regression tests to fail.
The Workaround: To make the tests pass, I had temporarily removed the 14px subtraction so that the visual output matched the old full-height baseline snapshots and forgot to re-add this.

vedansh-5:
Addressed the suggested changes.

vedansh-5:
Please review again @tonypzy @Manik-Khajuria-5 @Jayantparashar10 @rajnisht7 
The VR tests are bound to fail because of new changes.

Manik-Khajuria-5 (approved):

tonypzy (approved):
It looks good. Please resovle conflicts later

rajnisht7 (commented):
@vedansh-5 overall it looks good and works fine, just a suggestion, when we execute `npm run dev` it raises `DefinePlugin warning` as the mode is set to dev
which can be handled by modifying `webpack.common.js`
the approaches which i tested and worked removing
```
'process.env.NODE_ENV': JSON.stringify(
        process.env.NODE_ENV || 'production'
      ),
```
we can let webpack handle this

vedansh-5:
> @vedansh-5 overall it looks good and works fine, just a suggestion, when we execute `npm run dev` it raises `DefinePlugin warning` as the mode is set to dev which can be handled by modifying `webpack.common.js` the approaches which i tested and worked removing
> 
> ```
> 'process.env.NODE_ENV': JSON.stringify(
>         process.env.NODE_ENV || 'production'
>       ),
> ```
> 
> we can let webpack handle this

addressed this @rajnisht7. Thanks for your review.

vedansh-5:
Please review again @rajnisht7 @Jayantparashar10. This PR is ready from my end.

js/main.js
```diff
@@ -57,6 +57,46 @@ if (ACTIVE_ENV !== '') {
...
+  defaultHeight,
+}) => {
+  const { width, height, ref } = useResizeDetector();
+  const PANE_TITLE_BAR_HEIGHT = 14;
+
+  const finalWidth =
+    width !== undefined && width > 0 ? width - 2 : defaultWidth;
+  const finalHeight =
+    (height !== undefined && height > 0 ? height - 2 : defaultHeight) -
+    PANE_TITLE_BAR_HEIGHT;
```
rajnisht7, line 78: @vedansh-5 the PR looks good and its good to go just wanted to ask why are we reducing height and width by 2px

Jayantparashar10 (approved):

js/main.js
vedansh-5: We now dynamically measure the wrapper's dimensions directly from the DOM, because ResizeObserver gives the outer border-box dimensions of the container, the measured width and height include the 1px border on all sides.Subtracting these pixels ensures that the internal visualization elements fit exactly within the borders of the window without triggering unwanted scrollbars or layout clipping

rajnisht7 (approved):

Saksham-Sirohi (approved):

PR #1468, opened 2026-06-15
fix: replace assert with errno check in ensure_dir_exists to surface permission errors

Jayantparashar10:
## Description
`ensure_dir_exists` in `shared_utils.py` used an assert statement to distinguish "directory already exists" from other filesystem errors:

```python
  except OSError as e1:
      assert e1.errno == 17  # errno.EEXIST
```
This has two problems:
  
  1. Any OSError that is not EEXIST (e.g. permission denied, read-only filesystem) raises AssertionError instead of the original error — the real cause is hidden from the user.
  2. Under python -O (optimised mode), Python strips all assert statements — every OSError is silently swallowed, including permission errors.

The fix replaces the entire try/except block with exist_ok=True:
```python
  os.makedirs(os.path.abspath(path), exist_ok=True)
```
  Python's own implementation of `exist_ok=True` handles all cases correctly:
  - Directory does not exist → creates it
  - Directory already exists → suppresses the error
  - A file exists at that path → still raises `FileExistsError`
  - Permission denied → still raises `PermissionError`

## Motivation and Context
If Visdom cannot create its working directory due to a permission issue, the current code either hides the real error behind `AssertionError` or swallows it entirely under python -O. The fix ensures real filesystem errors are always
surfaced, removes a number (17), and simplifies the function to a single line using the standard Python 3 idiom for `os.makedirs`.

## How Has This Been Tested?
Verified manually by running the Visdom server on the patched branch — the server starts and creates `~/.visdom/` as expected. FileExistsError is a standard Python 3 idiom for `os.makedirs` error handling and does not change behaviour for the normal case (directory already exists).

## Types of changes

## Checklist:

## Summary by Sourcery

Improve error handling in directory creation utility to avoid masking underlying filesystem errors.

Bug Fixes:
- Ensure ensure_dir_exists re-raises non-EEXIST OSError instances so permission and other filesystem errors are surfaced instead of being converted to AssertionError.

Enhancements:
- Use errno.EEXIST when handling the "directory already exists" case in ensure_dir_exists for clearer and more robust error checking.


Manik-Khajuria-5 (approved):

tonypzy (approved):

rajnisht7 (approved):

Jayantparashar10:
@Saksham-Sirohi This PR is ready to merge.

Saksham-Sirohi (approved):

PR #1483, opened 2026-06-18
fix: move None checks before `np.asarray()` in dual_axis_lines so asserts can actually fire

Jayantparashar10:
## Description

Move the `None` checks for `X`, `Y1`, and `Y2` in `dual_axis_lines` to before the `np.asarray()` calls, so the assertions can actually fire when `None` is passed.

## Motivation and Context
`np.asarray(None)` does not return `None` — it returns `array(None, dtype=object)`, a valid numpy array. The original code called `np.asarray()` first and then asserted `X is not None`, meaning the assert could never fire regardless of what the caller passed. Passing `None` would silently proceed past the guard and fail later with a confusing shape error or traceback instead of the intended "X Cannot be None" message.

## How Has This Been Tested?

Tested with script:

```python
  import numpy as np
  from visdom import Visdom

  viz = Visdom(port=8097)
  X = np.array([1.0, 2.0, 3.0])
  Y = np.array([4.0, 5.0, 6.0])

  try:
      viz.dual_axis_lines(X=None, Y1=Y, Y2=Y)
  except AssertionError as e:
      assert str(e) == "X Cannot be None"

  try:
      viz.dual_axis_lines(X=X, Y1=None, Y2=Y)
  except AssertionError as e:
      assert str(e) == "Y1 Cannot be None"

  try:
      viz.dual_axis_lines(X=X, Y1=Y, Y2=None)
  except AssertionError as e:
      assert str(e) == "Y2 Cannot be None"

  try:
      viz.dual_axis_lines(X=X, Y1=np.array([1.0, 2.0]), Y2=Y)
  except AssertionError as e:
      assert "not in proper shape" in str(e)

  viz.dual_axis_lines(X=X, Y1=np.array([1.0, 4.0, 9.0]), Y2=np.array([0.1, 0.2, 0.3]))

```

*N/A*

## Types of changes

## Checklist:


Manik-Khajuria-5 (approved):
@Jayantparashar10 LGTM!

rajnisht7 (approved):

Manik-Khajuria-5 (approved):

tonypzy (approved):

Saksham-Sirohi (approved):

PR #1485, opened 2026-06-19
fix: isolate EventSystem subscriber errors to prevent event delivery failures

Jayantparashar10:
## Description

`EventSystem.publish` (`EventSystem.js`): Wrap each subscriber callback in try/catch so a crash in one subscriber does not stop the remaining subscribers from receiving the event.

## Motivation and Context
`EventSystem.publish` iterates subscribers with a bare `forEach`. If any subscriber throws, the loop stops and all subsequent subscribers never receive the event. In Visdom, `global.event` is shared by ImagePane, TextPane, and EmbeddingsPane simultaneously — a null ref or bad state in one pane silently kills keyboard and click events for every other pane on the dashboard. 

## How Has This Been Tested?
Start the Visdom server. Create a few panes — a Text pane, an Image pane, and an Embeddings pane — so that all three are subscribed to keyboard and click events simultaneously.

Open the browser DevTools console. Click on one pane to focus it and press a key. Before this fix, if any subscriber threw an error during that keypress, the remaining panes silently stopped receiving the event — the user would see some panes not responding to keyboard or click input with no visible error in the UI. After this fix, the error appears in the console prefixed with EventSystem: subscriber error and the remaining panes continue receiving events normally.

N/A

## Types of changes

## Checklist:

## Summary by Sourcery

Improve robustness of the frontend event system and fix duplicate event subscriptions in TextPane.

Bug Fixes:
- Prevent one failing event subscriber from blocking delivery to other subscribers by isolating callback errors and logging them.
- Ensure TextPane only registers a single global event subscription per focus change instead of re-subscribing on every render.


Manik-Khajuria-5 (approved):
@Jayantparashar10 LGTM!

Manik-Khajuria-5 (changes requested):
@Jayantparashar10 

Image That is not how react works. React 18 runs the effect cleanup before re-running the effect during the development double-mount cycle. If the unsubscribe  is working correctly, shouldn't that prevent duplicate active subscriptions?

tonypzy (changes requested):
I agree with @Manik-Khajuria-5 . There was never a duplicate subscription. This just avoids re-subscription churn on every render

Jayantparashar10:
@Manik-Khajuria-5 @tonypzy Thanks for the review and clarification. I've reverted the TextPane dependency array change and updated the PR to contain only the EventSystem subscriber error isolation fix.

Manik-Khajuria-5 (approved):
@Jayantparashar10 LGTM! and thanks for applying my suggestion

vedansh-5 (approved):

rajnisht7 (approved):

tonypzy (approved):

PR #1487, opened 2026-06-20
Feat: Server over https

### LINKED ISSUE #148: Serve over HTTPS
Is there any way to tell Visdom to make requests over HTTPS as opposed to HTTP?

When setting up a reverse proxy with HTTPS, pointing to `localhost:8097`, I get these messages (Chrome).

```
Mixed Content: The page at 'https://my.web.address/' was loaded over HTTPS, but requested an insecure script 'http://my.web.address/static/js/main.js?v=4b72317f763632006def9236be6408b4'. This content should also be served over HTTPS.
```

Which is fine for development purposes, however I also receive this error message that I'm not sure how to resolve.

```
WebSocket connection to 'ws://my.web.address/socket' failed: Error during WebSocket handshake: Unexpected response code: 301
```

And this (Edge)

```
SCRIPT5022: SecurityError
main.js (1,13767)
```

For completeness, here's the Caddy configuration on how I'm reverse proxying Visdom.

```ini
my.web.address {
    proxy / localhost:8097 {
        transparent
        websocket
    }

    log stdout
    errors stderr
}

```

rajnisht7:
## Description
This PR allows user to serve visdom over https

## Motivation and Context
currently visdom only operate over http and have no support for https, this PR enables this

Fixes: #148 

## How Has This Been Tested?

**the below commands are for dev only not for production**

generate certificate:
`openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -nodes`

then start server using
`python -m visdom.server -ssl_certfile cert.pem -ssl_keyfile key.pem`

test using:
```
import visdom
import numpy as np

viz = visdom.Visdom(server='https://localhost',env='test-https',ssl_verify=False)

viz.bar(
    X=np.array([1, 2, 3]),
    win='my_bar_window',
    opts=dict(title='bar9')
)


viz.save(['abcpqr'])
```

https://github.com/user-attachments/assets/11305e05-8b01-48b0-ba6e-d580799c8406

**to test with ssl_verify=True**

**mkcert** can be used

execute the following commands:

`sudo pacman -S mkcert nss`

`mkcert -install`

`mkcert -CAROOT` expected output `The certificate is at "./localhost+2.pem" and the key at "./localhost+2-key.pem"`

you can rename them using
```
mv localhost+2.pem cert.pem
mv localhost+2-key.pem key.pem
```

start the server
`python -m visdom.server -ssl_certfile cert.pem -ssl_keyfile key.pem`

and the graphs can be executed using

```
import visdom
import numpy as np

viz = visdom.Visdom(server="https://localhost",env='https-test1',ssl_verify="/path/to/rootCA.pem")

viz.bar(
    X=np.array([1, 2, 3]),
    win='my_bar_window',
    opts=dict(title='bar9')
)

```

if the user dont want to pass the path for ssl_verify

the user can just do

`export REQUESTS_CA_BUNDLE=/home/acer/.local/share/mkcert/rootCA.pem`

that is export the `rootCA.pem`

and then user can go directly with

`viz = visdom.Visdom(server="https://localhost",env='https-test1',ssl_verify=True)`

https://github.com/user-attachments/assets/67894856-21a7-4c75-8836-6840238b958b

https://github.com/user-attachments/assets/f3dffcee-d05a-4682-953d-f8af27e02da6

**tested it using remote-server setup too, worked fine**

## Types of changes

## Checklist:


py/visdom/__init__.py
```diff
@@ -535,10 +536,12 @@ def __init__(
         proxies=None,
         offline=False,
         use_polling=False,
+        ssl_verify=True,
     ):
         parsed_url = urlparse(server)
         if not parsed_url.scheme:
-            parsed_url = urlparse("http://{}".format(server))
+            scheme = "https" if not ssl_verify else "http"
```
tonypzy: Currently, if we want to use HTTPS, we cannot use SSL verification

tonypzy (commented):
please resolve

tonypzy:
I don't think `ssl_verify` should affect the transmission protocol.

rajnisht7:
@tonypzy if any user want to use http the user can go via the current approach `viz = visdom.Visdom(env='http-test1')` , more parameters are added only if any user want to use https

tonypzy:
> @tonypzy if any user want to use http the user can go via the current approach `viz = visdom.Visdom(env='http-test1')` , more parameters are added only if any user want to use https

I know what you mean. But this doesn't solve the coupling itself. Let's see what others think of @Manik-Khajuria-5 @Saksham-Sirohi @jay-karimi @vedansh-5

tonypzy (commented):
Please resolve

Manik-Khajuria-5 (changes requested):
@rajnisht7  I agree with @tonypzy  also ssl verify should only control certificate validation, while the choice between HTTP and HTTPS should remain independent. Could we decouple these two behaviors?

rajnisht7:
@Manik-Khajuria-5 @tonypzy coupling is used only for user convenience, that is if user starts the server at `https://localhost` then as the server already started at `https`  and user chose it explicitly there is no point of repeating it in code by making user write `server="https://localhost" ` the user only need to verify for the ssl and its good to go, but if user want to use any other domain then the user explicitly need to enter the server along with ssl, coupling is only used here for `localhost` and to make a bit easier for users

Manik-Khajuria-5 (changes requested):
@rajnisht7 The main concern is that ssl verify is currently being used to infer the transport scheme (http vs https). Certificate verification and protocol selection are independent concerns, and the current logic results in ssl verify=True defaulting to HTTP while ssl verify=False enables HTTPS. This makes the behavior somewhat counterintuitive and prevents users from expressing HTTPS with certificate verification through the auto-scheme path. I'd suggest keeping transport selection independent and using ssl verify only for certificate validation. @Saksham-Sirohi @vedansh-5 @tonypzy What do u think about it

Manik-Khajuria-5 (changes requested):
@rajnisht7 Please resolve the conflicts

Manik-Khajuria-5 (approved):

Manik-Khajuria-5 (changes requested):
@rajnisht7 Please resolve the conflict

Manik-Khajuria-5 (approved):
@rajnisht7 Nice work! One thing I noticed is that the fixed timeout = 20 could be a bit restrictive for large uploads on slower connections.

Jayantparashar10 (approved):

vedansh-5 (approved):

Saksham-Sirohi (approved):

PR #1489, opened 2026-06-20
fix: use np.nanmin/nanmax in histogram to handle NaN input, completing the TODO

### LINKED ISSUE #1499: numpy `.max()`/`.min()` calls silently propagate NaN, breaking plots and image scaling
**Bug Description**
  `py/visdom/__init__.py` uses `.max()` and `.min()` in 9 places across scatter, histogram, heatmap, and image functions. These propagate NaN silently instead of  ignoring it, causing broken plots, wrong axis ranges, and incorrect image normalization.

  **Reproduction Steps**

  1. Start visdom server
  2. Send data containing NaN values, e.g.:
     ```python
     import numpy as np
     viz.histogram(X=np.array([1, 2, float('nan'), 4, 5]))
  3. Histogram axes show NaN ranges instead of valid data range

  ## Expected behavior
  NaN values should be ignored when computing min/max. Plots and images should render correctly using the valid data range.

  ## Additional context

  There is an existing TODO at line 118 of __init__.py that flags this:

   TODO: In appropriate places, we need to change many numpy calls to use
         nan-aware ones, e.g., `X.max` => `np.nanmax(X)`.

  Fix is to replace .max()/.min() with np.nanmax()/np.nanmin() at all 9 locations.

Jayantparashar10:
## Description

`viz.histogram()` used `X.min()` and `X.max()` to compute the histogram range. Both methods propagate NaN by default, so if even one value in X is NaN, both return nan. `np.histogram(..., range=(nan, nan))` then raises:
```
ValueError: supplied range of [nan, nan] is not finite
```
Changed to `np.nanmin(X)` / `np.nanmax(X) `so NaN values are skipped and the range is computed from the finite values only. NaN entries are naturally excluded from binning by `np.histogram` itself.

Fixes #1499 

## Motivation and Context

NaN values appear regularly in ML training metrics — exploding loss, undefined gradients, sensor dropouts. A user logging any of these to a histogram hits a hard crash. This fix completes the TODO in `py/visdom/__init__.py`:
```
  # TODO: In appropriate places, we need to change many numpy calls to use
  #       nan-aware ones, e.g., `X.max` => `np.nanmax(X)`.
```
`histogram()` was the one function that was not updated.

## How Has This Been Tested?
Tested in isolation by replicating the histogram logic directly:
```python3
  import numpy as np
  X = np.array([1.0, 2.0, float('nan'), 3.0])
  minx, maxx = np.nanmin(X), np.nanmax(X)  # 1.0, 3.0
  bins = np.histogram(X, bins=4, range=(minx, maxx))[0]
  # [1, 0, 1, 1] — 3 finite values binned correctly, NaN ignored
```
Before the fix: ValueError: supplied range of [nan, nan] is not finite
After the fix: histogram renders correctly with NaN entries silently excluded.

Also verified that clean inputs (no NaN) are unaffected — `np.nanmin`/`np.nanmax` are identical to min/max when no NaN is present.

N/A

## Types of changes

## Checklist:


tonypzy (approved):
LGTM. We can move the NaN check to a separate PR, as np.nanmax() returns NaN when the input contains only NaNs

Manik-Khajuria-5 (approved):
@Jayantparashar10 LGTM!

rajnisht7 (approved):

vedansh-5 (approved):

PR #1509, opened 2026-06-21
fix: graph() mutable default opts and bare except swallowing KeyboardInterrupt

Jayantparashar10:
## Description
Two fixes in `graph()` in `py/visdom/__init__.py`:

1. Changed `opts=dict()` in the function signature to `opts=None`, with `opts = {} if opts is None else opts` at the top of the function body.
2. Changed except: on the `import networkx` block to except `ImportError:`.

## Motivation and Context
`opts=dict()` as a default argument is evaluated once when Python loads the module, not once per call. The function then writes into it:
  ```python3
opts["directed"] = opts.get("directed", False)
opts["showVertexLabels"] = opts.get("showVertexLabels", False)
```
So all calls that omit opts share the same dict object. Currently harmless since the written values match the defaults, but any future change to those defaults would silently corrupt state across calls.

`except:` catches everything — including `KeyboardInterrupt` and `SystemExit`. If the user presses Ctrl+C while import networkx is running, they get RuntimeError: networkx must be installed to plot Graph figures instead of the process actually stopping.

## How Has This Been Tested?
Verified the mutable default fix by calling `viz.graph(edges)` twice without passing opts and confirming each call gets a fresh independent dict.

Verified the `except ImportError`: fix by confirming `KeyboardInterrupt` now propagates correctly instead of being swallowed.

Clean inputs are unaffected — passing opts explicitly behaves identically to before.

N/A

## Types of changes

## Checklist:


tonypzy (approved):
LGTM

Manik-Khajuria-5 (approved):

rajnisht7 (approved):

vedansh-5 (approved):

PR #1513, opened 2026-06-22
Synchronize image history slider selections across clients

### LINKED ISSUE #1512: Image history slider selection is not synchronized across connected clients
# Image history slider selection is not synchronized across connected clients

## Summary

While testing `image_history` panes, I noticed that changing the history slider only updates the local browser state and is not propagated to other connected clients.

There is already a TODO in `js/panes/ImagePane.js` that appears to reference this missing functionality:

```javascript
// TODO add history update events here! need to send these to the client
// with sendPaneMessage
```

## Current Behavior

The slider handler currently only updates local React state:

```javascript
const updateSlider = (evt) => {
  setActualSelected(parseInt(evt.target.value));
};
```

No pane update event is emitted through `sendPaneMessage`.

## Reproduction

1. Start a Visdom server.
2. Create an `image_history` window (`store_history=True`).
3. Open the same Visdom environment in two browser tabs.
4. Move the history slider in Tab A.
5. Observe Tab B.

## Observed Result

The selected history frame changes only in the tab where the slider was moved.

Other connected clients remain on their previous selection.

Image

## Expected Result

History selection changes should be propagated through the pane messaging system so all connected clients remain synchronized.

## Additional Context

The existing TODO in `js/panes/ImagePane.js` suggests that history update events were intended to be sent through `sendPaneMessage`, but this has not yet been implemented.

Debajeet-1411:
Fixes #1512
## Summary

Synchronize `image_history` slider selections across connected clients.

Previously, moving the history slider only updated local React state in the active browser tab. Other connected clients viewing the same environment did not receive the updated selection, causing image history panes to become out of sync.

This change implements the existing TODO in `ImagePane.js` by propagating slider selection changes through the pane messaging system and broadcasting updates to connected clients.

### Frontend

**js/panes/ImagePane.js**

- Updated `updateSlider()` to emit an `image_update_selected` event via `sendPaneMessage`.
- Preserved existing local state updates for immediate UI responsiveness.

### Backend

**py/visdom/server/handlers/socket_handlers.py**

- Added handling for the new `image_update_selected` event.
- Updates the server-side selected image index.
- Regenerates `contentID` to trigger client updates.
- Uses `jsonpatch` to generate a minimal patch rather than rebroadcasting the entire image history payload.
- Broadcasts the update to connected clients.
- Continues forwarding the event to connected Python sources.

### Before

1. Start Visdom.
2. Create an `image_history` window (`store_history=True`).
3. Open the same environment in two browser tabs.
4. Move the history slider in Tab A.

Result:
- Tab A updates.
- Tab B remains on the previous selection.

### After

1. Start Visdom.
2. Create an `image_history` window (`store_history=True`).
3. Open the same environment in two browser tabs.
4. Move the history slider in Tab A.

Result:
- Tab A updates.
- Tab B updates to the same selected history frame.


Manik-Khajuria-5 (changes requested):
@Debajeet-1411 Please provide a video of impact of your PR and also please resolve conflicts

Debajeet-1411:
@Manik-Khajuria-5  Thanks for the review.

The merge conflicts have now been resolved.

While working on this fix, I noticed that the frontend changes in `js/panes/ImagePane.js` require running `npm run build` to regenerate the bundled frontend assets. This updates the generated files:

* `py/visdom/static/js/main.js`
* `py/visdom/static/js/main.js.map`

These generated files contain thousands of bundled changes, which significantly increase the PR diff size and were the source of the earlier conflicts.

For now, I removed those generated build artifacts from the PR to keep the review focused on the actual source changes:

* `js/panes/ImagePane.js`
* `py/visdom/server/handlers/socket_handlers.py`

Could you please advise on the preferred approach here?

* Should the regenerated `main.js` and `main.js.map` files be included in this PR?
* Or would you prefer to rebuild the frontend assets during merge/release from your side?

I've also attached a short video demonstrating the fix. 
Please let me know if any changes are needed.

https://github.com/user-attachments/assets/e462b464-7f90-4a5a-9820-43fc7f7d3f67

Jayantparashar10:
@Debajeet-1411 Thank you for your contribution! We're closing this PR since we already have a PR #1266 . Once again, thanks for your contribution!

PR #1519, opened 2026-06-25
fix: release WebGL context on unmount and clamp pane dimensions to prevent GL_INVALID_VALUE on resize

Jayantparashar10:
## Description
Release the WebGL rendering context when the Scene component unmounts, and clamp pane dimensions to prevent a zero/negative viewport during resize.

Two changes in `EmbeddingsPane.js`:
  - `componentWillUnmoun`t: call `renderer.forceContextLoss()` then `renderer.dispose()` on unmount to release the GPU context
  - `setUpScene`: clamp width and height with `Math.max(1, this.props.width)` before passing them to the renderer

## Motivation and Context
The Scene component re-mounts on every pane resize because its key prop includes the pane dimensions. Each mount creates a new THREE.WebGLRenderer — and therefore a new GPU context — but the old context was never released. 

Browsers cap the number of active WebGL contexts; after enough resizes the browser logs "Too many active WebGL contexts. Oldest context will be lost" and begins forcibly destroying earlier contexts, causing rendering to break.
  
`forceContextLoss()` calls `WEBGL_lose_context.loseContext()` synchronously, which returns the GPU context slot to the browser before the JS object is GC'd. `dispose()` then frees the `THREE.js`-internal state. Both are needed dispose() alone
does not release the GPU context.

The `GL_INVALID_VALUE: glViewport: negative width/height` error was triggered when the pane passed 0 or negative pixel dimensions during a fast drag resize. Clamping to `Math.max(1, ...)` prevents the renderer and camera from ever receiving
invalid dimensions.

## How Has This Been Tested?

- Opened the embeddings pane in the browser with `example/mnist-embeddings.py` (using openTSNE backend)
- After applying the fix, resized the pane repeatedly via drag confirmed no `GL_INVALID_VALUE: glViewport` warnings in the console
- After applying the fix, confirmed no "Too many active WebGL contexts" warnings after repeated resize cycles

Before Fix:

https://github.com/user-attachments/assets/872dfd5e-29bc-40ee-abc2-bcc411ba7dcd

After Fix:

https://github.com/user-attachments/assets/87b0523b-779d-445f-bbf6-50bae0c47e21

## Types of changes

## Checklist:

## Summary by Sourcery

Ensure the embeddings WebGL scene uses valid dimensions and releases its rendering context when unmounted.

Bug Fixes:
- Clamp the embeddings pane width and height to a minimum of 1 pixel to avoid invalid WebGL viewport dimensions during resize.
- Release the WebGL renderer context and dispose of renderer resources when the scene component unmounts to prevent leaked GPU contexts on repeated resizes.


Manik-Khajuria-5 (approved):
@Jayantparashar10 Nice fix

rajnisht7 (approved):

tonypzy (approved):

vedansh-5 (approved):
LGTM

PR #1548, opened 2026-07-01
Fix: infinite redirect loop on invalid env

rajnisht7:
## Description
This PR adds a back to dashboard icon in error.html, currently if we navigate to any env, the env is saved in local storage and if it is an invalid env the error page will  be rendered, but that env will get saved in local storage too, so if any user tries to go back it will render the invalid env as it is saved in local storage and it will lead to error page and this loop will never break

## Motivation and Context
improve user experience by breaking the never ending invalid env and error page loop

## How Has This Been Tested?
tested and verified the behavior before and after fix

https://github.com/user-attachments/assets/23bae321-1a41-4176-90ac-dfa4ecfeb40c

## Types of changes

## Checklist:

## Summary by Sourcery

Add a dashboard back link to the error page to break the redirect loop caused by invalid environments stored in local storage.

Bug Fixes:
- Prevent infinite redirect loop by clearing stored environment IDs when navigating back from the error page.

Enhancements:
- Improve error page UX with a visible back-to-dashboard link styled consistently with existing UI.


Manik-Khajuria-5 (changes requested):
@rajnisht7 Please add base url for this

Manik-Khajuria-5 (changes requested):
@rajnisht7 I went through the code instead of cleaning the whole envId's What we can do is drop invalid envIds

rajnisht7:
> @rajnisht7 I went through the code instead of cleaning the whole envId's What we can do is drop invalid envIds

@Manik-Khajuria-5 localStorage.getItem('envIDs') stores the last selected env, hence the current approach is whenever error.html is rendered at any env it will remove that env, due to which when user will press back button it will take back to main env
can be checked using
localStorage.getItem('envIDs')

tonypzy (approved):

vedansh-5 (approved):

Manik-Khajuria-5 (approved):

Jayantparashar10 (approved):

PR #1541, opened 2026-07-01
feat: Integrate Playwright E2E testing infrastructure

vedansh-5:
## Description
This Pull Request introduces the base Playwright E2E testing infrastructure to Visdom. As planned, Cypress and its dependencies remain fully intact while both test suites coexist in this transitional phase.

## Motivation and Context
Cypress has been prone to flakiness and slower execution speeds. This initial PR establishes the framework and layout for Playwright E2E and visual testing, allowing subsequent PRs to migrate the remaining suites.

## How Has This Been Tested?
Tested locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactor or cleanup (changes to existing code for improved readability or performance)

## Checklist:

- [ ] I adapted the version number under `py/visdom/VERSION` according to [Semantic Versioning](https://semver.org/)
- [ x ] My code follows the code style of this project.
- [ x ] My change requires a change to the documentation.
- [ x ] I have updated the documentation accordingly.

PR #1551, opened 2026-07-02
feat: Migrate misc and text E2E test suites to Playwright

vedansh-5:
## Description
This PR migrates the E2E tests for the `misc` (special graph plotting) and `text` (text pane basics, updates, downloads, and callbacks) suites from Cypress to Playwright.
As part of this migration, I also cleaned up and resolved a few E2E testing issues in `playwright/support/helpers.js`:
- **Python path resolver**: Added `getPythonExecutable()` to automatically search for local `.venv` or `venv` folders. This fixes `spawn python ENOENT` errors if the virtual environment is not explicitly activated in the running terminal.
- **Directory expansion hang**: Fixed a loop in `expandAllEnvGroups()` where index-based iteration caused Playwright to hang. Since expanding directories dynamically changes switcher classes, we now safely click the first closed directory in a simple loop.
- **Handling text input callbacks**: Visdom's text pane uses global keyup events and updates the component via WebSockets on every character, causing it to lose focus. I updated the callback test to focus the pane header and type character-by-character, keeping focus on the wrapper element.

Successor of #https://github.com/fossasia/visdom/pull/1541

## Motivation and Context
Continues the transition from Cypress to Playwright to make E2E runs faster, more stable, and easier to debug.

## How Has This Been Tested?
Tested locally in the workspace

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x ] My code follows the code style of this project.

### DIFF
--- a/playwright/tests/misc.spec.js
+++ b/playwright/tests/misc.spec.js
@@ -1,0 +1,26 @@
+/**
+ * Copyright 2017-present, The Visdom Authors
+ * All rights reserved.
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+const { test, expect } = require('@playwright/test');
+const { runDemo } = require('../support/helpers');
+
+test.describe('Misc Tests', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/');
+  });
+
+  test('plot_special_graph', async ({ page }) => {
+    await runDemo(page, 'plot_special_graph');
+
+    await expect(page.locator('svg line')).toHaveCount(6);
+    await expect(page.locator('svg path')).toHaveCount(6);
+    await expect(page.locator('svg text')).toHaveCount(12);
+    await expect(page.locator('svg g')).toHaveCount(6);
+  });
+});
--- a/playwright/tests/text.spec.js
+++ b/playwright/tests/text.spec.js
@@ -1,0 +1,76 @@
+/**
+ * Copyright 2017-present, The Visdom Authors
+ * All rights reserved.
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+const { test, expect } = require('@playwright/test');
+const { runDemo } = require('../support/helpers');
+
+test.describe('Text Pane', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/');
+  });
+
+  test('text_basic', async ({ page }) => {
+    await runDemo(page, 'text_basic');
+    const pane = page.locator('.layout .react-grid-item').first();
+    await expect(pane).toContainText('Hello World!');
+  });
+
+  test('text_update', async ({ page }) => {
+    await runDemo(page, 'text_update');
+    const pane = page.locator('.layout .react-grid-item').first();
+    await expect(pane).toContainText('Hello World! More text should be here');
+    await expect(pane).toContainText('And here it is');
+  });
+
+  test('check download button', async ({ page }) => {
+    await runDemo(page, 'text_update');
+    const pane = page.locator('.layout .react-grid-item').first();
+    const saveBtn = pane.locator('button[title="save"]').first();
+
+    const [download] = await Promise.all([
+      page.waitForEvent('download'),
+      saveBtn.click(),
+    ]);
+
+    expect(download.suggestedFilename()).toBe('visdom_text.txt');
+  });
+
+  test('text_callbacks', async ({ page }) => {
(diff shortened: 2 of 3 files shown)


playwright/tests/text.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    await bar.click();
+
+    const layoutWrapper = page.locator('.no-focus');
+
+    const text = 'checking callback :(';
+    for (let char of text) {
+      await layoutWrapper.focus();
+      await page.keyboard.type(char);
+      await page.waitForTimeout(50);
+    }
```
Debajeet-1411, line 60: I noticed this uses `page.waitForTimeout(50)` after every key press.

From what I understand, the Playwright docs generally recommend avoiding fixed timeouts since they can make tests slower and occasionally flaky.

Would it be possible to wait for the expected UI update (or another observable state change) here instead, or is the timeout required because of how the callback demo behaves?

playwright/tests/basic.spec.js
```diff
@@ -0,0 +1,81 @@
...
+    const mainSelection = page
+      .locator('.rc-tree-select')
+      .getByText('main')
+      .first();
+    await mainSelection.hover();
+    await page.waitForTimeout(100);
+
+    const removeBtn = page
+      .locator('.rc-tree-select .rc-tree-select-selection-item-remove')
+      .first();
```
Debajeet-1411, line 68: Would `await expect(removeBtn).toBeVisible()` work here instead of `page.waitForTimeout(100)`?

I was wondering if relying on Playwright's auto-waiting could make this a bit more deterministic.

playwright/tests/misc.spec.js
```diff
@@ -0,0 +1,26 @@
...
+
+test.describe('Misc Tests', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/');
+  });
+
+  test('plot_special_graph', async ({ page }) => {
+    await runDemo(page, 'plot_special_graph');
+
+    await expect(page.locator('svg line')).toHaveCount(6);
```
Debajeet-1411, line 21: These selectors appear to be page-wide (`svg line`, `svg path`, etc.).

Was that intentional, or would scoping them to the graph pane make the assertions a little more resilient if additional SVG elements are added elsewhere in the UI in the future?

playwright/tests/text.spec.js
vedansh-5: Playwright's keyboard simulator types at near-instant speed. Without the 50ms throttle, keyup events are sent faster than the React system can process the pane focus updates, resulting in swallowed characters. The 50ms delay is an intended throttle to let the React render cycle catch up.

playwright/tests/basic.spec.js
vedansh-5: Playwright checks visibility on the DOM level. During a CSS transition, the element is technically visible in the DOM but is still animating. If Playwright attempts to click the element while the animation is in progress, it can click the wrong coordinate or miss, causing flaky failures in CI. The 100ms wait ensures the CSS transition has fully completed and the element is static.

playwright/tests/misc.spec.js
vedansh-5: This test runs in an isolated environment block where only one graph is rendered so it has zero chances of colliding with other panes.

Debajeet-1411: @vedansh-5 Thanks for the explanation! That makes sense. I appreciate the context behind these implementation choices.

rajnisht7 (approved):

Jayantparashar10 (approved):
Just one thing that `basic.spec.js` is unrelated to this PR. Please look into this.

Manik-Khajuria-5 (changes requested):
@vedansh-5 I went through the code most of things are implemneted properly but in ci  server is started in a separate CI step from the tests I think they must be in single step as there are certain times when server hanging situation happens or what if server did not survive at all or there might be another approach please verify it once as I am also not too sure about it  !

vedansh-5:
> @vedansh-5 I went through the code most of things are implemneted properly but in ci server is started in a separate CI step from the tests I think they must be in single step as there are certain times when server hanging situation happens or what if server did not survive at all or there might be another approach please verify it once as I am also not too sure about it !

Thanks for pointing out - I have addressed these changes in the predecessor PR and will rebase them here once the base PR is merged. Refer https://github.com/fossasia/visdom/pull/1541

Manik-Khajuria-5 (approved):
@vedansh-5 Okk Once previous PR is merged please rebase this PR

Debajeet-1411 (approved):
Looks good to me

tonypzy (approved):
nice

Saksham-Sirohi (approved):

PR #1556, opened 2026-07-03
Fix startup with env_path=None

tonypzy:
## Description

Added a `None` check for `self.env_path` because `os.path.join` does not accept `None` as an input. Without this check, it raises an error that causes `Application(env_path=None)` to fail during initialization, ultimately preventing the server from starting.

## Motivation and Context

In the `load_layouts`, `load_state`, and `save_layouts` methods, there are already similar `None` checks for `env_path` to use memory instead of disk.

**Use Case**

- For unit test initialization, we don't want to specify a directory path, so we use app = Application(env_path=None)
- Avoid writing to disk due to sensitive data
- Another Python tool temporarily embeds Visdom.

## How Has This Been Tested?

Execute
```
import os
import sys
import unittest


sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "py"))


class ApplicationEnvPathNoneTests(unittest.TestCase):
    def test_application_starts_with_env_path_none(self):
        from visdom.server.app import Application

        app = Application(env_path=None)

        self.assertEqual(app.state, {"main": {"jsons": {}, "reload": {}}})
        self.assertEqual(app.layouts, "")
        self.assertEqual(app.user_settings["user_css"], "")


if __name__ == "__main__":
    unittest.main(verbosity=2)

```

Before fix:
image

After fix:
image

## Types of changes

## Checklist:


tonypzy:
@Manik-Khajuria-5 @Jayantparashar10 @rajnisht7 @vedansh-5 please review

vedansh-5 (approved):

Manik-Khajuria-5 (approved):

rajnisht7 (approved):

Jayantparashar10 (approved):

Saksham-Sirohi (approved):

PR #1560, opened 2026-07-04
fix: crashes caused by invalid saved layout JSON and show toasts for recovery

tonypzy:
## Description

This PR adds defensive JSON parsing (`safeJsonParse`) to handle malformed JSON and prevent frontend runtime exceptions. 

## Motivation and Context

The following malformed JSON cases are now safely handled by `safeJsonParse`:

```
localStorage.setItem('envIDs', '["main", "demo"'); # Truncated JSON
localStorage.setItem('layout_main_pane_1', 'undefined'); # Non-JSON text
localStorage.setItem('layout_main_pane_1', '{x: 0, y: 0}'); # Invalid object syntax
localStorage.setItem('envIDs', ''); # Empty persisted value
```

## How Has This Been Tested?

Check code manually

N/A

## Types of changes

## Checklist:


tonypzy:
@vedansh-5 @rajnisht7 @Manik-Khajuria-5 @Jayantparashar10 please review

Debajeet-1411:
@tonypzy Nice improvement. Consider adding a Cypress test for malformed localStorage JSON to verify the recovery path and guard against future regressions.

Debajeet-1411 (commented):
One potential edge case came to mind—I've left an inline comment with the details.

js/main.js
```diff
@@ -55,7 +66,10 @@ if (ACTIVE_ENV !== '') {
     use_envs = [ACTIVE_ENV];
   }
 } else {
-  use_envs = JSON.parse(localStorage.getItem('envIDs')) || ['main'];
+  let storedEnvIDs = safeJsonParse(localStorage.getItem('envIDs'), null, () => {
```
Debajeet-1411, line 73: If envIDs contains valid JSON but not an array (e.g. "{}"), safeJsonParse succeeds, so the invalid value isn't removed from localStorage. The code falls back to ['main'], but the stale entry remains and will trigger the same fallback on every load. It might be worth clearing the stored value whenever the parsed value fails validation.

Jayantparashar10 (approved):

js/main.js
tonypzy: Thanks for the careful review. This PR is scoped to detect malformed JSON values. We can create a new PR to handle the case you mentioned, where the value is valid but semantically incorrect.

tonypzy:
> @tonypzy Nice improvement. Consider adding a Cypress test for malformed localStorage JSON to verify the recovery path and guard against future regressions.

cytest is not meant for handling such trivial cases. it is primarily designed to verify higher-level functionality, such as plotting features and screenshot pixel comparisons.

tonypzy:
> @tonypzy Nice improvement. Consider adding a Cypress test for malformed localStorage JSON to verify the recovery path and guard against future regressions.

Or we can add to pytest #1543

Manik-Khajuria-5 (approved):

Debajeet-1411:
@tonypzy Thanks for the clarification. That makes sense-this is better suited as a unit test than an end-to-end Cypress test.

rajnisht7 (commented):
@tonypzy it works good in ui but what about adding info for the same

Manik-Khajuria-5 (approved):
@tonypzy I agree with @rajnisht7 idea of adding in the ui section also

tonypzy:
> @tonypzy it works good in ui but what about adding info for the same

This is the toast effect.
image

Saksham-Sirohi (commented):

- No blocking issues.

rajnisht7 (approved):

vedansh-5 (approved):

PR #1557, opened 2026-07-04
Migrate Cypress export.js E2E tests to Playwright

Debajeet-1411:
## Summary

- Migrate `cypress/integration/export.js` to `playwright/tests/export.spec.js`.
- Preserve the existing Cypress test coverage using Playwright APIs.

References #1552.

This PR is a continuation of #1555 (`feat: migrate basic E2E test suite to Playwright`).

Issue #1552, opened 2026-07-02
Migrate Cypress tests to Playwright


rajnisht7:
As the Cypress tests are quite flaky and even a piexl diff can lead to test failure, along with this cypress isnt much optimized too. Hence, shifting from cypress test to Playwright will be a better long term solution and is also easy to handle.

Choose the function/file you wish to migrate to Playwright and kindly update it in the comments, so that others can avoid duplicate effort and choose tasks accordingly

Before starting, kindly ensure that you have enough time to complete the migration for the selected file/function. This will help to avoid leaving partially migrated work that someone else has to pick up later, making the migration process smoother and more organized.

Debajeet-1411:
@rajnisht7  I'd like to work on migrating `cypress/integration/basic.js` to Playwright if no one else is already working on it.

If that's okay, I'll start working on it. Thanks!

Debajeet-1411:
@rajnisht7 I'd like to take up the migration of `cypress/integration/export.js` to Playwright

Debajeet-1411:
@rajnisht7 I'd like to take up the migration of `cypress/integration/image.js`.

tayyabazahid147-art:
@rajnisht7 I'd like to take up the migration of cypress/integration/misc.js to Playwright, if it's still available.?

rajnisht7:
> [@rajnisht7](https://github.com/rajnisht7) I'd like to take up the migration of cypress/integration/misc.js to Playwright, if it's still available.?

@tayyabazahid147-art Thank you for your interest but there is already a PR for the same that is #1551, if you had like to contribute you can check other issues or the codebase

tayyabazahid147-art:
@rajnisht7 and what about these 3? cypress/integration/text.js
cypress/integration/properties.js
cypress/integration/modal.js

Debajeet-1411:
@tayyabazahid147-art, text.js is already being migrated in #1551 along with misc.js. As far as I can tell, properties.js and modal.js don't have migration PRs yet, so those should still be available if you'd like to take them up.

tayyabazahid147-art:
@rajnisht7 yes I have done properties.js

inver-byte:
Hi! I'd like to work on migrating modal.js to Playwright, if it's still available. Let me know if that's alright!

PR #1561, opened 2026-07-05
Migrate Cypress image.js E2E tests to Playwright

Debajeet-1411:
## Summary

This PR migrates the Cypress `image.js` E2E test suite to Playwright as part of the ongoing test migration effort.

Addresses #1552.

This work continues the Playwright migration following #1557.


vedansh-5:
Hi @Debajeet-1411, thanks for your contribution, please change the title of this PR to better address the scope of changes.

Debajeet-1411:
@vedansh-5 I've updated the PR title to better reflect the scope of this migration.

Debajeet-1411:
The Install and Build workflow appears to be failing during dependency installation due to an SSL handshake error while downloading the PyTorch CPU wheel from download-r2.pytorch.org. It doesn't appear to reach the project build or test steps.

Manik-Khajuria-5 (changes requested):
@Debajeet-1411 Please fix the failing test , the ssl handshake failure might be  temporary please fix it Thanks for contribution

Debajeet-1411:
@Manik-Khajuria-5 Thanks! I've addressed the failure and pushed the fixes. All CI checks are now passing

Manik-Khajuria-5 (approved):

Manik-Khajuria-5 (changes requested):
@Debajeet-1411 Please make sure a proper timeout is added so that we don’t end up waiting indefinitely.

rajnisht7 (commented):
@Debajeet-1411 Kindly resolve conflicts

Debajeet-1411:
@vedansh-5 @Manik-Khajuria-5 @rajnisht7 
I investigated this issue further and restored the helper close to the implementation used in the earlier Playwright migrations. At this point, 26 tests are passing and only the two compare-mode tests remain failing. Both failures are consistently occurring while opening the environment tree in openEnv(), where Playwright resolves hidden tree nodes even after expansion. I tried multiple approaches (different selectors, visibility checks, tree scoping, and restoring the original helper logic), but couldn't find a solution that fixes these tests without risking regressions in the rest of the suite. I believe the migration itself is complete, but I'd really appreciate guidance on the expected way to interact with this environment tree in Playwright.

rajnisht7:
> @vedansh-5 @Manik-Khajuria-5 @rajnisht7 I investigated this issue further and restored the helper close to the implementation used in the earlier Playwright migrations. At this point, 26 tests are passing and only the two compare-mode tests remain failing. Both failures are consistently occurring while opening the environment tree in openEnv(), where Playwright resolves hidden tree nodes even after expansion. I tried multiple approaches (different selectors, visibility checks, tree scoping, and restoring the original helper logic), but couldn't find a solution that fixes these tests without risking regressions in the rest of the suite. I believe the migration itself is complete, but I'd really appreciate guidance on the expected way to interact with this environment tree in Playwright.

seems like wrong ordering issue, have you tried this
first wait for attached state, then call expandAllEnvGroups and then wait for visible state

have you faced similar issue when you tested locally or was it all green

Debajeet-1411:
Running the full image.spec.js suite locally exposes a few additional failures related to image movement/zoom assertions (small position differences) and callback coordinate assertions, whereas CI consistently reports only the two compare-mode failures.

Debajeet-1411:
I also tried the proposed ordering (attached → expandAllEnvGroups() → visible → click), but the timeout still occurs in both CI and locally for the compare-mode tests. The node is attached successfully, but it never becomes visible after the expansion step.

At this point, could you please advise on the expected approach for selecting environments with the current implementation? Has there been any recent change to the environment tree or rc-tree-select behavior that the Playwright helper should account for?

Debajeet-1411:
Thanks for the suggestion @rajnish7! I tried the proposed ordering, but it wasn't sufficient on its own. I further refined the shared helper (visible tree expansion, more robust dropdown handling, and safer environment selection), and the full image.spec.js Playwright suite now passes locally (17/17). I'd appreciate a review to confirm this aligns with the intended approach.

Jayantparashar10 (approved):

Manik-Khajuria-5 (approved):

vedansh-5 (approved):

vedansh-5 (approved):

tonypzy (approved):

rajnisht7 (approved):

Saksham-Sirohi (commented):

- Nice use of `getByText(.

Saksham-Sirohi (approved):

PR #1565, opened 2026-07-06
Fix: numeric title 0 raises AssertionError due to false check in _title2str

Jayantparashar10:
## Description
Change the false check in `_title2str` from `if opts.get("title"):` to `if opts.get("title") is not None:` so numeric titles like 0 are converted to strings before `_assert_opts` runs.

## Motivation and Context
Passing a numeric title of 0 to any plot function crashes with `AssertionError: title should be a string`:

```python3
  viz.line(X=[1, 2], Y=[1, 4], opts={"title": 0})
  # AssertionError: title should be a string
```

` _title2str` already handles numeric titles by converting them to strings with a warning, but the guard `if opts.get("title"):` treats 0 as false and skips conversion entirely. `_assert_opts` then sees an int and raises.

The fix is `is not None`, which is the right check when the intent is "was a title provided at all."

## How Has This Been Tested?
Ran `viz.line()` with `opts={"title": 0}` against a live Visdom server before and after the fix.

Before:
  `AssertionError: title should be a string`

After:
  `Numerical title 0 has been cast to a string`
  `Numerical title 0.0 has been cast to a string`

All three cases passed — `title=0`, `title=0.0`, and `title=None` — with the correct windows rendered.

image

## Types of changes

## Checklist:

## Summary by Sourcery

Bug Fixes:
- Allow numeric plot titles such as 0 to be correctly converted to strings instead of causing an assertion error in title validation.


Manik-Khajuria-5 (approved):
@Jayantparashar10 LGTM!

rajnisht7 (approved):

tonypzy (approved):

vedansh-5 (approved):

PR #1545, opened 2026-07-01
Feat/issue 686 Add vis.pointcloud3d(): WebGL 3D point cloud pane

AfrinHossai:
## Description
Adds a first-class 3D point cloud visualization primitive to Visdom, `vis.pointcloud3d(xyz, rgb=None, win=None, env=None, opts=None)`, backed by Three.js BufferGeometry and a custom OrbitController. `xyz` is an (N, 3) float array of point coordinates; `rgb` is an optional (N, 3) color array accepted as integer [0, 255], float [0, 1], or float [0, 255]. Supported `opts` keys: `title`, `markersize` (default 2.0), `opacity` (default 1.0), `bgcolor` (default '#ffffff'), `show_axes` (default True), `default_color` (default [40, 40, 40]), `max_points`, `downsample` ('stride' or 'random'), and `seed`. On the Python side, input is validated, optionally downsampled (hard cap of 500,000 points), and the resulting typed arrays are base64-encoded into a versioned payload envelope, then sent over the existing `_send` JSON event path (no new server routes or transport). On the JavaScript side, a new `PointCloudPane` React class component renders the cloud using `THREE.BufferGeometry` + `THREE.Points` with size attenuation, driven by a dirty render loop (RAF fires only on data/camera/resize changes, not continuously) and a custom `OrbitController` supporting left-drag rotate, scroll-to-zoom, shift-drag pan, double-click reset, and WebGL context-loss recovery. Reactive opts updates (opacity, markersize, default_color, show_axes) are applied without a full geometry rebuild when the underlying point data is unchanged. On the server side, `UpdateHandler` and the `wrap_func` type guard were extended to recognize the `pointcloud3d` type; a manual JSON patch builder was added for pc3d updates to avoid running `jsonpatch.make_patch` (which would block the IOLoop) over multi-MB base64 payloads, and to ensure `contentID` only changes when point data actually changes (so opts-only updates don't defeat the JS reactive fast path). Rendering opts are now routed through a nested `p["opts"]` dict for this pane type, and the WebSocket `max_message_size` was raised to 64 MiB to accommodate large point clouds with RGB data.
## Motivation and Context
[...]

Issue #686, opened 2019-10-28
When will it support 3D point cloud visualization?


AkiraHero:
**Is your feature request related to a problem? Please describe.**
It will be convenient if Visdom support  point cloud visualization in deep learning about 3D vision,  like project using PointNet.
 
**Describe the solution you'd like**
Considering large scale point cloud collected for autonomous driving or 3D reconstruction, a visualization using WebGL maybe a good choice, which supports mouse interaction.

**Describe alternatives you've considered**

**Additional context**

JackUrb:
We don't currently have any plans to extend beyond the support of what Plotly's 3d scatterplots (which do have [webgl support](https://github.com/facebookresearch/visdom#visscatter)) are able to manage. Let us know if the webgl-backed plotly plots are enough to support your case.

If not, if you're interested in scoping out options for supporting even larger scale 3d point plots please put any implementation ideas you may find here (existing libraries, etc.) so that anyone interested in contributing to this issue would be able to work on it.

mjstevens777:
@AkiraHero I have made a wrapper around THREE.js to display point clouds and other 3D shapes in visdom (as well as jupyter notebooks). It takes numpy arrays as input, converts them into HTML/javascript, and renders the result using `viz.text()`. Let me know if you would like me to share.

I imagine it would also be pretty easy to extend something like [pythreejs](https://pythreejs.readthedocs.io/en/stable/index.html) to do the same thing.

AfrinHossai:
Hi! I’m interested in working on this issue as part of my AI301 open-source contribution project. 

I’m especially interested in the 3D computer vision use case here. An interactive point-cloud visualization pane would be useful for workflows involving PointNet-style models, autonomous-driving point clouds, robotics data, and 3D reconstruction. My initial plan is to first understand how Visdom currently implements panes/plot types, then explore a WebGL-based prototype with basic interactions such as rotate, pan, and zoom. I'll start by reproducing the problem locally this week. Any pointers on where to start in the codebase would be appreciated!

A few initial questions:
1. Is this feature still desired in the current Visdom codebase?
2. Would maintainers prefer this as a built-in pane type or as an extension/custom visualization component?
3. Are there preferred point-cloud input formats or data structures I should support first, such as NumPy arrays with XYZ or XYZRGB values?

Jayantparashar10:
Issue is already solved using scatter plot 3d

PR #1575, opened 2026-07-08
[visdom-686] feat(pointcloud3d): add vis.pointcloud3d() Python client API

AfrinHossai:
[visdom-686] feat(pointcloud3d): add vis.pointcloud3d() Python client API

## Description
Adds the Python client method for Visdom's new 3D point cloud pane: `vis.pointcloud3d(xyz, rgb=None, win=None, env=None, opts=None)`.

- `xyz`: an `(N, 3)` float array of point coordinates.
- `rgb`: an optional `(N, 3)` color array, accepted as integers in `[0, 255]`, floats in `[0, 1]`, or floats in `[0, 255]`.
- `opts`: `title`, `markersize` (default `2.0`), `opacity` (default `1.0`), `bgcolor` (default `'#ffffff'`), `show_axes` (default `True`), `default_color` (default `[40, 40, 40]`), `max_points`, `downsample` (`'stride'` or `'random'`), and `seed`.

Internally, input is validated and normalized (`_pc3d_validate_xyz`, `_pc3d_validate_rgb`, `_pc3d_validate_opts`), optionally downsampled (`_pc3d_downsample`, hard cap of 500,000 points), and the resulting typed arrays are base64-encoded into a versioned payload envelope (`_pc3d_encode_array`, `_pc3d_compute_bounds`), then sent over the existing `_send` JSON event path — no new server routes or transport. A `py/visdom/__init__.pyi` type stub is included for the new method. Also bumps `py/visdom/VERSION` from `0.2.4` to `0.3.0` (see checklist note below).

## Motivation and Context
Visdom currently has no native way to visualize unstructured 3D point data (e.g. LiDAR scans, 3D reconstructions, point-based ML outputs) other than shoehorning it into a 2D scatter/scatter3d plot, which doesn't support interactive orbit/pan/zoom navigation or efficient rendering of large point sets. Resolves the client-API portion of #686.

This is **1 of 4 PRs** splitting up what was originally #1545, which had grown into a single 3,300+ line PR across the whole stack and become too large to review as one unit:
1. **This PR** — Python client API
2. Server-side wiring — branch `feat/pointcloud3d-2-server-handlers`
3. Frontend `PointCloudPane` component — branch `feat/pointcloud3d-3-frontend-pane`
4. Examples, Cypress test, docs — branch `feat/pointcloud3d-4-examples-docs`

[...]

PR #1576, opened 2026-07-08
[visdom-686] feat(pointcloud3d): wire pointcloud3d pane type into the server

AfrinHossai:
[visdom-686] feat(pointcloud3d): wire pointcloud3d pane type into the server

## Description
Extends the Tornado server to recognize the new `"pointcloud3d"` pane type end-to-end (window creation + live updates):

- `py/visdom/utils/server_utils.py`: `window()` routes `pointcloud3d` through a flat `content` payload (no `data` subkey), matching the shape the client sends, instead of the Plotly `data`/`layout` shape used by most pane types. `update_window()` routes rendering opts (`markersize`, `opacity`, `bgcolor`, `show_axes`, `default_color`, ...) through a nested `p["opts"]` dict instead of scattering them as top-level window keys. `title`/`width`/`height` remain top-level since they're UI-frame concerns the pane wrapper reads directly.
- `py/visdom/server/handlers/web_handlers.py`: `UpdateHandler._update_packet_pc3d()` is a hand-written JSON-patch builder used only for this type, instead of `jsonpatch.make_patch()`. Two reasons: (1) diffing multi-MB base64 point-cloud payloads with `jsonpatch` would block the Tornado IOLoop; (2) `contentID` should only bump when point data actually changes, not on opts-only updates — otherwise the JS reactive fast path (which skips a full geometry rebuild when `contentID` is unchanged) would never fire. `update()` gets a `pointcloud3d` branch for full-content replacement, and the packet is added to the exclusion list that skips the generic `content["data"][0]` assumptions the Plotly-style branches rely on.
- `py/visdom/server/handlers/base_handlers.py`: raises `BaseWebSocketHandler.max_message_size` to 64 MiB (from Tornado's 10 MiB default) so a full-size point cloud (up to 500k points with RGB, ~10-12 MB base64-encoded) isn't silently dropped by the websocket transport.

Also bumps `py/visdom/VERSION` from `0.2.4` to `0.3.0` (see checklist note below).

## Motivation and Context
Part of #686. Without this, the server has no idea what a `pointcloud3d` packet is: `window()` would either raise a `KeyError` reaching for `content["data"]`, or (if it fell into the generic branch) attempt to diff a multi-MB payload with `jsonpatch` on every update, blocking the event loop for other connected clients.

This is **2 of 4 PRs** splitting up what was originally #1545:
1. Python client API — branch `feat/pointcloud3d-1-client-api`
2. **This PR** — server-side wiring
3. Frontend `PointCloudPane` component — branch `feat/pointcloud3d-3-frontend-pane`
4. Examples, Cypress test, docs — branch `feat/pointcloud3d-4-examples-docs`

This PR has no dependency on PR 1 — the tests below hand-build fixture dicts and call `server_utils`/`web_handlers` directly rather than going through `vis.pointcloud3d()`, so it's independently mergeable in either order relative to PR 1.

## How Has This Been Tested?
Added `test/test_pointcloud3d_server.py` with 6 unit tests (`TestServerWindowHandler`) covering: `window()` routes a `pointcloud3d` packet via the content path (not the Plotly branch) and doesn't raise `KeyError`; `window()` passes opts through correctly on creation; `update_window()` updates the nested `opts` dict (not top-level scatter); and `update_window()` is no-op-safe on an empty update and preserves existing opts keys not touched by a given update.

Ran via `pytest test/test_pointcloud3d_server.py`: **6 passed**. Also ran `black --check` (v23.1.0, CI-pinned) on all three touched `py/` files: clean.

Not covered by an automated test in this PR: `UpdateHandler._update_packet_pc3d()`'s manual patch-building logic itself (contentID/version bump behavior) is exercised indirectly through `update()`/`update_window()` but has no direct unit test of its own — flagging this as a gap worth a follow-up test rather than silently leaving it uncovered.

Not applicable — server-internal change, no UI.

## Types of changes

## Checklist:

- [x] I adapted the version number under `py/visdom/VERSION` according to [Semantic Versioning](https://semver.org/) — bumps `0.2.4` → `0.3.0`. Note: this repo's usual convention (per `git log -- py/visdom/VERSION` and `.agents/skills/release-process/`) is that version bumps land in a separate maintainer release PR to `master`, not in feature PRs — bumping it here, and identically in PRs 1/3/4, is a deliberate exception for this stack. All 4 PRs bump to the exact same target value, so they won't conflict with each other regardless of merge order.
- [ ] My change requires a change to the documentation. — this is internal server wiring with no separate user-facing docs entry; the `vis.pointcloud3d()` API docs (which describe user-visible behavior) are covered by PR 1 and PR 4.


rajnisht7:
Closing in favor of #1575

PR #1603, opened 2026-07-15
feat: add VisdomSklearnLogger with auto-logging for sklearn plain estimators and CV search

Jayantparashar10:
## Description
Adds `py/visdom/loggers/` with `VisdomSklearnLogger` — a context manager that patches sklearn estimator `fit()` calls to log results to Visdom automatically.

- `GridSearchCV` / `RandomizedSearchCV`: bar chart of `mean_test_score` per parameter combination, summary pane showing `best_score_`, `best_params_` and fit time.
- All other estimators: summary pane showing dataset shape, fit time, train score, and the full `get_params()` dict. Both panes render as HTML tables.
 
Also added `example/train_sklearn_example.py` demonstrating all three paths (plain classifier, plain regressor, grid search).

## Motivation and Context
There was no way to visualize sklearn training runs in Visdom without manually calling `viz.text()` or `viz.bar()` after every fit. This adds opt-in auto-logging behind a single `enable()` call or context manager, with no changes required to model or fit code.

## How Has This Been Tested?
Ran `example/train_sklearn_example.py` against a live Visdom server. 
Verified:
- One text pane per plain estimator (RandomForestClassifier, Ridge), each showing correct dataset shape, fit time, train score, and params.
- One bar chart, one summary pane for GridSearchCV, with correct `mean_test_score` values per combo and correct `best_params_`.

image

## Types of changes

## Checklist:


Manik-Khajuria-5 (changes requested):
@Jayantparashar10 Can we make auto log re callable ?

Jayantparashar10:
> @Jayantparashar10 Can we make auto log re callable ?

@Manik-Khajuria-5 Thanks for pointing this out, I have updated the code so the patched `fit()` uses a shared `_active` logger at call time instead of freezing it. Please review it again.

tonypzy (changes requested):
log_cv() assumes that cv_results["mean_test_score"] always exists, but the score key may instead be mean_test_acc or mean_test_f1. Can we select a key matching mean_test_*?

Jayantparashar10:
> log_cv() assumes that cv_results["mean_test_score"] always exists, but the score key may instead be mean_test_acc or mean_test_f1. Can we select a key matching mean_test_*?

@tonypzy Thanks for pointing this out, I have updated the code please review it again

tonypzy (approved):

Manik-Khajuria-5 (changes requested):
@Jayantparashar10 I think importing visdom inside the function and make client of visdom inside it is not the best way What would u say ?

Jayantparashar10:
> @Jayantparashar10 I think importing visdom inside the function and make client of visdom inside it is not the best way What would u say ?

I've added the import to make the call simple for the user, no need to `import visdom` or write `viz = visdom.Visdom()` just call the logger:
```python3
from visdom.loggers import VisdomSklearnLogger
VisdomSklearnLogger.autolog(env="sklearn")
```
For a custom server viz is still passed explicitly `autolog(viz, env)`, documented in the README, so the internal construction is only a convenience for the default localhost case.

rajnisht7 (approved):

vedansh-5 (approved):

Manik-Khajuria-5 (approved):

PR #1498, opened 2026-06-21
 feat: render hyper-parameter pane on the frontend (Layer 3 - 2)

Manik-Khajuria-5:
## Description
Adds the frontend `hparams` pane so `vis.hparams()` windows are drawn in the browser instead of falling through to the plot fallback. This is the frontend counterpart to the existing `/experiments/hparams` endpoint.

## Integration note

This PR should be merged after #1613 , as it depends on the lower-layer changes introduced there.

## Changes
- **`js/panes/HParamsPane.js`**  new pane (PlotPane-style `React.memo`) that reads the flattened records payload from window content and renders empty / error / summary states. Leaves an `.hparams-views` container as the mount point for the table / parallel-coordinates / SPLOM / filter views in follow-up PRs.

- **`js/settings.js`**  register `hparams` in `PANES` and `PANE_SIZE`.
- **`py/visdom/static/css/hparams.css`** (+ `index.html` link)  pane styling.
- **`experiments_handler.py`**  switch `/experiments/hparams` from the persist-only `_store_window` to the standard `register_window`, so the pane broadcasts to clients and appears live (broadcasting was deliberately deferred until this pane existed).

## Screenshot : 

screenshot-2026-07-18_17-29-30

## Testing
- ESLint + Prettier clean; black clean.
- Manually verified: `vis.hparams()` opens a live pane showing the run summary and per-run status.

## Scope
Pane scaffold only  the table, parallel-coordinates, SPLOM, filter and comparison views land in subsequent PRs.

PR #1616, opened 2026-07-19
feat(hparams): sortable, filterable run table with metric color spine (Layer3 - 3)

Manik-Khajuria-5:
## Description
Replaces the placeholder run list in the `hparams` pane (from the Layer-3 B1 scaffold) with `HParamsTable`  a real run comparison matrix:

- **Sort** by any column via header click or a `sort by` tree dropdown, with a 3-state direction toggle; sort semantics mirror the backend (`_order_key`/`_sort_pairs`): numbers before strings, missing/NaN always last.
- **Filter** runs client-side across name + all param/metric/tag values.
- **Color spine**  `color by` shades a numeric param or metric column on a ramp off the Visdom blue, so "which run won" is scannable at a glance.
- **Grouped columns** (`run │ params │ metrics │ tags`) with vertical dividers, sticky header + run column, per-run selection (hook for a later compare view).
- Dropdowns reuse `rc-tree-select` to match the environment selector; a centered pane title is added via a per-type bar class (no other pane affected).

## Integration note

This PR should be merged after #1498 , as it depends on the lower-layer changes introduced there.

## Motivation and Context
B1 landed the hparams pane but only rendered a placeholder run list. This makes it a usable dashboard for comparing experiment runs by hyper-parameter and metric (Layer 3 of the experiment-management work).

## How Has This Been Tested?
- `npm run build`; manual smoke by seeding 4 runs and opening the pane via `vis.hparams(env_ids=[...])`  verified sort/filter/color-by/sele two empty states.
- Sort/format/filter helpers exercised in Node against the backend
- Backend hparams tests pass (`test_experiment_hparams.py`, `test_hparams.py`).
- ESLint + Prettier clean.

## Video (if appropriate):

https://github.com/user-attachments/assets/becf5b4a-9cc3-45da-a311-88fb7985b1f2

## Types of changes

## Checklist:

## Summary by Sourcery

Turn the hparams pane into a usable run comparison dashboard.

New Features:
- Replace the placeholder hyperparameter run list with an interactive comparison table for sorting, filtering, coloring, grouping, and selecting experiment runs.

Enhancements:
- Align client-side value formatting and sorting behavior with backend ordering, including handling of missing and non-finite values.
- Improve the hparams pane layout with a scrollable table, sticky headers and run columns, grouped fields, accessible controls, and contrast-aware metric coloring.

Chores:
- Set a default Hyperparameters title for hparams windows while preserving caller-provided options.


Saksham-Sirohi (commented):

No issues in the sampled diff.

tonypzy:
image
I found the comment icon on the right side disappears. Please confirm if this is expected behavior.

tonypzy:
ideally, it should be like this 
image

tonypzy (changes requested):
please resolve comments

Manik-Khajuria-5 (commented):
@tonypzy Can u please review this PR again I have resolved the conficts

vedansh-5 (changes requested):
Nice work — the pure-helper split in `hparamsUtils.js` is the right call, the memoization on rows/columns is done properly, and the `overflow: auto` → `hidden` swap on `.content-hparams` is correct given `.hparams-body` already establishes the flex column. A few things need addressing before this lands, mostly around the backend-parity claim and the color spine.

I checked the integration points too, and they're clean: `draggableHandle={'.bar'}` means the filter input can't start a pane drag, `rc-tree-select/assets/index.less` is already imported globally in `main.js`, and the panes that subscribe to `global.event` all gate on `isFocused`, so typing in the filter won't fire stray `KeyPress` callbacks at the server.

## Blocking

**1. Boolean sorting doesn't match the backend.**

`_order_key` in `experiments/store.py` returns `(1, 0.0, str(value))` for bools, and Python's `str(True)` is `"True"`. The JS `orderKey` returns `String(value)` — `"true"`. Both land in the string bucket and compare as text, so case flips the result:

> A column holds `True` and `"alpha"`. Backend: `"True" < "alpha"` (`T`=84 < `a`=97) → True sorts first. Frontend: `"true" > "alpha"` → alpha sorts first.

Since the file header says the comparator "intentionally matches `visdom.experiments.store._order_key`", this should be `value ? 'True' : 'False'` in the sort key. Display formatting can stay lowercase — separate concern.

**2. Large non-integer values render incorrectly.**

`formatValue` does `parseFloat(value.toPrecision(4))` with no scientific-notation fallback, so a metric of `1234567.89` displays as `1235000`. Integers escape via the `Number.isInteger` branch, so this only hits floats — which is exactly where metrics live. A run comparison table showing a plausible-looking but wrong number is worse than one that shows nothing, so I'd fix this before the rest.

**3. The color spine fails WCAG AA in its mid-range.**

The `t > 0.62` text-colour threshold sits in a contrast dead zone. Measured against the ramp:

| `t` | background | text | contrast |
|---|---|---|---|
| 0.60 | `rgb(129,149,191)` | `#333` | **4.20:1** |
| 0.62 | `rgb(126,146,189)` | `#fff` | **3.11:1** |

AA needs 4.5:1 for 12px text. The endpoints are fine (11.0:1 and 6.8:1) — it's only the crossover. Deriving the text colour from the mixed colour's relative luminance instead of a fixed `0.62` fixes the choice; the ramp may also need compressing so both sides clear 4.5.

**4. `selected` is never reconciled with `records`.**

hparams panes re-render on live updates, and env_ids of runs that have disappeared stay in the set. The toolbar reads `selected.size`, so it can report "5 selected" with 2 rows visible — also after filtering, since selection deliberately survives filter changes. Either prune the set when `records` changes, or derive the displayed count from visible rows.

**5. No tests.**

`hparamsUtils.js` was extracted specifically to be testable, and the description says the helpers were "exercised in Node" — please commit that. Nothing in this stack has E2E coverage yet, and this PR adds the pane's entire interactive surface (sort, filter, colour-by, selection), so it's the natural place to start. A comparator parity test in particular is exactly what would have caught #1.

## Non-blocking

**6. `Infinity` is indistinguishable from missing, and sorts differently than the backend.** `isNumeric(Infinity)` is `false`, so it renders as `—` (same glyph as absent), isn't right-aligned, and `orderKey` puts it in the string bucket as `"Infinity"`. The backend's `isinstance(value, (int, float))` check puts it in the numeric bucket. A diverged loss is a realistic way to hit this.

**7. NaN parity.** `isMissing` treats NaN as missing and sorts it last; `_sort_pairs` only routes absent/`None` to `missing`, so NaN stays in `present`. Either match it or narrow the comment's parity claim to absent values.

**8. `sort.by` and `colorBy` go stale when columns change.** If a re-send changes `metricKeys`, `accessorFor` and the extent lookup both return null and silently fall back to unsorted/uncoloured, but the TreeSelect still shows the dead value and the caret still points at a column that may be gone. Worth resetting both when `columns` changes.

**9. Selection keys on `env_id` while row keys fall back to `index`.** If a record ever lacks `env_id`, every such row shares the key `undefined` and toggles as one. The backend does always set it, so this is a guard-rail — but the inconsistent fallback invites it.

**10. Two controls, two sort behaviours.** `handleSortSelect` uses `prev.dir || 'asc'`, so picking a column from the dropdown after a descending sort starts it descending, while clicking that same column's header always starts ascending.

**11. Direction button labels disagree.** `title` describes the action ("Ascending — click for descending"), `aria-label` describes the state ("Sort ascending"). Screen reader users get the opposite semantics from sighted users.

**12. Missed memoization.** `bands`, `sortTreeData`, `colorTreeData`, `colorParamCols` and `colorMetricCols` rebuild on every render while everything around them is `useMemo`'d. The changing `treeData` identity makes rc-tree-select rebuild its tree on each keystroke in the filter box.

**13. Spine extent uses all `records`, not `filtered`.** Probably deliberate — a stable scale beats one that rescales as you type — but it's non-obvious enough to deserve a comment.

**14. Dead CSS.** `.hparams-runs` and `.hparams-run` are orphaned now that the `<ul>` is gone, and `.hparams-summary`'s `position: sticky` is inert now that scrolling has moved into `.hparams-table-scroll`.

## Questions

- Just to flag it for whoever merges: `dev` currently has no hparams files at all, so this branch only builds on top of #1498. I reviewed it on that assumption. Worth keeping the stated merge order visible so this doesn't get picked up out of sequence.
- Could you check both dropdowns in a pane resized down to roughly 200px tall? rc-tree-select portals to `document.body` by default so it should survive `overflow: hidden` and the grid's transforms, but the env selector is the only existing usage and it lives in the navbar.
- The grid wrapper has `onBlur={blurPane}`, so opening a portaled dropdown moves focus outside it. Harmless for hparams itself, but does the pane visibly lose its focus state each time?
- The description mentions "a centered pane title via a per-type bar class (no other pane affected)" — I don't see anything in these five files touching `Pane.js` or adding a bar class. Is that part of #1498, or did it get dropped?
- `opts.setdefault("title", "Hyperparameters")` flows into `p["title"]` in `window()` and gets persisted into env state, so existing hparams windows will pick up a title on their next update. Fine by me, but worth a line in the description since it's user-visible.

vedansh-5:
@Manik-Khajuria-5 I triggered a review from AI - please address the concerns / comment on the ones you do not feel are necessary.

Manik-Khajuria-5 (commented):
@vedansh-5 Yeah, I checked these points. The first and fourth ones are already outdated since the dev merge includes the HParams files and the centered title was removed . I also verified the dropdown and onBlur behavior the dropdown can overflow a small pane, and onBlur doesn’t trigger because focus stays on the tree-select input. I’ll fix the dropdown behavior and update the PR description to mention the persisted default title. I have also made changes according to recent suggestions Please review this PR again

Jayantparashar10 (approved):

vedansh-5 (approved):

rajnisht7 (approved):

PR #1626, opened 2026-07-21
Optimize update handler patch generation for streaming updates

Pragati5-DEBUG:
## Summary
- Optimize `UpdateHandler.update_packet()` for high-frequency streaming updates (`text`, `image_history`, `plot_history`)
- Add targeted fast-path JSON patch generation to avoid expensive `jsonpatch.make_patch` + deep-copy diffing on predictable update paths
- Add roundtrip correctness tests and a micro-benchmark script to validate patch behavior and measure throughput

## Why
Visdom update handling can become CPU-heavy under frequent streaming updates. This change reduces per-update server work in common hot paths while preserving patch correctness.

## Benchmark (10,000 iterations)
Before:
- text: 97,417 ops/s
- image_update_selected: 45,934 ops/s
- plot_history_append: 61,109 ops/s

After:
- text: 118,360 ops/s
- image_update_selected: 77,409 ops/s
- plot_history_append: 82,224 ops/s

## Test plan
- [x] `pytest py/tests/test_update_handler_patch.py`
- [x] `python py/tests/bench_update_packet.py --iterations 10000`
- [ ] `pytest -m "not server"`
- [ ] Manual smoke test: start server and verify text/image/plot history updates still render correctly in browser

Screenshot 2026-07-21 144026

## Summary by Sourcery

Optimize UpdateHandler JSON patch generation for common streaming update pane types to reduce server CPU usage while preserving patch correctness.

New Features:
- Add a fast-path update_packet helper that generates targeted JSON patch operations for text, image_history, and plot_history panes on predictable update types.
- Introduce a micro-benchmark script for measuring update_packet throughput across common streaming update scenarios.

Tests:
- Add roundtrip tests to verify that fast-path patch operations produce the same document state as full jsonpatch diffs for text, image_history selection updates, and plot history appends.


Pragati5-DEBUG:
Hi @Saksham-Sirohi @mariobehling @hpdang — this PR is ready for review/merge when you have a moment. Thanks!

Manik-Khajuria-5 (changes requested):
@Pragati5-DEBUG I read ur code what I found that the append path could use add content instead of replacing the whole content array . Also, it might be worth guarding selected with if selected in p to avoid emitting null on the early return path.

Pragati5-DEBUG:
Thanks @Manik-Khajuria-5 — makes sense.

On the append path: I’ll switch to `add` on `/content/-` for the new item instead of replacing the whole array.

One question: for `image_history`, when we hit `max_image_history` and trim old items, should we still fall back to `replace /content` (so client stays in sync), and only use `/content/-` when nothing was trimmed?

Also will guard `selected` with `if "selected" in p` so we don’t emit null.

Manik-Khajuria-5 (changes requested):
@Pragati5-DEBUG good question, but I'd say no fallback that would basically cancel the optimization and also please mention in pr description json_store .py changes also  and also please resolve the conflicts

Manik-Khajuria-5 (approved):

Saksham-Sirohi (changes requested):

The update fast path always emits JSON Patch `add` for `/contentID` and `/selected`. Those paths usually already exist; `add` fails under RFC 6902 and can desync `fast-json-patch` clients. Use `replace` when the key is present.

Jayantparashar10:
@Pragati5-DEBUG Thank you for your contribution. This PR is scheduled to be closed due to inactivity for more than a month.

Jayantparashar10:
@Pragati5-DEBUG Thank you for your contribution! Since there has been no activity on this PR for more than a month, we are closing it for now. Thanks again for your contribution!

PR #1655, opened 2026-07-26
Fix matplot pt-unit stripping inflating whole-number dims 100x (#1654)

### LINKED ISSUE #1654: matplot(resizable=True) can silently produce 100x oversized panes for whole-number SVG dimensions
**Bug Description**
matplot() pulls the SVG height/width from two different places depending on settings, and both go through the same cleanup line meant to strip the "pt" unit: `height = height.replace("pt", "00")`

For values with a decimal point (like "345.6pt"), this happens to work fine by coincidence. For whole numbers with no decimal (like "300pt"), it doesn't strip anything, it just tacks on two extra digits, turning "300" into "30000". That's a silent 100x size bug, no error, no warning, just a way-oversized panel.

**Reproduction Steps**
1. Run a plot through matplot() with opts={"resizable": True} (needs 
   bs4 and lxml installed)
2. Use a plot whose SVG height/width comes out as a whole number of 
   points (no decimal)
3. The resulting pane renders 100x too large, silently

**Expected behavior**
The "pt" unit should be stripped correctly regardless of whether the value has a decimal point. Tested this myself:

"300pt".replace("pt", "00") -> "30000" -> 30000.0 (should be ~300)
"345.6pt".replace("pt", "00") -> "345.600" -> 345.6 (correct, by luck)

**Screenshots**

Image

Image

**Client logs:**
no errors or warnings related to matplot/SVG rendering. The pane renders successfully, just at the wrong (100x) size, with no console signal indicating anything is wrong. although i did get a `vendor.js:132 Uncaught (in promise) Error: Uncaught Error: No Listener: tabs:outgoing.message.ready
` it's very likely unrelated noise from a browser extension, not part of visdom at all.

**Server logs:**
Server logs show nothing related to the bug at all, confirming this is a silent failure with no error signal. The only ERROR line present is an unrelated 404 for /favicon.png, not connected to matplot/SVG handling in any way. Full relevant excerpt:

```shell
INFO:tornado.access:200 POST /env/main (::1) 0.18ms
INFO:tornado.access:101 GET /socket (::1) 0.18ms
INFO:root:Opened visdom sub socket from ip: ::1
INFO:tornado.access:200 POST /env/main (::1) 0.64ms
INFO:tornado.access:200 POST /env/main (::1) 1.19ms
``` 

Nothing here hints that the rendered pane is ~100x oversized -- the server has no idea anything is wrong, since the bug is purely in the client-side string math before the SVG is ever sent.

**Additional context**
Only shows up when resizable=True is used and the plot's dimensions happen to be a whole number of points. Plots with decimal dimensions (common default in matplotlib) dodge it by coincidence, which is probably why it's gone unnoticed. Happy to send a PR fixing the suffix-stripping logic if welcome.

Maya-Mohamed:
## Description  
Fixes the `pt`-unit stripping logic in `matplot()` when `resizable=True`.  
The cleanup line used `.replace("pt", "00")`, which substitutes the literal  
`pt` for `00` instead of removing it. For whole-number SVG dimensions like  
`"300pt"` this produced `"30000"` — a silent 100x oversized pane. Decimal  
values like `"345.6pt"` only worked by coincidence (`"345.600"`). Changed  
both the height and width lines to `.replace("pt", "")` so the unit is  
stripped correctly regardless of decimals.  
  
## Motivation and Context  
Fixes #1654. Panes rendered ~100x too large with no error or warning  
whenever a matplotlib plot's SVG dimensions came out as whole points.  
  
## How Has This Been Tested?  
- Reproduced the bug locally with a fake plot emitting `height="300pt"  
  width="400pt"`: before the fix `matplot(..., opts={"resizable": True})`  
  computed `height=42000.0, width=54000.0`; after the fix it computes the  
  correct `height=420.0, width=540.0` (1.4*300, 1.35*400).  
- Confirmed the decimal path is unchanged (`"345.6pt"` still correct).  
- Ran the existing suite: `python -m unittest py.tests.test_plots -v`  
  — all 42 tests pass.  
- Ran `black py` — no formatting changes.  
  
## Types of changes  
  
## Checklist:  
- [x] I adapted the version number under `py/visdom/VERSION` according to Semantic Versioning  

## Summary by Sourcery

Correct matplotlib SVG point-unit handling to preserve accurate resizable pane dimensions.

Bug Fixes:
- Fix matplotlib resizable pane dimensions so whole-number SVG dimensions with `pt` units are no longer inflated by 100x.

Tests:
- Add coverage for whole-number and decimal SVG point dimensions in resizable matplotlib plots.


Maya-Mohamed:
Quick note on testing: the `matplot()` resizable path currently has no  
test coverage, so this fix isn't guarded against regressions.  
  
I'm happy to add a small regression test if useful. My plan would be a  
single `unittest.TestCase` in `py/tests/test_plots.py` (following the  
existing `send=False` + `patch` pattern used by `TestLine`/`TestScatter`)  
that:  
  
- feeds `matplot()` a tiny fake plot whose `savefig` emits an SVG with  
  `height="300pt" width="400pt"`,  
- patches `self.viz.svg` to capture the computed `opts`,  
- asserts `opts["height"] == 420.0` and `opts["width"] == 540.0`  
  (i.e. `1.4*300` / `1.35*400`) instead of the pre-fix `42000.0`/`54000.0`,  
- plus one decimal case (`"345.6pt"`) to confirm that path still works.  
  
It'd be guarded with `@unittest.skipUnless(...)` so it's skipped where  
`bs4`/`lxml` aren't installed, and it only needs `math` (already a stdlib  
import in `visdom/__init__.py`).  
  
Would you like this in this PR, or would you prefer to keep the change  
minimal as-is?

Manik-Khajuria-5 (approved):

Manik-Khajuria-5 (commented):
@rajnisht7 @Jayantparashar10 @tonypzy @vedansh-5 Please review this PR

Saksham-Sirohi (commented):

- Fix looks correct (`"10pt"` no longer becomes `"1000"`).
- Confirm whether the `0.

Maya-Mohamed:
Thanks for the review @Saksham-Sirohi!  
  
Addressed both points:  
  
1. **Regression test** — added a `TestMatplotResizable` case in  
   `py/tests/test_plots.py` using a mocked SVG with `height="432pt"`  
   (and `width="640pt"`). It routes through the `resizable=True`  
   BeautifulSoup path and asserts `opts["height"] == 1.4 * ceil(432)  
   == 604.8` instead of the pre-fix `60480.0`. I also added a decimal  
   case (`345.6pt`) to confirm the values that worked by coincidence  
   before still work. The test is guarded with  
   `@unittest.skipUnless(visdom.BS4_AVAILABLE, ...)` so it skips  
   cleanly where bs4/lxml aren't installed.  
  
2. **Version bump** — good call, I've reverted it. Looking at the  
   history, `py/visdom/VERSION` is only changed in dedicated release  
   commits by maintainers, never bundled with a fix, so I've kept this  
   PR limited to the fix + test and left the release bump to you all.

Manik-Khajuria-5 (changes requested):
@Maya-Mohamed Please resolve the conflicts

Maya-Mohamed:
> @Maya-Mohamed Please resolve the conflicts

Done.

py/tests/test_plots.py
```diff
@@ -631,5 +632,52 @@ def test_layout_is_flat(self):
...
+            '<svg xmlns="http://www.w3.org/2000/svg" '
+            'height="{}" width="{}"></svg>'.format(height, width)
+        )
+
+    def savefig(self, buffer, format="svg"):
+        buffer.write(self._svg)
+
+
+@unittest.skipUnless(visdom.BS4_AVAILABLE, "requires bs4/lxml")
+class TestMatplotResizable(unittest.TestCase):
```
Manik-Khajuria-5, line 655: @Maya-Mohamed Please make it compatible with pytest .

Manik-Khajuria-5 (commented):
@Maya-Mohamed Code changes LGTM! But please go through below mentioned test and see whether it is compatible with current codebase or some changes in requirement.txt is required

Maya-Mohamed:
@Manik-Khajuria-5 Added beautifulsoup4 and lxml to test-requirements.txt. CI installs from  
that file before running pytest, so the TestMatplotResizable tests now  
actually run under pytest instead of being skipped (the skipUnless guard  
was skipping them because bs4/lxml weren't installed in CI). Confirmed  
locally: `pytest py/tests/test_plots.py::TestMatplotResizable -v` shows  
2 passed. Let me know if you'd instead prefer a pytest-native  
`@pytest.mark.skipif` — happy to switch.

Manik-Khajuria-5 (changes requested):
@Maya-Mohamed Thanks for the fix and for confirming it locally! Since we're integrating pytest, I'd lean towards using @pytest.mark.skipif

Maya-Mohamed:
> @Maya-Mohamed Thanks for the fix and for confirming it locally! Since we're integrating pytest, I'd lean towards using @pytest.mark.skipif

thanks for the review! I made the switch to @pytest.mark.skipif.

Manik-Khajuria-5 (changes requested):
@Maya-Mohamed Could you please resolve the merge conflicts? I’ve reviewed your changes, and they LGTM!

Maya-Mohamed:
> @Maya-Mohamed Could you please resolve the merge conflicts? I’ve reviewed your changes, and they LGTM!

done.

Manik-Khajuria-5 (approved):
@Maya-Mohamed LGTM!

Jayantparashar10 (approved):

rajnisht7 (approved):

vedansh-5 (approved):

PR #1677, opened 2026-08-01
Feat: Add dedicated table API

rajnisht7:
## Description
This PR adds dedicated table API having multiple features

## Motivation and Context
adding a table which can be exported

## How Has This Been Tested?
verified the feature and functionalities

can be tested using

```

import visdom

viz = visdom.Visdom(env="table test")

headers = ["Name", "Score", "City"]
rows = [
    ["alpha", 92, "Delhi"],
    ["beta", 85, "Mumbai"],
    ["charlie", 78, "Pune"],
]

win_table = viz.table(
    data=rows,
    headers=headers,
    opts={"title": "Leaderboard (editable)"},
)


records = [
    {"Metric": "Accuracy", "Value": 0.912},
    {"Metric": "Loss", "Value": 0.184},
    {"Metric": "F1", "Value": 0.897},
]

viz.table(
    data=records,
    opts={"title": "Run metrics"},
)


viz.table(
    data=rows,
    headers=headers,
    opts={"title": "Leaderboard (read-only)", "editable": False},
)

```

[Screencast_20260802_021952.webm](https://github.com/user-attachments/assets/b32399a3-2b0e-4514-8a21-f7544b369bf2)

image

image

## Types of changes

## Checklist:

## Summary by Sourcery

Introduce a native, editable table pane type with server-side state management and event broadcasting, alongside a retained HTML table helper.

New Features:
- Add a dedicated `table` API that renders structured, editable tables with configurable headers and persistent server-side state.
- Support table editing operations (cell/header edits, row/column add/delete) via a new `table_edit` socket command and corresponding client API.
- Add a React `TablePane` component with resizing, inline editing, row/column management controls, and CSV export for table data.

Enhancements:
- Preserve the previous HTML table helper under the new `html_table` name to avoid breaking existing usage.
- Ensure table pane metadata includes an `editable` flag so tables can be configured as read-only or editable.


Jayantparashar10 (approved):

py/visdom/server/handlers/socket_handlers.py
```diff
@@ -322,6 +322,197 @@ def on_message(self, message):
...
+            broadcast_packet = {
+                "command": "window_update",
+                "win": win,
+                "eid": eid,
+                "content": patch,
+                "version": p["version"],
+            }
+            broadcast(self, json.dumps(broadcast_packet, cls=NanSafeEncoder), eid)
+
+            tornado.ioloop.IOLoop.current().run_in_executor(
```
Manik-Khajuria-5: @rajnisht7 I don't think we should hand the live mutable state directly to run_in_executor. We should either serialize from a snapshot or debounce saves so the worker never races with ongoing mutations

Manik-Khajuria-5 (changes requested):
@rajnisht7 Please go through the below mentioned review and make changes as required

py/visdom/server/handlers/socket_handlers.py
rajnisht7: implemented

Manik-Khajuria-5 (approved):

vedansh-5 (approved):

Manik-Khajuria-5 (changes requested):
@rajnisht7, I was exploring a few different approaches for this.

The best approach I found is to follow the same pattern used by the other socket handler endpoints. We can keep the environment in the server state and serve it from there. Then, when the backend shuts down, it can automatically persist the state, similar to how the other endpoints already handle it. This way, we can reuse the existing server-side state management instead of introducing a separate persistence flow.

Manik-Khajuria-5 (approved):

PR #1703, opened 2026-08-09
 perf: run environment writes off the IOLoop on a dedicated storage worker (AsyncIO - 1)

### LINKED ISSUE #771: Moving to asyncio
**Problem**
Visdom appeals to me because of the iterative send updates to a server. However, I'm always frustrated to see that visdom get and post requests are 10 times (!) slower than my analysis payload script. This is because visdom uses `requests` under the hood, which is sequential by design.

Switching to asyncio coroutines. In particular, using [`asyncio.create_task`](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task). The `await` mechanism needs to be discussed (when to await? - Either return coroutines to the users to let them handle awaits or make a public function `await` in the visdom client.) One could also add `async` flag in `Visdom.__init__` which can be False by default.

[uvloop](https://github.com/MagicStack/uvloop) is a fast, drop-in replacement of the built-in asyncio event loop. uvloop is implemented in Cython and uses libuv under the hood.

**Additional context**
Take a look at `visdom.scatter` profile botelnecks:
```
[time spent]   [line]
40.9 %         exists = self.win_exists(win, env)
54.0 %         return self._send(data_to_send, endpoint=endpoint)
```
-----
Let me know if you're interested in switching to asyncio.

Best,
Danylo

Manik-Khajuria-5:
## Description

Moves environment persistence off the Tornado IOLoop onto an app-owned `ThreadPoolExecutor(max_workers=1)`, and makes `check_auth` / `check_readonly` transparent to `async def` handlers.

Adds `Application.storage_executor`, `shutdown_storage()`, off-loop save/load helpers in `server_utils.py`, and `LazyEnvData.prime()`. Groundwork only no handler becomes `async def` here.

## Motivation and Context

Tornado runs one thread, so every disk write blocks all other requests. Two defects fixed along the way:

1. `save_all` handed a worker thread the **live** `self.state` while the loop kept mutating it a data race. It now offloads a snapshot taken on the loop.

2. Both decorators discarded the wrapped method's return value, which Tornado  needs to await. Any handler converted to `async def` would have answered an empty 200 with the body never running. Behaviour-neutral today.

## How Has This Been Tested?

Full suite green: **1642 passed** (1159 unit, 483 integration). 

## Types of changes

- [x] Code refactor or cleanup (changes to existing code for improv)

## Checklist:

- [ ] I adapted the version number under `py/visdom/VERSION` accordhttps://semver.org/)

## Summary by Sourcery

Move environment persistence off the Tornado IOLoop onto an application-owned storage executor, tighten server auth/readonly handling, and restructure tests and CI around explicit unit/integration markers.

Enhancements:
- Introduce an Application-level storage executor with atomic, off-loop environment save/load helpers and graceful shutdown plumbing.
- Refine handler decorators and socket/web handlers to correctly propagate async return values, enforce readonly semantics for HTTP routes, and improve error and auth handling.
- Harden plotting and client helpers against malformed data (NaNs/Infs, categorical axes, empty histories) and cap in-memory histories for text, embeddings, images, and plot history panes.
- Make JSONStore writes atomic via staged temp files and adjust package discovery to exclude tests from the distributed package.
- Restructure pytest layout into unit/integration buckets with shared test utilities, expand coverage of client/server behavior, and add a fast unit-test gate plus coverage reporting in CI.

CI:
- Split python-tests workflow into a fast unit gate and a full matrix run with coverage reporting, without enforcing a coverage threshold.

Documentation:
- Document the new testing layout, markers, and CI expectations in AGENTS and testing context docs.

Tests:
- Add extensive unit and integration test suites for client content builders, TSNE helpers, metrics utilities, storage layer, socket command handling, HTTP handlers, and environment/window lifecycle.
- Introduce manual visual-check scripts and detailed testing documentation to guide where new tests belong and how markers should be used.

## Summary by Sourcery

Move environment persistence off the IOLoop while making storage ordering, autosave tracking, asynchronous handler behavior, and shutdown reliable.

Bug Fixes:
- Prevent environment deletions from being overwritten by queued saves by serializing persistence operations on a dedicated worker.
- Preserve pending changes and avoid data races by saving snapshots rather than live in-memory state.
- Ensure socket-originated environment mutations are tracked for autosave and readonly HTTP routes enforce write protection.

Enhancements:
- Move environment persistence and loading off the Tornado IOLoop onto an application-owned single-worker executor with graceful shutdown draining.
- Allow authentication and readonly decorators to propagate wrapped handler results, including coroutines, and add lazy environment priming for asynchronous loads.
- Make JSON environment writes atomic and harden client/server helpers against malformed or unbounded data.

CI:
- Split Python tests into unit and integration CI gates with expanded coverage reporting.

Documentation:
- Document background storage writes and update testing guidance and visual-check documentation.

Tests:
- Expand unit and integration coverage for storage execution, autosave ordering, environment lifecycle, socket commands, HTTP handlers, and client utilities.

Chores:
- Exclude test modules from distributed package discovery.


tonypzy (changes requested):
It seems that deleted environments are being restored by autosave. How about writing a test case?

Manik-Khajuria-5 (commented):
@tonypzy Can u review this PR again I have made suggested changes

tonypzy (changes requested):
good. just need to solve conflicts

vedansh-5 (approved):

Jayantparashar10 (approved):

rajnisht7 (approved):

Saksham-Sirohi (approved):

PR #1704, opened 2026-08-09
chore: Save changed environments automatically instead of only at exit

vedansh-5:
## Description

Environments are now written to disk on a timer rather than only when asked.
The -save_interval option (default 30s) saves any environment changed since the
last pass, and -save_threshold (default 50 updates) writes a busy one out early so
it is not left unsaved for a whole interval. Only changed environments are written.
Setting both to 0 restores the previous behaviour.

SIGTERM now raises SystemExit so the existing atexit save runs.

## Motivation and Context

The only automatic write was atexit.register(app.storage.save_all, app.state) in
run_server.py. Python's default disposition for SIGTERM ends the process without
running atexit handlers, so docker stop, and any process manager that terminates
rather than interrupts, discarded every plot since the last manual save.

Reproduced on current dev: start the server, POST one window to /events, send
SIGTERM, and no environment file is written.

## How Has This Been Tested?

New tests in py/tests/integration/autosave.py cover dirty tracking, both save
triggers, retry when the backend raises, and the handlers that mark their writes.
Full suite: 418 passed.

Manually, against a running server (Python 3.14, tornado):

- a plot reaches disk mid-run on the timer, with the server still up
- with autosaving disabled, SIGTERM alone persists it
- a restarted server lists the environment
- the same sequence on an unmodified dev checkout loses it

## Types of changes

## Checklist:

- [ ] I adapted the version number under py/visdom/VERSION

## Summary by Sourcery

Introduce automatic environment saving with configurable intervals and thresholds, and ensure shutdown via SIGTERM persists environments via existing atexit handlers.

New Features:
- Add configurable automatic saving of changed environments based on a timer and per-environment update threshold.
- Expose -save_interval and -save_threshold command-line options with sensible defaults to control autosave behavior.

Bug Fixes:
- Ensure SIGTERM leads to a normal interpreter shutdown so existing atexit-based environment saving runs instead of discarding unsaved data.

Enhancements:
- Track dirty environments centrally and persist only those that have changed, avoiding unnecessary writes and allowing retries on backend failure.
- Update server handlers to mark environments as dirty whenever windows are registered, updated, or closed, so autosave has accurate change tracking.

Documentation:
- Document the new autosave behavior and the -save_interval and -save_threshold options in the getting-started command-line options guide.

Tests:
- Add integration tests covering dirty environment tracking, autosave triggers (timer and threshold), retry behavior on storage failures, and handler marking of writes.
- Extend test fakes to track dirtied environments to support autosave-related tests.


Manik-Khajuria-5 (approved):
@vedansh-5 LGTM!

Jayantparashar10 (approved):

rajnisht7 (commented):
@vedansh-5 overall looks good, just a few things
- in `UpdateHandler.wrap_func` in `web_handlers.py` when the pane type is embedding, it returns after the update before the call of `mark_dirty(eid)`
- and in `socket_handlers.py` close and undo also modifies the env, kindly look into this too

vedansh-5:
> @vedansh-5 overall looks good, just a few things
> 
> * in `UpdateHandler.wrap_func` in `web_handlers.py` when the pane type is embedding, it returns after the update before the call of `mark_dirty(eid)`
> * and in `socket_handlers.py` close and undo also modifies the env, kindly look into this too

PTAL, I have pushed the fixes

tonypzy (approved):

rajnisht7 (approved):

PR #1714, opened 2026-08-12
fix: use plotlyRef instead of contentID string for Plotly.react target

### LINKED ISSUE #1715: PlotPane crashes intermittently: Plotly.react receives contentID string instead of DOM ref
## Bug

`PlotPane.js`'s render effect calls `Plotly.react(contentID, ...)`, passing the
pane's string id instead of the DOM node already held in `plotlyRef`. Every other
Plotly call in this file (`toImage`, `downloadImage`, `Plots.resize`) correctly
uses `plotlyRef.current`.

Passing a string requires Plotly to resolve it internally via
`document.getElementById`, which can fail if that call runs before the target
element is queryable in the DOM. When it fails, the plot pane crashes and never
renders.

## Console trace

Error: DOM element provided is null or undefined
at Object.Zrt [as getGraphDiv] (plotly-plotly.min.js)
at Object.gvt [as react] (plotly-plotly.min.js)
at R (PlotPane.js:338:12)
at PlotPane.js:205:5
TypeError: t.on is not a function
at PlotPane.js:247:17

## Reproduction

I hit this repeatedly in a real browser session (uploading/switching plot
environments) but could not produce a reliable, deterministic automated repro —
tried both a fresh single render and rapid repeated updates via Playwright, both
passed even against the unfixed code. It appears timing/condition-dependent
rather than deterministic on every render.

## Fix

`plotlyRef.current` is the live DOM ref already used everywhere else in this
file — using it instead of the `contentID` string avoids the
`document.getElementById` race entirely and is unconditionally correct.

piyush182004:
## Summary - fix: use plotlyRef instead of contentID string for Plotly.react target- #1714

Fixes #1715

- PlotPane's render effect called `Plotly.react(contentID, ...)`, passing the pane's
  string id instead of the DOM node already held in `plotlyRef`. Every other Plotly
  call in this file uses `plotlyRef.current`.
- Passing a string requires Plotly to resolve it via `document.getElementById`
  internally, which can fail if that call runs before the element is queryable,
  throwing `Error: DOM element provided is null or undefined` / `TypeError: t.on is
  not a function`, and breaking the plot pane entirely.
- I hit this repeatedly in a real session (console trace below) but could not produce
  a reliable automated repro — tried both a fresh single render and rapid repeated
  updates via Playwright, both passed even against the unfixed code, so this appears
  timing/condition-dependent rather than deterministic. The fix is unconditionally
  correct regardless: `plotlyRef.current` is the live ref already used everywhere
  else in this file, so there's no reason to rely on string-based DOM lookup here.

## Console trace observed
Error: DOM element provided is null or undefined
    at Object.Zrt [as getGraphDiv] (plotly-plotly.min.js)
    at Object.gvt [as react] (plotly-plotly.min.js)
    at R (PlotPane.js:338:12)
    at PlotPane.js:205:5
TypeError: t.on is not a function
    at PlotPane.js:247:17

## Test plan
- [x] `npx eslint js/panes/PlotPane.js playwright/tests/pane.spec.js` — clean
- [x] `pytest -m "not server"` — unaffected (pre-existing unrelated failures only)
- [x] Added two Playwright tests in `pane.spec.js` asserting no uncaught page errors
      during initial plot render and rapid successive updates — general regression
      coverage for this class of error, not a reproduction of this specific race
- [x] Manually verified the fix resolves the crash in-browser

## Summary by Sourcery

Ensure Plotly plots render reliably by targeting the graph DOM node via the existing ref instead of a string content ID and add regression coverage for plot rendering errors.

Bug Fixes:
- Prevent Plotly.react from crashing on plot render by using the pane's plotlyRef DOM node rather than a string-based content ID lookup.

Tests:
- Add Playwright tests that assert no uncaught page errors during initial line plot render and during rapid successive plot updates.
- Extend Playwright test helpers with waitForPlotRender to reliably await plot initialization in UI tests.

## Summary by Sourcery

Fix Plotly plot updates to use the pane’s live graph reference and add regression coverage for rendering errors.

Bug Fixes:
- Prevent Plotly rendering failures by targeting the existing graph DOM reference when updating plots.

Tests:
- Add Playwright coverage for uncaught errors during initial plot rendering and rapid successive plot updates.


piyush182004:
@vedansh-5 @Jayantparashar10 @Manik-Khajuria-5 @rajnisht7 please review

Manik-Khajuria-5 (approved):
@piyush182004 Thanks for contribution This LGTM!

Jayantparashar10 (approved):

rajnisht7 (approved):

vedansh-5 (approved):
Thank you for your contribution. Looks good to me!

PR #1744, opened 2026-08-20
fix: pr_curve computes wrong AUC when precision/recall have tied recall values

### LINKED ISSUE #1745: pr_curve() computes incorrect AUC when precision/recall points have tied recall values
## Bug
`_coerce_curve_xy` in py/visdom/__init__.py sorts precomputed curve
points by x (recall, for pr_curve) with no tiebreak on y (precision).
Real classifier output regularly produces multiple points sharing the
same recall value (sklearn's own precision_recall_curve() output does
this routinely). The arbitrary tiebreak that results can pair the wrong
precision with a recall point, silently corrupting the AUC sum in
_average_precision — no error, no warning, just a wrong number.

## Verified
Across 220+ randomized trials comparing against
sklearn.metrics.average_precision_score, every trial containing tied
recall values showed a mismatch, from <1% up to ~13% off. Also confirmed
live in the browser across 12 independent datasets.

## Fix
PR:- #1744

piyush182004:
## Description

`_coerce_curve_xy` sorted precomputed curve points with `np.argsort(x)`, which orders by `x` alone. For a precision-recall curve that leaves the order of a tied-recall group to the sort's tie-breaking, pairing a recall with the wrong precision and silently moving the area that `_average_precision` computes from it.

`sklearn.metrics.precision_recall_curve` returns several points sharing a recall value for any data whose classes overlap, so this is the normal case rather than an edge case.

The sort now breaks ties on `y`. `pr_curve` passes `y_tiebreak_descending=True`, since precision only decreases as more points are accepted at a fixed recall, so the higher-precision point must come first. `roc_curve` keeps the plain ascending order: it integrates a trapezoid, where a tied-`x` group contributes zero width whatever the order within it.

The tie-break value is widened with `.astype(np.float64)` before negation. Negating an unsigned array wraps rather than changing sign, so a `uint8` precision array of `0`/`1` would otherwise come back as `0`/`255` and order the tied group backwards.

## Motivation and Context

The AUC printed in the pane title is one of the few numbers a user reads directly off a visdom plot, and it was wrong with no error or warning to signal it.

Across ten random datasets, every one diverged from `sklearn.metrics.average_precision_score` — by under 1% in the mildest case and about 13% in the worst.

Fixes #1745

## How Has This Been Tested?

Environment: Windows 11, Python 3.12 (conda), installed from source with `pip install -e .`.

- Confirmed the bug still reproduces on current `dev`: 10/10 random trials diverged from `average_precision_score`
- After the fix: 0/30 PR-curve mismatches and 0/30 ROC mismatches, the latter confirming ROC is unaffected
- Added `test_pr_curve_auc_matches_sklearn_when_recall_ties`, verified to fail against the pre-fix code and pass with the fix
- Added `test_roc_curve_auc_is_unaffected_by_the_pr_tiebreak` as a guard against the tie-break regressing ROC
- Added `test_pr_curve_breaks_recall_ties_whatever_the_precision_dtype`, parametrized over `uint8`/`int16`/`float32`, verified to fail on `uint8` without the widening
- 118 tests in `py/tests/unit/client_metrics.py` pass; `black --check` clean
- Remaining failures in a full local run are pre-existing Windows `MAX_PATH` ones in `data_model.py` and friends, which also fail on a clean `dev` checkout here and pass on the Linux runners

## Types of changes

## Checklist:

## Summary by Sourcery

Fix precision-recall curve AUC calculations for tied recall values without changing ROC curve behavior.

Bug Fixes:
- Correct precision-recall AUC calculations when precomputed points contain tied recall values.
- Preserve ROC AUC behavior while ensuring tie handling works across precision dtypes, including unsigned arrays.

Enhancements:
- Improve curve point ordering by applying an appropriate secondary sort for tied x-values.

Tests:
- Add coverage comparing precision-recall AUC against scikit-learn with tied recalls.
- Add tests for tie ordering across signed, unsigned, and floating-point precision types.
- Add regression coverage confirming ROC AUC remains unaffected.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved precision-recall AUC calculations when multiple points share the same recall value, including support for unsigned, signed, and floating-point data.
  - Preserved correct ROC AUC results while applying precision-recall tie handling.
  - Added an option to skip preflight window checks for image and scatter updates, enabling more direct append workflows.

- **Tests**
  - Added regression coverage for tied-value precision-recall curves and append behavior.

### DIFF
--- a/py/tests/unit/client_metrics.py
+++ b/py/tests/unit/client_metrics.py
@@ -567,6 +567,59 @@ def test_pr_curve_from_precomputed_points_infers_the_baseline(capture_send):
     assert sent["payload"]["data"][1]["y"] == [0.4, 0.4]
 
 
+def test_pr_curve_auc_matches_sklearn_when_recall_ties(capture_send):
+    """Tied recall points must not skew the average precision.
+
+    ``precision_recall_curve`` returns several points sharing a recall value
+    for any data the classes overlap in. Sorting on recall alone leaves their
+    order to the sort's tie-breaking, which pairs a recall with the wrong
+    precision and quietly moves the area.
+    """
+    sklearn_metrics = pytest.importorskip("sklearn.metrics")
+
+    for seed in range(10):
+        rng = np.random.RandomState(seed)
+        size = rng.randint(15, 60)
+        y_true = rng.randint(0, 2, size=size)
+        if y_true.sum() in (0, size):
+            continue
+        # Rounding the scores forces the ties this is about.
+        y_score = np.round(
+            y_true * rng.uniform(0.3, 0.7, size=size) + rng.uniform(0, 1, size=size), 1
+        )
+        precision, recall, _ = sklearn_metrics.precision_recall_curve(y_true, y_score)
+
+        sent = capture_send(lambda v: v.pr_curve(precision=precision, recall=recall))
+
+        title = sent["payload"]["opts"]["title"]
+        auc = float(title.rsplit("AUC=", 1)[1].rstrip(")"))
+        expected = sklearn_metrics.average_precision_score(y_true, y_score)
+        assert auc == pytest.approx(expected, abs=5e-5)
+
+
+def test_roc_curve_auc_is_unaffected_by_the_pr_tiebreak(capture_send):
+    """ROC integrates a trapezoid, so its points keep their plain ordering."""
+    sklearn_metrics = pytest.importorskip("sklearn.metrics")
+
+    for seed in range(10):
+        rng = np.random.RandomState(seed)
+        size = rng.randint(15, 60)
+        y_true = rng.randint(0, 2, size=size)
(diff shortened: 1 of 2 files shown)


piyush182004:
Can you please review this @vedansh-5 @Jayantparashar10 @Manik-Khajuria-5 @rajnisht7

Jayantparashar10 (approved):

rajnisht7 (approved):

Manik-Khajuria-5 (approved):

Jayantparashar10:
@piyush182004 Thanks for your contribution. Please rebase the PR brach with the dev branch to resolve conflicts.

vedansh-5:
Hi, Thanks for your contribution. Please resolve AI comments.

vedansh-5 (approved):

PR #1765, opened 2026-08-26
fix: pr_curve baseline always showed 1.0 instead of true class prevalence

### LINKED ISSUE #1766: pr_curve() baseline always shows 1.0 instead of true class prevalence
## Bug
When `pr_curve()` is called with precomputed `precision`/`recall` points
(rather than raw `y_true`/`y_score`), the "Baseline" line drawn on the
chart is meant to represent the true class prevalence (fraction of
positive examples). Instead, it always evaluates to 1.0 regardless of
the actual data.

py/visdom/__init__.py, ~line 3047 (before fix):
    positive_rate = float(precision[0]) if float(recall[0]) == 0.0 else None

`precision[0]` after sorting is always the `(precision=1, recall=0)`
sentinel point that sklearn's `precision_recall_curve()` appends by
convention — not a measure of prevalence. Precision and recall are both
ratios and don't preserve the underlying class counts, so true
prevalence cannot actually be recovered from precomputed points alone.

## Verified
Across 10 varied datasets with real prevalence ranging from 5% to 95%,
the displayed baseline was 1.0 every time. Confirmed live in the
browser: a dataset with ~28% true prevalence showed a flat baseline
pinned at the top of the chart.

piyush182004:
## Description
When `pr_curve()` is called with precomputed `precision`/`recall` points
(rather than raw `y_true`/`y_score`), the "Baseline" line drawn on the chart is
supposed to represent the true class prevalence. Instead it always evaluated to
1.0 regardless of the actual data, because `precision[0]` (after sorting) is
always the `(precision=1, recall=0)` sentinel point that sklearn's
`precision_recall_curve()` appends by convention — not a measure of prevalence.

Precision and recall are both ratios; neither preserves the underlying class
counts, so true prevalence genuinely cannot be recovered from precomputed
`(precision, recall)` points alone. The fix stops drawing an incorrect baseline
for that path (`positive_rate = None`) rather than guessing. The
`y_true`/`y_score` path was already correct (computes real prevalence directly
from the labels) and is unaffected.

Also updated the `pr_curve` docstring and the README to state that the baseline
— and therefore the second `opts.legend` label — applies only to the
`y_true`/`y_score` input mode.

## Motivation and Context
The baseline on a PR curve is the no-skill reference line; it should sit at the
positive class rate. Drawing it at 1.0 misrepresents model performance — on a
dataset with ~28% true prevalence the baseline was drawn at the very top of the
chart instead of near 0.28, making a good model look no better than chance.

Fixes #1766

## How Has This Been Tested?
- `pytest -m "not server"` — unaffected (pre-existing unrelated failures only)
- `black --check` — clean
- Updated `py/tests/unit/client_metrics.py::test_pr_curve_from_precomputed_points_omits_the_baseline`
  (previously `..._infers_the_baseline`), which asserted the old behavior.
  Confirmed it fails against the pre-fix code and passes with the fix.
- Added `TestPrCurveBaseline` in `py/tests/unit/plots.py` covering both the
  precomputed path (no baseline trace) and the raw path (correct prevalence).
- 10 varied edge cases checked programmatically (balanced classes, 5%/95%
  extreme prevalence, tiny/huge datasets, perfect separation, heavy score ties,
  single-positive-example) — precomputed-points path correctly omits the
  baseline in all 10; `y_true`/`y_score` path still draws the correct baseline
  value in both sanity checks
- Manually verified in-browser: the wrong flat baseline at y=1.0 is gone for the
  precomputed-points case; the `y_true`/`y_score` path still draws a
  correctly-positioned baseline (checked at both low ~15% and high ~80%
  prevalence)

## Types of changes

## Checklist:

## Summary by Sourcery

Correct PR-curve baseline handling by omitting it when class prevalence cannot be recovered from precomputed points.

Bug Fixes:
- Prevent PR curves built from precomputed precision/recall points from displaying an incorrect baseline at 1.0.
- Continue displaying the correct class-prevalence baseline when raw labels and scores are provided.

Documentation:
- Document that the PR-curve baseline and its legend label apply only when using y_true/y_score inputs.

Tests:
- Add coverage confirming omission of the baseline for precomputed points and correct prevalence for raw-label inputs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Precision-recall curves generated from precomputed points no longer display an inferred prevalence baseline.
  - Prevalence baselines remain available for curves generated from raw labels and scores.
  - Excess legend labels are now trimmed to match the plotted traces for precision-recall and ROC curves.
  - Legend requirements are correctly adjusted when a precision-recall baseline is not displayed.

- **Documentation**
  - Updated precision-recall curve documentation to clarify baseline visibility and legend configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1084,7 +1084,7 @@ A baseline showing the true class prevalence is only drawn for the
 
 The following `opts` are supported:
 - `opts.title`      : plot title (`string`; default includes PR-AUC)
-- `opts.legend`     : two legend labels for curve and baseline (`list`); the second label is unused when no baseline is drawn
+- `opts.legend`     : two legend labels for curve and baseline (`list`); only one label is needed when no baseline is drawn
 - `opts.xlabel`     : x-axis label (`string`; default = `Recall`)
 - `opts.ylabel`     : y-axis label (`string`; default = `Precision`)
 - `opts.layoutopts` : additional backend layout options (`dict`)
--- a/README.md
+++ b/README.md
@@ -1078,9 +1078,13 @@ It accepts either:
 - raw binary labels and scores via `y_true` and `y_score`, or
 - precomputed curve points via `precision` and `recall`.
 
+A baseline showing the true class prevalence is only drawn for the
+`y_true`/`y_score` path, since prevalence cannot be recovered from
+`precision`/`recall` points alone.
+
 The following `opts` are supported:
 - `opts.title`      : plot title (`string`; default includes PR-AUC)
-- `opts.legend`     : two legend labels for curve and baseline (`list`)
+- `opts.legend`     : two legend labels for curve and baseline (`list`); the second label is unused when no baseline is drawn
 - `opts.xlabel`     : x-axis label (`string`; default = `Recall`)
 - `opts.ylabel`     : y-axis label (`string`; default = `Precision`)
 - `opts.layoutopts` : additional backend layout options (`dict`)
--- a/py/tests/unit/client_metrics.py
+++ b/py/tests/unit/client_metrics.py
@@ -311,9 +311,14 @@ def test_curve_legend_warns_when_a_legend_is_unusable(legend):
 
 
 def test_curve_legend_passes_a_valid_legend_through():
-    """Two or more elements are accepted verbatim, as a list."""
+    """A legend of the required length is accepted verbatim, as a list."""
     assert _curve_legend(("a", "b"), ["ROC", "Chance"]) == ["a", "b"]
-    assert _curve_legend(["a", "b", "c"], ["ROC", "Chance"]) == ["a", "b", "c"]
+
+
+def test_curve_legend_drops_labels_beyond_the_required_count():
+    """Extra labels would name traces the plot does not draw."""
+    assert _curve_legend(["a", "b", "c"], ["ROC", "Chance"]) == ["a", "b"]
+    assert _curve_legend(["a", "b"], ["PR", "Baseline"], required=1) == ["a"]
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


piyush182004:
@vedansh-5 @Jayantparashar10 @Manik-Khajuria-5 @rajnisht7 please review

Manik-Khajuria-5 (changes requested):
@piyush182004 Thanks for contribution .  Could we update/remove the test for particular change also?

piyush182004:
@Manik-Khajuria-5 thanks for the suggestion,I have added a test in latest commit please check and let me know if any changes are required

Jayantparashar10:
@piyush182004 Please follow the PR template and update the PR details.

piyush182004:
> @piyush182004 Please follow the PR template and update the PR details.

I have updated the pr please check @Jayantparashar10

piyush182004:
please check @vedansh-5 @Jayantparashar10 @Manik-Khajuria-5 @rajnisht7

Manik-Khajuria-5 (changes requested):
@piyush182004 Thanks for contribution. I noticed a small issue with pr curve legend where a single custom label gets ignored. Can u please check it again

piyush182004:
@Manik-Khajuria-5 thanks for informing ...changed it..can you verify once

Manik-Khajuria-5 (changes requested):
@piyush182004 Now I think if the user gives multiple legend labels they are not reduced when only one curve is created Can u please check again

piyush182004:
@Manik-Khajuria-5 please check once fixed it also added 3 regression test and updated legend pass

Manik-Khajuria-5 (approved):
@piyush182004 LGTM!

Jayantparashar10 (changes requested):
@piyush182004 Update the PR branch with the base branch changes.

Jayantparashar10 (changes requested):
@piyush182004 The PR description mentions `TestPrCurveBaseline` in `plots.py` but that file isn't part of this PR. Please look into this.

PR #1775, opened 2026-08-30
fix: bump pane version on every update so patches are not discarded

tarone-saloni:
## Description

`p["version"] += 1` lived in `update_window()`. `UpdateHandler.update()` returns before that helper for `text`, `image_history`, `plot_history` and `table` panes, and the embeddings route never calls it at all — so those types stayed pinned at version `1` for the life of the pane while the server kept broadcasting `window_update` messages against them.

The frontend applies an incremental patch only when the message carries `pane.version + 1` (`updateWindow`, `js/main.js:387`):

```js
if (
  (cmd.win in storeData.panes &&
    cmd.version == storeData.panes[cmd.win].version + 1) || ...
) {
  addPaneBatched(cmd);           // apply the patch
} else if (!_envReloadInFlight.current) {
  _envReloadInFlight.current = true;
  sendEnvQuery(selection.envIDs); // re-download the whole environment
}
```

With both sides stuck at `1` the comparison reads `1 == 2`, which can never be true, so **every** update fell through to `sendEnvQuery()`. The incremental protocol was effectively disabled for the pane types the library is most used
for; only `scatter` and the other trace-based plots ever worked.

### What changed

- The counter moves to a new `UpdateHandler.bump_version()`, called from `update_packet()` — the single point where an accepted update becomes a broadcast — so it advances regardless of which branch `update()` took.
- `update_window()` no longer touches `version`; its docstring now says so and points at the new owner.
- The embeddings route builds its JSON patch by hand, so it bumps too **and** emits the matching `/version` op. Without that op the browser's copy would keepthe old number and fail the check on the *next* update instead of this one.
- An unrecognised embeddings `update_type` now broadcasts nothing at all, rather than an empty patch carrying a version the client cannot reconcile (which cost a full reload for a no-op).
- `bump_version()` reads through `.get()` rather than `+= 1`, so an environment persisted before panes carried a version updates instead of raising `KeyError`out of a request.

### Files touched

| File | Change |
| --- | --- |
| `py/visdom/utils/server_utils.py` | Remove the bump from `update_window()`; document the moved responsibility |
| `py/visdom/server/handlers/web_handlers.py` | Add `UpdateHandler.bump_version()`; call it from `update_packet()` and both embeddings branches; skip the broadcast on an empty embeddings diff |
| `py/tests/unit/pane_versions.py` | **New** — 28 cases covering all six pane types |
| `py/tests/unit/window_builder.py` | The two `update_window` version tests now pin the opposite contract |

## Motivation and Context

A training loop that calls `vis.text()` once per step made the browser re-download the entire environment once per step. The same applied to every `image_history` frame, every `plot_history` frame, and every embeddings
selection. On a large environment this is the difference between shipping a few-hundred-byte JSON patch and re-serialising every pane in the environment, for every single update.

The bug was invisible because the one pane type that *did* work — plots — is the type most tests exercise, so a single working path masked four broken ones.

## How Has This Been Tested?

**Environment:** Python 3.13, `pip install -e .`, `pytest` from `test-requirements.txt`. Branched from `dev` at `792e6e7`.

### Before / after, measured against a live server

Four `/update` calls per pane, reading `state[eid]["jsons"][win]["version"]` afterwards:

| Pane type | Before | After |
| --- | --- | --- |
| `text` | `1` | `5` |
| `scatter` | `5` | `5` |
| `image_history` | `1` | `5` |

Broadcasts captured from a subscribed socket, three text updates:

```
before:  command=window_update  version=1
         command=window_update  version=1
         command=window_update  version=1

after:   command=window_update  version=2
         command=window_update  version=3
         command=window_update  version=4
```

The `after` sequence is exactly the `pane.version + 1` ladder the client walks, so the patches are now applied instead of triggering a reload.

### Test suite

```
pytest -m unit          # passes (CI gate)
pytest -m "not server"  # passes
black --check py        # clean
```

New `py/tests/unit/pane_versions.py` (`pytestmark = pytest.mark.unit`) asserts, for **all six** pane types — `plot`, `text`, `image_history`, `plot_history`, `table`, `embeddings`:

- a single update advances `version` by one;
- repeated updates advance by exactly one each;
- the emitted patch carries a `/version` replace op with the new value;
- the versions a subscribed socket actually receives are the consecutive run starting one past where the pane began;
- an unknown embeddings `update_type` changes nothing and broadcasts nothing;
- a pane with no `version` key at all (an environment persisted before the field existed) updates rather than raising.

Parametrising over every pane type is deliberate: the gap that let one working type mask four broken ones cannot reopen.

The two `update_window` version assertions in `window_builder.py` were inverted rather than deleted, so the helper's new contract — *it does not sequence the broadcast* — is pinned rather than merely unstated.

## Screenshots (if appropriate)

N/A — no visual change. The user-visible effect is that panes update incrementally instead of the dashboard re-fetching the environment on each update.

## Types of changes

## Checklist

- [ ] I adapted the version number under `py/visdom/VERSION` according to [Semantic Versioning](https://semver.org/) — *left at `0.2.4`; happy to bump the patch version if maintainers prefer it on bug fixes*
- [x] My code follows the code style of this project. — `black==23.1.0` clean
- [x] I have updated the documentation accordingly. — docstrings on `update_window()` and `bump_version()` record where the counter lives and why; no README or `__init__.pyi` change needed, as no public API or wire format changed

## Summary by Sourcery

Restore incremental pane updates by consistently advancing and broadcasting pane versions for every accepted update.

Bug Fixes:
- Ensure every accepted pane update advances its version so incremental frontend patches are applied instead of triggering full environment reloads.
- Keep embeddings updates synchronized with client versions and suppress broadcasts for unrecognized update types.
- Allow updates to panes from persisted environments that lack a version field.

Enhancements:
- Centralize pane version sequencing at the point updates are converted into broadcasts rather than in the window-merging helper.

Documentation:
- Document ownership and behavior of pane version sequencing in the update helpers.

Tests:
- Add unit coverage for version progression, emitted patch versions, subscriber broadcasts, embeddings behavior, and legacy panes across all supported pane types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update sequencing so pane versions advance consistently across standard, embedding, and selection updates.
  * Ensured subscribers receive consecutive version numbers for successive pane updates.
  * Prevented unrecognized embedding updates from producing empty patches or marking environments as changed.
  * Added support for updating panes that do not yet have a stored version.
  * Improved validation for layout creation and append-window requests.
  * Improved environment loading, forking, deletion, and experiment tag handling.

### DIFF
--- a/py/tests/unit/pane_versions.py
+++ b/py/tests/unit/pane_versions.py
@@ -1,0 +1,257 @@
+#!/usr/bin/env python3
+
+# Copyright 2017-present, The Visdom Authors
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Every accepted ``/update`` advances the pane's broadcast sequence number.
+
+``version`` is the handshake behind incremental updates: the server broadcasts
+``window_update`` carrying the pane's new version, and the frontend applies the
+patch only when it reads ``pane.version + 1`` (``updateWindow`` in
+``js/main.js``). A version that fails to move is not a cosmetic problem -- the
+client discards the patch and re-requests the entire environment instead.
+
+The bump used to live in ``update_window()``, which ``UpdateHandler.update()``
+reaches for plot panes but returns before for text, image_history, plot_history
+and table panes, and which the embeddings route never calls at all. Those four
+types stayed at version 1 for the life of the pane while the server kept
+broadcasting updates against them, so every ``vis.text()`` append cost a full
+environment reload. These tests pin the counter to ``update_packet()`` -- the
+one place an accepted update becomes a broadcast -- and assert it for every
+pane type rather than for the single type that happened to work.
+"""
+
+import json
+
+import pytest
+
+from visdom.server.defaults import (
+    DEFAULT_MAX_IMAGE_HISTORY,
+    DEFAULT_MAX_OLD_CONTENT,
+    DEFAULT_MAX_PLOT_HISTORY,
+    DEFAULT_MAX_TEXT_LINES,
+)
+from visdom.server.handlers.web_handlers import UpdateHandler
+from visdom.utils.shared_utils import get_rand_id
+
+from testutils.fakes import FakeHandler
+
+pytestmark = pytest.mark.unit
+
+
+def _pane(ptype, **extra):
+    pane = {
+        "command": "window",
+        "version": 1,
+        "id": "win_{}".format(ptype),
+        "title": ptype,
+        "contentID": get_rand_id(),
+        "type": ptype,
+    }
+    pane.update(extra)
+    return pane
+
+
+def _plot_pane():
+    return _pane(
+        "plot",
+        content={
+            "data": [{"type": "scatter", "x": [1], "y": [1], "name": "a"}],
+            "layout": {},
+        },
+    )
+
+
+def _text_pane():
+    return _pane("text", content="line0")
+
+
+def _image_history_pane():
+    return _pane("image_history", content=[{"src": "img0"}], selected=0)
+
+
+def _plot_history_pane():
+    return _pane("plot_history", content=[{"data": [], "layout": {}}], selected=0)
+
+
+def _table_pane():
+    return _pane("table", content=[["a", "b"]], editable=True)
+
+
+def _embeddings_pane():
+    return _pane(
+        "embeddings",
+        content={"data": [[1, 2]], "selected": None, "has_previous": False},
+        old_content=[],
+    )
+
+
+#: ``(id, pane factory, update args)`` for every type ``/update`` accepts.
+#: ``table`` is here because the handler broadcasts for it even though the
+#: update itself is refused -- a broadcast the client cannot reconcile is the
+#: bug under test, whether or not the content changed.
+PANE_CASES = [
+    ("plot", _plot_pane, {"data": [{"type": "scatter", "x": [2], "y": [2]}]}),
+    ("text", _text_pane, {"data": [{"content": "line1"}]}),
+    (
+        "image_history",
+        _image_history_pane,
+        {"data": [{"type": "image_history", "content": {"src": "img1"}}]},
+    ),
+    (
+        "plot_history",
+        _plot_history_pane,
+        {"data": [{"type": "plot_history", "content": {"data": [], "layout": {}}}]},
+    ),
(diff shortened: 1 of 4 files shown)


Manik-Khajuria-5 (approved):
@tarone-saloni I went through the changes and the version bump placement looks correct Thanks for contribution

Manik-Khajuria-5 (changes requested):
@tarone-saloni Please resolve the conflicts and rebase the pr

Manik-Khajuria-5 (approved):

Jayantparashar10 (approved):

tarone-saloni:
Hi @Manik-Khajuria-5  and @Jayantparashar10 , It has been open for a while, and I have resolved the requested conflicts and pushed the updated changes.
Could you please let me know if there are any remaining issues or changes needed from my side before it can be merged?

Jayantparashar10:
> Hi @Manik-Khajuria-5 and @Jayantparashar10 , It has been open for a while, and I have resolved the requested conflicts and pushed the updated changes. Could you please let me know if there are any remaining issues or changes needed from my side before it can be merged?

@tarone-saloni There are no issues from my side. It will be merged soon. Thanks for your contribution!

rajnisht7 (approved):
The main fix looks correct moving the version bump into update_packet() (and the embeddings path) properly covers the pane types that used to stay stuck at version 1.

one edge case: if a pane was persisted before the version field existed, the client still stores it as undefined. On the first update the check becomes 2 == undefined + 1 (NaN), so it falls back to a full env reload once; after that reload the pane has a real version and incremental updates work normally, a tiny normalize to version: 1 in addPaneBatched would make even that first update incremental, this can be followed up.

tarone-saloni:
> The main fix looks correct moving the version bump into update_packet() (and the embeddings path) properly covers the pane types that used to stay stuck at version 1.
> 
> one edge case: if a pane was persisted before the version field existed, the client still stores it as undefined. On the first update the check becomes 2 == undefined + 1 (NaN), so it falls back to a full env reload once; after that reload the pane has a real version and incremental updates work normally, a tiny normalize to version: 1 in addPaneBatched would make even that first update incremental, this can be followed up.

Thanks @rajnisht7, good catch on the legacy panes. I'll send the addPaneBatched normalization as a small follow-up PR once this one is merged, since it only takes effect together with the server-side change here.

tarone-saloni:
Hi @mariobehling , could you please take a look at this PR when you get a chance? I'd appreciate a review. Thanks!

PR #1781, opened 2026-08-31
fix: skip unreadable experiment blobs instead of failing every search

adity1raut:
## Description

`ExperimentStore._read_metadata()` rebuilt an `Experiment` from the stored blob with no guard.
`Experiment.from_dict` indexes `data["env_id"]` directly, `Param`/`Metric`/`Tag.from_dict` index
`data["key"]`, and `Experiment.__post_init__` rejects any status outside `VALID_STATUSES` — so a
blob that was hand-edited or only half-written raised `KeyError`/`TypeError`/`ValueError` straight
out of the read.

That read is the one every bulk walk goes through: `_scan()` visits every environment via
`iter_experiments()`, so the exception did not hide a single run — it escaped the scan and took the
whole request with it. **A single malformed env file made `POST /experiments/search` return 500 for
every query, for every user, until somebody found and deleted the file.**

The storage layer below already refuses to behave that way: `load_env` and `list_envs` swallow
`OSError`/`ValueError` and skip the file. This change catches `(KeyError, TypeError, ValueError)`
around `from_dict`, logs the offending `env_id` at `WARNING` so the bad file stays findable, and
returns `None`. An unreadable blob now reads as "this env has no experiment", exactly like a file
`JSONStore` cannot parse.

| File | Change |
| --- | --- |
| `py/visdom/experiments/store.py` | The guard, the `logging` import, and a docstring paragraph on why this read must not raise |
| `py/tests/test_experiment_metadata_reads.py` | New `TestMalformedMetadataIsSkipped` (store layer) |
| `py/tests/integration/experiment_search.py` | Endpoint-level coverage on `TestSearchEndpoint` |

The write path (`_read`) is deliberately left raising. It touches one named environment at a time,
so an error there is visible to the caller who asked for that env and does not affect anyone else's
search.

## Motivation and Context

The storage layer underneath is explicitly defensive about exactly this case — `JSONStore.load_env`
and `JSONStore.list_envs` swallow `OSError`/`ValueError` and skip the file, precisely so that one
bad file cannot take down a listing. The experiment layer stacked on top inherited none of that,
which is what turned a single corrupt blob into a permanent, server-wide search outage with no way
to recover from the UI.

Reproduced before the change:

```
search with 1 healthy env                        -> 200
search with corrupt blob (status "cancelled")    -> 500
search with corrupt blob (missing env_id key)    -> 500
search with corrupt blob (params not a list)     -> 500
```

All three corrupt cases now return 200 and list the healthy runs.

## How Has This Been Tested?

New tests at both layers. Each fails on a clean checkout and passes with the fix.

**Unit** — `py/tests/test_experiment_metadata_reads.py::TestMalformedMetadataIsSkipped`

Eight malformed blob shapes, each asserted against `get_experiment`, `search`, `search_page` and
`iter_experiments`:

- status outside `VALID_STATUSES`
- `env_id` missing entirely
- `params` holding an object
- `params` holding a scalar
- a param that is not an object
- a param without a key
- a metric that is not an object
- a tag without a key

Plus: that the skip is logged with the env id, that the live `env_provider` branch is guarded too,
and that a healthy blob is still read in full.

**Integration** — `py/tests/integration/experiment_search.py::TestSearchEndpoint`

`POST /experiments/search` answers 200 and returns the readable runs with a corrupt env file sitting
on disk — both unfiltered and with a query plus `sort_by`.

**Verification run**

```
$ pytest py/tests -m "unit or integration"
2008 passed, 2 skipped
```

The two skips are the optional `xgboost` and built-frontend suites.

Reverting only `store.py` and re-running turns every one of the new cases red, including a real
`500 POST /experiments/search` in the endpoint test log — so the tests pin the bug, not just the fix.

`black py` reports no changes.

Not applicable — server-side behaviour only, no UI change.

## Types of changes

## Checklist:

- [ ] I adapted the version number under `py/visdom/VERSION` according to [Semantic Versioning](https://semver.org/) — `VERSION` is bumped in dedicated release commits in this repo, not per PR.

No API, message-protocol or auth change: the endpoint's request and response shapes are untouched,
and a valid blob reads exactly as it did before.

## Summary by Sourcery

Handle unreadable experiment metadata gracefully so affected environments are skipped while healthy experiments and recovery operations remain available.

Bug Fixes:
- Skip malformed experiment metadata during reads so one unreadable environment cannot fail searches, tag operations, or other bulk requests.
- Preserve recovery and deletion paths for environments containing unreadable experiment metadata.

Enhancements:
- Centralize defensive experiment-blob reconstruction with warning logs and consistent handling across persisted and live environments.

Tests:
- Add unit coverage for malformed metadata across reads, searches, iteration, logging, and recovery operations.
- Add integration coverage confirming experiment search and tag endpoints remain successful when corrupt environment blobs are present.

## Summary by CodeRabbit

- **Bug Fixes**
  - Malformed experiment metadata no longer causes searches, listings, or tag views to fail.
  - Search results continue to return valid experiments when corrupted records are present.
  - Tag operations safely skip invalid metadata and can repair affected experiments.
  - Updates to valid experiments remain available when another experiment has corrupted metadata.
  - Writes can replace unreadable experiment metadata.
  - Deleting an experiment with malformed metadata removes the record cleanly while preserving its environment and windows.

### DIFF
--- a/py/tests/integration/experiment_search.py
+++ b/py/tests/integration/experiment_search.py
@@ -16,6 +16,7 @@ the ``Visdom.search_experiments`` message shape with a mocked transport (no serv
 """
 
 import json
+import os
 import tempfile
 import unittest
 from unittest.mock import Mock, patch
@@ -461,6 +462,46 @@ class TestSearchEndpoint(tornado.testing.AsyncHTTPTestCase):
         for char in ("<", ">", "&"):
             self.assertNotIn(char, raw)
         self.assertEqual(json.loads(raw)["query"], query)
+
+    def write_corrupt_env(self, eid, blob):
+        """Write an env whose experiment blob the model cannot rebuild.
+
+        Written by hand rather than through the store, because the shapes that
+        break the read are exactly the ones :class:`Experiment` never produces:
+        a status it would have rejected, a field its writer always emits, a
+        list that is no longer a list.
+        """
+        with open(os.path.join(self._tmp_dir, eid + ".json"), "w") as fn:
+            fn.write(json.dumps({"jsons": {}, "reload": {}, "experiment": blob}))
+
+    def test_a_corrupt_blob_does_not_500_the_endpoint(self):
+        """One unreadable env is skipped; the readable ones still answer.
+
+        The scan reads every environment, so an exception escaping one blob is
+        not one missing run — it is the endpoint returning 500 for every query
+        until somebody finds the offending file and deletes it.
+        """
+        for label, blob in (
+            ("status outside VALID_STATUSES", {"env_id": "bad", "status": "cancelled"}),
+            ("env_id missing entirely", {"name": "bad"}),
+            ("params holding an object", {"env_id": "bad", "params": {"lr": 0.1}}),
+            ("a param without a key", {"env_id": "bad", "params": [{"value": 1}]}),
+        ):
+            with self.subTest(label):
+                self.write_corrupt_env("bad", blob)
+                body = self.search_ok({})
+                self.assertEqual(
+                    [e["env_id"] for e in body["experiments"]],
+                    ["run-c", "run-b", "run-a"],
+                )
+                self.assertEqual(body["total"], 3)
+
+    def test_a_corrupt_blob_does_not_break_a_filtered_search(self):
+        """Filtering and sorting read through the same guard, so they survive it."""
+        self.write_corrupt_env("bad", {"env_id": "bad", "status": "cancelled"})
+        body = self.search_ok({"query": "lr < 0.01 AND acc > 0.9", "sort_by": "lr"})
+        self.assertEqual([e["env_id"] for e in body["experiments"]], ["run-b"])
+        self.assertEqual(body["total"], 1)
 
     def test_search_sees_an_experiment_logged_over_http(self):
         """An experiment logged through /experiments/log is searchable at once."""
--- a/py/tests/integration/experiment_tags_handler.py
+++ b/py/tests/integration/experiment_tags_handler.py
@@ -81,6 +81,49 @@ class TestTagsEndpoint(VisdomHTTPTestCase):
             {"run-a": {"owner": "alice"}, "run-b": {"owner": "bob"}},
         )
 
+    def corrupt_env(self, eid, blob):
+        """Make ``eid`` a resident env whose experiment blob will not rebuild.
+
+        Put into live state rather than written to disk, because that is the
+        branch the handler parses itself: ``_experiment_from_env`` reads the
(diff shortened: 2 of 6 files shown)


Manik-Khajuria-5 (changes requested):
@adity1raut Please handle these cases for update and deletion as well, so corrupted experiment blobs don't make these operations fail.

adity1raut:
@Manik-Khajuria-5 Thanks — done, and you were right that the first pass was too narrow. Guarding only `_read_metadata` left a corrupt environment visible-but-frozen: search skipped it, but every operation that could have repaired it still raised, so the only recovery was shell access to the env directory.

I lifted the rebuild into a shared `experiment_from_blob()` helper and routed every call site through it.

**Update** — `ExperimentStore._read()` (which backs `log_experiment`, `log_metric`, `update_tags`, `finish_experiment` and `delete_experiment`), plus `update_tags`' `env_data` branch, which parses a blob of its own when the tags handler hands the store the server's live env. An unreadable blob now reads as "no experiment", which is already what these do for an env that never had metadata — so logging a param, a metric or a tag **replaces** the bad blob with a valid one instead of failing on it forever. `finish_experiment` and `compare` still refuse, but with the ordinary `KeyError` rather than a parse error out of the model.

**Deletion** — `delete_experiment()` now keys off the blob being *present* rather than on its rebuilding. The blob an operator most needs to delete is precisely the one that will not load, and deciding on the parsed experiment refused exactly that.

One more of the same shape while I was in there: `TagsHandler._experiment_from_env` is run over every resident env by `_experiment_map`, so a single corrupt blob returned 500 for the whole `GET /tags` map. Guarded too.

New tests, each of which fails without the change:

- a write-path class covering log/metric/tag repair, both `KeyError` paths, and deletion of four unreadable shapes (including a blob that is not an object at all) with the env's windows left intact;
- tags-endpoint coverage that `GET /tags` and a single-env tag read answer 200 with a corrupt blob in live state.

Also rebased onto current `dev`, so the merge conflict is gone.

`pytest py/tests -m "unit or integration"` → 2045 passed, 4 skipped (optional deps). `black py` reports no changes.

adity1raut:
@vedansh-5, can you merge this PR

adity1raut:
@Saksham-Sirohi  @vedansh-5  PTAL

Manik-Khajuria-5 (changes requested):
@adity1raut I checked the changes and found out  if experiment from blob fails, it could overwrite the existing experiment with an empty one and cause data loss.

Jayantparashar10 (commented):
@adity1raut Please address the reviews

Issue #1751, opened 2026-08-21
Bug: pie(), boxplot(), stem(), surf(), and contour() crash on size-1 inputs due to unconditional np.squeeze()


Pcmhacker-hero:
## Bug Description

Several plotting methods crash with `AssertionError` when given size-1 inputs (e.g. a single-slice pie chart or a single-row surface plot). The root cause is an unconditional `np.squeeze()` call that collapses valid dimensions before the `ndim` assertion runs.

## Steps to Reproduce

```python
import visdom
import numpy as np

viz = visdom.Visdom()

# 1. Single-slice pie chart (e.g. "100% complete")
viz.pie(X=[100])
# AssertionError: X should be one-dimensional

# 2. Single-point boxplot
viz.boxplot(X=[10.5])
# AssertionError: X should be one or two-dimensional

# 3. Single-point stem plot
viz.stem(X=[5.0])
# AssertionError: X should be one or two-dimensional

# 4. Single-row surface plot (1×5 grid)
viz.surf(X=np.ones((1, 5)))
# AssertionError: X should be two-dimensional

# 5. Single-column contour plot (5×1 grid)
viz.contour(X=np.ones((5, 1)))
# AssertionError: X should be two-dimensional
```

Multi-element inputs work fine for all methods (e.g. `pie([30, 70])`, `surf(np.ones((5, 5)))`).

## Expected Behavior

- `pie(X=[100])` should render a single-slice pie chart.
- `boxplot(X=[10.5])` should render a single-observation boxplot.
- `stem(X=[5.0])` should render a single stem.
- `surf(X=np.ones((1, 5)))` and `contour(X=np.ones((5, 1)))` should render the 2D matrix as-is (valid Plotly `z` data).

## Root Cause

In `py/visdom/__init__.py`, the affected methods call `X = np.squeeze(X)` **unconditionally** before asserting dimensionality:

| Method | Line | What happens |
|---|---|---|
| `pie()` | L3946 | `np.squeeze([100])` → shape `(1,)` → `()` (ndim=0) → fails `assert X.ndim == 1` |
| `boxplot()` | L3621 | `np.squeeze([10.5])` → shape `(1,)` → `()` (ndim=0) → fails `assert X.ndim == 1 or X.ndim == 2` |
| `stem()` | L3857 | Same as boxplot |
| `_surface()` | L3672 | `np.squeeze(np.ones((1, 5)))` → shape `(5,)` (ndim=1) → fails `assert X.ndim == 2` |

[...]

PR #1788, opened 2026-09-02
Fix unconditional np.squeeze() crashing bar/histogram/sunburst on size-1 inputs

### LINKED ISSUE #1787: Bug: bar(), histogram(), and sunburst() crash on size-1 inputs due to unconditional np.squeeze()
**Bug Description**
Several plotting methods crash with AssertionError when given size-1 inputs. The root cause is an unconditional np.squeeze() call that collapses valid dimensions before the ndim assertion runs. This is the same bug class already partially fixed in #1751/#1762 (pie, boxplot, stem, _surface) — bar(), histogram(), and sunburst() have the identical unguarded pattern but weren't covered by that fix.

**Reproduction Steps**
Enter steps to reproduce the behavior:
1. Start a visdom server: python -m visdom.server
2. In a Python terminal:
`import visdom
viz = visdom.Visdom()
viz.bar(X=[5])
#AssertionError: X should be one or two-dimensional
viz.histogram(X=[42])
 #AssertionError: X should be one-dimensional
viz.sunburst(labels=["A"], parents=[""], values=[7])
#AssertionError: values should be one-dimensional`

3. Observe each call crashes with an AssertionError instead of rendering.

**Expected behavior**

- bar(X=[5]) should render a single bar.
- histogram(X=[42]) should render a single-bin histogram.
- sunburst(...., values=[7]) single-segment sunburst with a value shown

**Screenshots**
N/A — this is a server-side AssertionError, no UI is reached.

**Client logs:**
N/A — crash happens server-side before reaching the frontend.

**Server logs:**
`AssertionError: X should be one or two-dimensional
AssertionError: X should be one-dimensional
AssertionError: values should be one-dimensional`

**Additional context**

Confirmed on current dev branch, py/visdom/__init__.py:

- bar() — L3646: X = np.squeeze(X), unguarded
- histogram() — L3713: X = np.squeeze(X), unguarded
- sunburst() values param — L4155: values = np.squeeze(values), unguarded. Note that labels/parents in this same function are already correctly guarded with np.atleast_1d() — only values still has the bug.

violin() (L4740) already uses the correct guarded pattern:
`X = np.asarray(X)
if X.ndim > 2:
    X = np.squeeze(X)`

munazzaghazali7-hash:
## Description
`bar()`, `histogram()`, and `sunburst()` called `np.squeeze()` unconditionally on their inputs. For size-1 arrays like `[5]` (shape `(1,)`), this collapsed them to 0-d arrays *before* the `ndim` assertions ran, causing those assertions to fail even though the input was valid.

Updated based on review feedback to preserve `np.squeeze()`'s original behavior for 2D row/column vectors (e.g. `[[1,2,3,4,5]]` still collapses to `(5,))`, while still preventing the size-1 crash.

```python
X = np.atleast_1d(np.squeeze(X))
```

`sunburst()`'s `labels`/`parents` params were already correctly guarded with `np.atleast_1d()` — only `values` had the bug.

## Motivation and Context
Fixes #1787 . Size-1 inputs are a legitimate, common case (e.g. a single bar, single-bin histogram, or single-segment sunburst), and they crashed with a server-side `AssertionError` instead of rendering. This is the same bug class already partially fixed in #1751 / #1762 for `pie()`, `boxplot()`, `stem()`, and `_surface()` — `bar()`, `histogram()`, and `sunburst()` had the identical unguarded pattern but weren't covered by that fix.

## How Has This Been Tested?
Manually reproduced the crash on the original code with the exact repro steps from #1787:
```python
viz.bar(X=[5])          # AssertionError: X should be one or two-dimensional
viz.histogram(X=[42])   # AssertionError: X should be one-dimensional
viz.sunburst(labels=["A"], parents=[""], values=[7])  # AssertionError: values should be one-dimensional
```
After the fix, all three calls succeed and render correctly in the UI (single bar, single-bin histogram, single-segment sunburst).

Also ran the full existing test suite (`pytest py/tests/`) to confirm no regressions: **2079 passed, 3 skipped** (skips are unrelated, due to missing optional deps `xgboost`/`bs4`/`lxml`).

Screenshot 2026-09-02 at 6 13 43 PM

## Types of changes

## Checklist:

## Summary by Sourcery

Prevent size-one inputs from collapsing to zero-dimensional arrays in bar, histogram, and sunburst plots.

Bug Fixes:
- Allow bar, histogram, and sunburst to render valid size-one inputs instead of failing dimensionality assertions.

Enhancements:
- Preserve existing squeezing behavior for multi-element row and column vectors while normalizing scalar results to one-dimensional arrays.

Tests:
- Add regression coverage for single-value bar, histogram, and sunburst inputs, including bar X/Y handling and preserved row-vector squeezing.

## Summary by CodeRabbit

* **New Features**
  * Added an option to skip preflight checks when updating image-history and scatter visualizations.
  * Added support for configuring 3D plot aspect modes and positive aspect ratios across applicable visualizations.
* **Bug Fixes**
  * Improved average precision calculations for precision-recall curves with tied recall values.
  * Fixed bar, histogram, and sunburst plots so scalar and single-value inputs render correctly.
  * Preserved expected row-vector behavior by collapsing them into a single plot trace.

### DIFF
--- a/py/tests/unit/client_payloads_graph.py
+++ b/py/tests/unit/client_payloads_graph.py
@@ -74,6 +74,18 @@ def test_sunburst_sends_values_when_given(capture_send):
     assert trace(sent)["values"] == [1, 2]
 
 
+def test_sunburst_size_one_values_renders(capture_send):
+    """Regression test for #1787: a single-value `values` array used to
+    collapse to a 0-d array via an unconditional np.squeeze() and fail the
+    ndim assert, even though labels/parents were already correctly guarded."""
+    sent = capture_send(
+        lambda v: v.sunburst(
+            labels=np.array(["a"]), parents=np.array([""]), values=np.array([7])
+        )
+    )
+    assert trace(sent)["values"] == [7]
+
+
 def test_sunburst_reads_its_styling_from_opts(capture_send):
     """The four style opts are spelled differently in the trace than in opts."""
     sent = capture_send(
--- a/py/tests/unit/plots.py
+++ b/py/tests/unit/plots.py
@@ -428,6 +428,29 @@ def test_bar_legend_with_rownames_on_1d_raises(offline_client):
         offline_client.bar(np.array([1.0, 2.0, 3.0]), opts=opts)
 
 
+def test_bar_size_one_x_renders(capture_send):
+    """Regression test for #1787: a single-value X used to collapse to a
+    0-d array via an unconditional np.squeeze() and fail the ndim assert."""
+    sent = capture_send(lambda v: v.bar(np.array([5.0])))
+    assert sent["payload"]["data"][0]["y"] == [5.0]
+
+
+def test_bar_size_one_y_renders(capture_send):
+    """Regression test for #1787: a single-value Y hit the same unconditional
+    np.squeeze() bug as X, one call deeper in bar()."""
+    sent = capture_send(lambda v: v.bar(np.array([5.0]), Y=np.array([10.0])))
+    assert sent["payload"]["data"][0]["x"] == [10.0]
+
+
+def test_bar_2d_row_vector_still_squeezes(capture_send):
+    """A 2D row vector must still collapse to a single 1D trace, matching
+    np.squeeze()'s original behavior -- only the size-1 case should be
+    guarded, not squeezing in general."""
+    sent = capture_send(lambda v: v.bar(np.array([[1.0, 2.0, 3.0, 4.0, 5.0]])))
+    assert len(sent["payload"]["data"]) == 1
+    assert sent["payload"]["data"][0]["y"] == [1.0, 2.0, 3.0, 4.0, 5.0]
+
+
 def test_bar_stacked_sets_barmode(capture_send):
     """stacked=True stacks the columns instead of grouping them."""
     X = np.array([[1.0, 2.0], [3.0, 4.0]])
@@ -474,6 +497,13 @@ def test_histogram_bin_edges_span_data_range(capture_send):
     x = sent["payload"]["data"][0]["x"]
     assert x[0] == 0.0
     assert x[-1] == 99.0
+
+
+def test_histogram_size_one_x_renders(capture_send):
+    """Regression test for #1787: a single-value X used to collapse to a
+    0-d array via an unconditional np.squeeze() and fail the ndim assert."""
+    sent = capture_send(lambda v: v.histogram(np.array([42.0])))
+    assert sum(sent["payload"]["data"][0]["y"]) == 1
 
 
 # ------------------------------------------------------------------- boxplot ----
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -3643,9 +3643,8 @@ class Visdom(object):
         - `opts.stacked` : stack multiple columns in `X`
             - `opts.legend`  : `list` containing legend labels
         """
-        X = np.asarray(X)
-        if X.ndim > 2:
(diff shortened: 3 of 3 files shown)


munazzaghazali7-hash:
@Manik-Khajuria-5  , @tonypzy , @rajnisht7 and @vedansh-5  I’ve addressed the issue in this PR and would really appreciate it if you could take a look when you get a chance. Thank you! 🙌

Manik-Khajuria-5 (changes requested):
@munazzaghazali7-hash Please preserve np.squeeze() behavior for 2 D row vectors?

munazzaghazali7-hash:
> @munazzaghazali7-hash Please preserve np.squeeze() behavior for 2 D row vectors?

Good catch — thanks for flagging this. I switched to X = np.atleast_1d(np.squeeze(X)) instead of the ndim > N guard, which preserves the original squeeze behavior for 2D row/column vectors (e.g. [[1,2,3,4,5]] still collapses to (5,) as before) while also fixing the size-1 crash from #1787 

Verified both cases manually:
`viz.bar(X=[5])                  # size-1 no longer crashes
viz.bar(X=[[1, 2, 3, 4, 5]])     # 2D row vector still collapses to a normal 5-bar chart`

Screenshot 2026-09-03 at 8 50 49 PM


Also re-ran the full test suite (pytest py/tests/) still passing, no regressions. Pushed the update to this branch.

Manik-Khajuria-5 (approved):

Jayantparashar10 (approved):

rajnisht7 (commented):
@munazzaghazali7-hash, the PR looks good just a few things
- update description
- rebase the branch (7 commits behind)
- add a regression test

munazzaghazali7-hash:
Hi, @rajnisht7 Thanks for the review! I've addressed all three points:

Description — updated to reflect the current fix. It now uses `X = np.atleast_1d(np.squeeze(X)`) instead of the earlier if `X.ndim > 2`: guard, since that preserves `np.squeeze()`'s original behavior for 2D row/column vectors (per your earlier comment) while still fixing the size-1 crash.

Rebase - rebased onto the latest `dev`, force-pushed with `--force-with-lease`.

Regression tests - added 5 new tests:

`test_bar_size_one_x_renders`
`test_bar_size_one_y_renders`
`test_bar_2d_row_vector_still_squeezes `(covers the 2D row vector case specifically)
`test_histogram_size_one_x_renders`
`test_sunburst_size_one_values_renders`

I verified these against the original buggy code first, all 4 size-1 tests failed there as expected (confirming they actually catch the bug), and the 2D-row-vector test passed. After applying the fix, all 5 pass, and the full suite is still green: 2056 passed, 3 skipped.

rajnisht7 (approved):
The PR looks good, but the same pattern(that is unguarded np.squeeze + ndim assert pattern) appears in
- pie()
- boxplot()
- stem()
- histogram2d()
- _surface()

Needs to follow up

munazzaghazali7-hash:
@rajnisht7  Alright, I'll open a separate issue and PR to fix these, keeping this one scoped to bar/histogram/sunburst as originally planned.

munazzaghazali7-hash:
Hi @vedansh-5, @mariobehling — both this PR and #1821 (the follow-up) have been approved by reviewers but are blocked on two things:

Needs 2 approving reviews from someone with write access all current approvals are read only
CI workflows are awaiting maintainer approval to run (CodeQL is pending on both)

Could you take a look when you get a chance? Happy to make any changes if needed. Thanks!

PR #1789, opened 2026-09-02
Release v0.3.0

vedansh-5:
Release branch for **v0.3.0**, cut from `dev` at `33b88244` (31 July) with the
experiment-layer and Playwright completion work cherry-picked on top.

## What's in it

Everything on `dev` up to `33b88244`, plus 8 commits:

| PR | |
| --- | --- |
| #1602 | Experiment query parser (injection-safe filter language) |
| #1604 | Experiment search endpoint + `vis.search_experiments()` |
| #1605 | Compare experiments field by field + `vis.compare_experiments()` |
| #1609 | `/experiments/suggest` stub endpoint + API docs |
| #1561 | Migrate Cypress image.js E2E tests to Playwright |
| #1691 | Switch E2E tests to Playwright |
| #1705 | Remove retired Cypress |
| — | Align prettier pin with lockfile; bump `VERSION` to 0.3.0 |

The experiment PRs complete the Layer 2 work that `#1595` started. The Playwright
PRs finish the Cypress migration, so the release notes' claim that the E2E suite
has moved to Playwright is accurate — Cypress is fully removed and `npm test`
runs Playwright.

## Verification

- `384 passed` on the full pytest suite
- `import visdom` clean; `experiment`, `log_metrics`, `finish_experiment`,
  `search_experiments`, `compare_experiments` all present
- All four `/experiments/*` routes registered in `app.py`
- 0 Cypress files remain; 13 Playwright specs
- `py/visdom/VERSION` = `0.3.0`
- Every PR and commit hash cited in the release notes resolves against this branch

## Notes

- `/experiments/suggest` is server-side only in this release; there is no
  `vis.suggest_experiments()` client method.
- Two conflicts came up during the cherry-picks and were resolved: a docstring
  collision on `_experiment_request` (#1605), and the Cypress removal touching
  `package.json`/`yarn.lock` (#1705). `pixelmatch` and `pngjs` were dropped along
  with Cypress — they were only used by `cypress.config.js`, not by the
  Playwright suite, which has its own screenshot comparison.

## After merge

Tag `v0.3.0` on the squashed commit and publish with `RELEASE_NOTES_v0.3.0.md`.
The existing `v0.3.0` tag points at `c9117e44`, which is no longer on `dev` after
the history rewrite, so it needs to be moved or deleted first.

Issue #1739, opened 2026-08-18
Comparing environments crashes with IndexError when a shared plot window has empty data


piyush182004:
## Bug

`compare_envs` in `py/visdom/utils/server_utils.py` crashes with
`IndexError: list index out of range` when comparing two or more
environments that share a plot window title, if that window's data list
is empty.

Lines 341 and 353 do:
    if "name" not in destWidJson["content"]["data"][0]:

This indexes `data[0]` without first checking the list is non-empty. A
plot window that exists but has no data plotted into it yet (a normal,
reachable state) triggers this immediately.

## Repro
1. Create two environments, each with a plot window sharing the same
   title, with no data plotted (or data plotted then removed).
2. Select both environments and click Compare.
3. Server returns HTTP 500; traceback shows the IndexError above.

## Fix
Guard both occurrences with a check that `data` is non-empty before
indexing `[0]`.

FIX: #1738

PR #1796, opened 2026-09-03
fix: guard contributing plot data in compare_envs

### LINKED ISSUE #1795: Bug: compare_envs crashes or half-merges when a contributing plot pane has no data
## Bug

`compare_envs` in `py/visdom/utils/server_utils.py` guards the *base* env's plot data before indexing it, but not the *contributing* env's. The plot branch reads:

```python
base_data = destWidJson["content"].get("data") or []
if not base_data or "name" not in base_data[0]:
    continue                                  # base side: guarded
...
else:
    destWidJson["has_compare"] = True
    for _dataIdx, data in enumerate(win["content"]["data"]):   # contributor side: not guarded
```

This is the same class of problem as #1739, which was fixed for the base side only. Two distinct failures remain on the contributor side.

### 1. `KeyError: 'data'` when a contributing pane has no data key

A plot pane whose `content` exists but carries no `"data"` key makes `win["content"]["data"]` raise. The base side already uses `.get("data")`, which implies this shape is considered reachable.

Reproduced as a unit test against `compare_envs`:

```
KeyError: 'data'
  py/visdom/utils/server_utils.py:615
```

### 2. A contributor with no traces produces a one-sided "comparison"

`has_compare` is set to `True` *before* iterating the contributor's traces. If that list is empty, the loop body never runs, so nothing is appended and nothing clears the flag. The pane then survives the `has_compare` filter while containing only the base env's data.

Reproduced with env `a` holding a `loss` plot with one trace and env `b` holding a same-titled plot with no traces:

```
titles=['loss', 'compare_legend']  win_present=True  traces=['a_train']
```

The pane is presented as a comparison of two envs while showing one env's trace.

This also contradicts the behaviour the suite already pins for the analogous case. `test_a_later_env_with_an_unnamed_trace_clears_has_compare` documents the intended rule as *"One malformed contributor withdraws the whole pane, rather than half-merging"* — an empty contributor should withdraw the pane the same way, but currently does not.

## Expected

- A contributing pane without a `"data"` key is skipped rather than crashing the comparison.
- A contributing pane with no traces withdraws the pane, consistent with the unnamed-trace rule.

## Fix

Read the contributor's data with `.get("data") or []` and clear `has_compare` when it is empty, mirroring the base-side guard. I have a fix and regression tests ready and will open a PR.

thisisanubhav:
## Description

`compare_envs` guards the base env's plot data before indexing it, but indexes the contributing env's directly:

```
base_data = destWidJson["content"].get("data") or []
if not base_data or "name" not in base_data[0]:
    continue                                                   # base: guarded
...
else:
    destWidJson["has_compare"] = True
    for _dataIdx, data in enumerate(win["content"]["data"]):    # contributor: not guarded
```

That leaves two distinct failures on the contributor side:

1. **`KeyError: 'data'`** — a contributing pane whose `content` carries no `"data"` key crashes the whole comparison. The base side already uses `.get("data")`, so this shape is treated as reachable there.
2. **A one-sided "comparison"** — `has_compare` is set to `True` *before* the loop. An empty trace list means the body never runs, so nothing is appended and nothing clears the flag; the pane then survives the `has_compare` filter while showing only the base env's traces.

This change reads the contributing data with `.get("data") or []` and withdraws the pane when it is empty, mirroring the base-side guard.

## Motivation and Context

Fixes #1795.

#1739 fixed this class of bug on the base side only, so the symmetric contributor-side hole remained. Case 1 is a user-visible HTTP 500 when comparing environments. Case 2 is quieter and arguably worse: no error, just a pane presented as a comparison of several environments while containing one environment's data.

Case 2 also contradicts a rule the test suite already pins. `test_a_later_env_with_an_unnamed_trace_clears_has_compare` documents the intended behaviour as *"One malformed contributor withdraws the whole pane, rather than half-merging"* — an empty contributor should withdraw the pane the same way, and did not.

## How Has This Been Tested?

Environment: Python 3.12, `pytest -m unit`, installed from source with `pip install -e .`.

Both failures were first reproduced as tests against `compare_envs` before any fix was written:

```
KeyError: 'data'                                          # missing data key
titles=['loss', 'compare_legend'] traces=['a_train']      # empty contributor, pane kept
```

Three regression tests were added to `py/tests/unit/compare_envs.py`:

- `test_a_later_env_with_no_traces_clears_has_compare` — an empty contributor withdraws the pane
- `test_a_later_env_without_a_data_key_is_skipped` — a missing `data` key is skipped, not a crash
- `test_an_empty_contributor_does_not_block_a_later_valid_one` — a valid contributor after an empty one still compares, giving `["a_loss", "c_loss"]`

Results:

- `py/tests/unit/compare_envs.py` → **50 passed**
- Full `-m unit` marker → **101 passed**, excluding `keras_logger`/`sklearn_logger`/`pytorch_logger`, whose optional ML dependencies are not installed locally

The tests were also mutation-tested rather than simply observed passing: reverting the guard fails exactly the two tests describing the bugs — one with `KeyError: 'data'`, one on the half-merge — and restoring it returns to 50 passing. The third test passes both before and after, and is there to catch this fix over-correcting into dropping panes that should still compare.

Not applicable — server-side merge logic with no UI change.

## Types of changes

## Checklist:

<sub>Note on the version checkbox: left unticked deliberately, since #1789 is preparing the v0.3.0 release and bumping `VERSION` here would conflict with it. Happy to bump if maintainers prefer.</sub>

## Note on scope

I kept the existing semantics for a malformed contributor rather than inventing new ones: an empty contributor now behaves exactly like an unnamed-trace contributor. Both remain order-dependent in the same way — a later valid env can set `has_compare` back to `True` — which the third test documents rather than changes. Happy to tighten that if maintainers would prefer a stricter rule, but that felt like a behavioural decision beyond this fix.

## Summary by Sourcery

Harden environment plot comparisons against malformed trace data while preserving valid multi-environment comparisons.

Bug Fixes:
- Prevent environment comparisons from crashing or producing one-sided panes when contributor plot data is missing, empty, non-list, or malformed.
- Ensure malformed base and contributor traces are skipped without partially renaming or merging plot data.

Enhancements:
- Validate all plot traces before modifying comparison panes, preserving valid comparisons while avoiding partial merges.

Tests:
- Add regression coverage for missing, empty, non-list, and malformed base or contributor plot data, including valid environments before or after invalid contributors.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved environment comparisons when plot panes contain missing, empty, or malformed trace data.
  - Skipped panes with invalid or partly named trace entries without causing errors or partially renaming traces.
  - Prevented incomplete base-pane traces from producing unreliable comparisons.
  - Valid plot data continues to merge correctly regardless of pane order.
  - Existing comparisons are preserved when later panes lack usable plot data.

### DIFF
--- a/py/tests/unit/compare_envs.py
+++ b/py/tests/unit/compare_envs.py
@@ -270,6 +270,38 @@ def test_a_later_env_with_an_unnamed_trace_clears_has_compare(fake_socket, store
     assert _by_title(fake_socket, "loss") is None
 
 
+def test_a_later_env_with_no_traces_clears_has_compare(fake_socket, store):
+    """An empty contributor withdraws the pane instead of half-merging it."""
+    state = {
+        "a": _env(_plot_pane("w1", "loss")),
+        "b": _env(_plot_pane("w2", "loss", names=())),
+    }
+    compare_envs(state, ["a", "b"], fake_socket, store)
+    assert _by_title(fake_socket, "loss") is None
+
+
+def test_a_later_env_without_a_data_key_is_skipped(fake_socket, store):
+    """A contributor whose content carries no data at all is not a crash."""
+    pane = _plot_pane("w2", "loss")
+    del pane["content"]["data"]
+    state = {"a": _env(_plot_pane("w1", "loss")), "b": _env(pane)}
+    compare_envs(state, ["a", "b"], fake_socket, store)
+    assert _by_title(fake_socket, "loss") is None
+
+
+def test_an_empty_contributor_does_not_block_a_later_valid_one(fake_socket, store):
+    """A pane still compares when a good contributor follows an empty one."""
+    state = {
+        "a": _env(_plot_pane("w1", "loss")),
+        "b": _env(_plot_pane("w2", "loss", names=())),
+        "c": _env(_plot_pane("w3", "loss")),
+    }
+    compare_envs(state, ["a", "b", "c"], fake_socket, store)
+    win = _by_title(fake_socket, "loss")
+    assert win is not None
+    assert _trace_names(win) == ["a_loss", "c_loss"]
+
+
 def test_an_empty_title_is_never_merged(fake_socket, store):
     state = {"a": _env(_plot_pane("w1", "")), "b": _env(_plot_pane("w2", ""))}
     compare_envs(state, ["a", "b"], fake_socket, store)
--- a/py/tests/unit/compare_envs.py
+++ b/py/tests/unit/compare_envs.py
@@ -368,6 +368,27 @@ def test_a_contributor_whose_data_is_not_a_list_is_skipped(fake_socket, store):
     assert _by_title(fake_socket, "loss") is None
 
 
+def test_a_base_pane_with_a_none_trace_after_a_valid_one_is_skipped(fake_socket, store):
+    """Validation covers every base trace, not just the first."""
+    pane = _plot_pane("w1", "loss")
+    pane["content"]["data"] = [{"type": "scatter", "name": "loss"}, None]
+    state = {"a": _env(pane), "b": _env(_plot_pane("w2", "loss"))}
+    compare_envs(state, ["a", "b"], fake_socket, store)
+    assert _by_title(fake_socket, "loss") is None
+
+
+def test_a_partly_named_base_pane_is_never_half_renamed(fake_socket, store):
+    """A base pane with one unnamed trace withdraws instead of leaking it."""
+    pane = _plot_pane("w1", "loss")
+    pane["content"]["data"] = [
+        {"type": "scatter", "name": "loss"},
+        {"type": "scatter"},
+    ]
+    state = {"a": _env(pane), "b": _env(_plot_pane("w2", "loss"))}
+    compare_envs(state, ["a", "b"], fake_socket, store)
+    assert _by_title(fake_socket, "loss") is None
+
+
 def test_an_empty_title_is_never_merged(fake_socket, store):
     state = {"a": _env(_plot_pane("w1", "")), "b": _env(_plot_pane("w2", ""))}
     compare_envs(state, ["a", "b"], fake_socket, store)
--- a/py/tests/unit/compare_envs.py
+++ b/py/tests/unit/compare_envs.py
@@ -315,6 +315,32 @@ def test_an_empty_contributor_does_not_block_a_later_valid_one(fake_socket, stor
     assert _trace_names(win) == ["a_loss", "c_loss"]
 
 
+def test_a_malformed_contributor_does_not_undo_an_earlier_valid_one(fake_socket, store):
+    """An unnamed trace withdraws its own env, not comparisons already made."""
+    state = {
+        "a": _env(_plot_pane("w1", "loss")),
+        "b": _env(_plot_pane("w2", "loss")),
+        "c": _env(_plot_pane("w3", "loss", names=(None,))),
+    }
+    compare_envs(state, ["a", "b", "c"], fake_socket, store)
+    win = _by_title(fake_socket, "loss")
+    assert win is not None
+    assert _trace_names(win) == ["a_loss", "b_loss"]
+
+
+def test_a_partly_named_contributor_merges_none_of_its_traces(fake_socket, store):
+    """A contributor is all-or-nothing: one unnamed trace drops the whole env."""
+    state = {
+        "a": _env(_plot_pane("w1", "loss")),
+        "b": _env(_plot_pane("w2", "loss", names=("good", None))),
+        "c": _env(_plot_pane("w3", "loss")),
+    }
+    compare_envs(state, ["a", "b", "c"], fake_socket, store)
+    win = _by_title(fake_socket, "loss")
+    assert win is not None
+    assert _trace_names(win) == ["a_loss", "c_loss"]
+
+
 def test_an_empty_title_is_never_merged(fake_socket, store):
     state = {"a": _env(_plot_pane("w1", "")), "b": _env(_plot_pane("w2", ""))}
     compare_envs(state, ["a", "b"], fake_socket, store)
--- a/py/tests/unit/compare_envs.py
+++ b/py/tests/unit/compare_envs.py
@@ -341,6 +341,33 @@ def test_a_partly_named_contributor_merges_none_of_its_traces(fake_socket, store
     assert _trace_names(win) == ["a_loss", "c_loss"]
 
 
+def test_a_contributor_with_a_none_trace_is_skipped(fake_socket, store):
+    """A trace list holding None is malformed input, not a crash."""
+    pane = _plot_pane("w2", "loss")
+    pane["content"]["data"] = [None]
+    state = {"a": _env(_plot_pane("w1", "loss")), "b": _env(pane)}
+    compare_envs(state, ["a", "b"], fake_socket, store)
+    assert _by_title(fake_socket, "loss") is None
+
+
+def test_a_base_pane_with_a_none_trace_is_skipped(fake_socket, store):
+    """The same malformed trace list in the base env is also survivable."""
+    pane = _plot_pane("w1", "loss")
+    pane["content"]["data"] = [None]
+    state = {"a": _env(pane), "b": _env(_plot_pane("w2", "loss"))}
+    compare_envs(state, ["a", "b"], fake_socket, store)
+    assert _by_title(fake_socket, "loss") is None
+
+
+def test_a_contributor_whose_data_is_not_a_list_is_skipped(fake_socket, store):
+    """Content that is not a trace list at all is skipped rather than iterated."""
+    pane = _plot_pane("w2", "loss")
+    pane["content"]["data"] = {"name": "loss"}
+    state = {"a": _env(_plot_pane("w1", "loss")), "b": _env(pane)}
+    compare_envs(state, ["a", "b"], fake_socket, store)
+    assert _by_title(fake_socket, "loss") is None
+
+
 def test_an_empty_title_is_never_merged(fake_socket, store):
     state = {"a": _env(_plot_pane("w1", "")), "b": _env(_plot_pane("w2", ""))}
     compare_envs(state, ["a", "b"], fake_socket, store)
--- a/py/tests/unit/compare_envs.py
+++ b/py/tests/unit/compare_envs.py
@@ -289,6 +289,19 @@ def test_a_later_env_without_a_data_key_is_skipped(fake_socket, store):
     assert _by_title(fake_socket, "loss") is None
 
 
+def test_an_empty_contributor_does_not_undo_an_earlier_valid_one(fake_socket, store):
+    """A comparison already established survives a later empty contributor."""
+    state = {
+        "a": _env(_plot_pane("w1", "loss")),
+        "b": _env(_plot_pane("w2", "loss")),
+        "c": _env(_plot_pane("w3", "loss", names=())),
+    }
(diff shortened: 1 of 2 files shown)


thisisanubhav:
@sourcery-ai review

Re-requesting a review: the earlier attempt hit `unable to review this pull request ... this pull request has 307 [files]`, which was the stale diff from when I had mistakenly based this on `main` instead of `dev`. The base has since been corrected and the real change is 2 files, +40/-1.

thisisanubhav:
Good catch — this was a real regression I introduced, and it's fixed in 7d0a769.

I verified it rather than taking it on trust, comparing the same scenario (envs `a` and `b` both contributing, `c` empty) against the parent commit:

| | result |
| --- | --- |
| before my change | `traces=['a_loss', 'b_loss']` — comparison survives |
| after my change | `win_present=False` — **comparison deleted** |

Clearing `has_compare` on an empty contributor was wrong: it discarded work an earlier env had already contributed, and the cleanup pass then dropped a perfectly valid pane.

The fix is to skip an empty contributor without touching the flag at all. If an earlier env already contributed, that comparison stands; if none has, the flag is still `False` from the base env, so the pane drops out exactly as intended. That keeps the original bug fixed while restoring the behaviour I broke.

Added `test_an_empty_contributor_does_not_undo_an_earlier_valid_one` covering the ordering you described.

**Verification** — `py/tests/unit/compare_envs.py` → **51 passed**, full `-m unit` → **102 passed** (excluding the three logger modules whose optional ML deps aren't installed locally).

Mutation-tested in both directions:

- re-introducing the `has_compare = False` reset fails exactly the new ordering test
- reverting to the original pre-fix code fails exactly the two tests for the bugs this PR set out to fix (`KeyError: 'data'` and the half-merge)

So the three orderings are now pinned independently: empty-only drops the pane, empty-then-valid still compares, and valid-then-empty keeps the comparison.

I rebased onto the `dev` merge that had landed on the branch in the meantime; no conflicts, and `server_utils.py` was untouched by it.

Manik-Khajuria-5 (changes requested):
@thisisanubhav Looks good overall.  Please handle the malformed trace case

thisisanubhav:
@Manik-Khajuria-5 thanks — handled in aee97f5.

I read "the malformed trace case" as the unnamed-trace path having the same two problems the empty-contributor path had. If that's not what you meant, tell me and I'll redo it.

I checked the current behaviour first rather than guessing:

| scenario | before |
| --- | --- |
| envs `a`, `b` valid, `c` has an unnamed trace | `win=False` — the whole `a`/`b` comparison **deleted** |
| contributor `b` with traces `("good", None)` | `['a_loss', 'b_good', 'c_loss']` — **half-merged** |

Both mirror what CodeRabbit caught above. The first destroys work earlier envs already contributed, because a malformed contributor cleared `has_compare`. The second is the "half-merging" the suite's own docstring says the code avoids: `b_good` was appended before the loop hit the unnamed trace and broke, and a later valid env then set the flag back to `True`, so the partial merge survived into the output.

The fix makes a contributor all-or-nothing: validate every trace before appending any, and skip a malformed contributor without touching `has_compare`. One bad env now withdraws only itself.

After:

- `a`, `b` valid, `c` unnamed → `['a_loss', 'b_loss']`
- `("good", None)` contributor → `['a_loss', 'c_loss']`, nothing half-merged

The existing `test_a_later_env_with_an_unnamed_trace_clears_has_compare` still passes unchanged: with only `a` and a malformed `b`, nothing ever set the flag, so the pane still drops.

**Verification** — `compare_envs.py` → **53 passed**, full `-m unit` → **111 passed** (excluding the three logger modules whose optional ML deps aren't installed locally). Mutation-tested: restoring the old append-as-you-go loop fails exactly the two new tests.

thisisanubhav:
Fixed in 0bcf688.

Verified the crash first, and it is real — a trace list holding `None` makes `"name" not in trace` raise:

```
TypeError: argument of type 'NoneType' is not iterable
```

Worth being precise about provenance: this one **predates the PR**. I ran the same case against the pre-PR code and it raises identically, so it is not a regression from my change. It is the same failure mode the rest of this PR addresses — malformed contributor data taking down the whole comparison — so it seemed wrong to leave it behind now that it is known.

Added `_is_named_trace`, which checks the mapping shape before testing for the key, and required the data to be a list on both sides. The base side had the same exposure via `"name" not in base_data[0]`, so it is guarded too rather than fixing only the half the review pointed at.

Three tests added: a `None` trace in a contributor, a `None` trace in the base pane, and content whose `data` is a dict rather than a list.

**Verification** — `compare_envs.py` → **56 passed**, full `-m unit` → **114 passed** (excluding the three logger modules whose optional ML deps aren't installed locally). Mutation-tested: dropping the `isinstance` check from `_is_named_trace` fails exactly the two `None`-trace tests.

Note for maintainers: CI on this branch has gone back to `action_required` after the last few pushes, so Python Tests, CodeQL and the Playwright jobs have not run against the current head. They were all green on an earlier commit. A re-approval when convenient would confirm the final state.

thisisanubhav:
Fixed in f1e2a1c. Verified both halves of the finding first:

| base trace list | before |
| --- | --- |
| `[{"name": "loss"}, None]` | `TypeError: argument of type 'NoneType' is not iterable` |
| `[{"name": "loss"}, {"type": "scatter"}]` | merged as `['a_loss', None, 'b_loss']` |

The second is the more interesting one: the guard passed on `base_data[0]`, the rename loop renamed the first trace, hit the unnamed second, and broke — leaving the pane merged with a nameless trace in the output.

Validating the whole base list up front fixes both and lets the rename loop drop its own mid-loop check, since it can no longer stop half way. The base side is now all-or-nothing, matching the contributor side.

**Verification** — `compare_envs.py` → **58 passed**, full `-m unit` → **116 passed** (excluding the three logger modules whose optional ML deps aren't installed locally). Mutation-tested: restoring the `base_data[0]`-only guard fails exactly the two new tests.

---

@Manik-Khajuria-5 — a note on scope, since this has grown well past what you first reviewed.

It started as a 2-file, +40/-1 fix for the contributor-side guard. It is now six commits, because each review round surfaced another instance of the same underlying pattern in `compare_envs`: trace data being trusted before its shape is checked, and validation interleaved with mutation so a malformed env could half-merge or wipe out an existing comparison. The final state applies one rule consistently — validate a pane's traces completely, then merge all of them or none — on both the base and contributor sides.

Two of those defects predate this PR (the `None`-trace crashes on both sides). I folded them in because they are the same failure mode, but I am happy to split them into a separate PR if you would rather keep this one to the original scope. Just say which you prefer.

Also worth flagging: CI has been sitting at `action_required` since the recent pushes, so Python Tests, CodeQL and Playwright have not run against the current head. They were green on an earlier commit. Everything above is from local runs against Python 3.12.

Manik-Khajuria-5 (changes requested):
@thisisanubhav Could we also check that content is a dict before calling get function?

thisisanubhav:
@Manik-Khajuria-5 good catch — done in ca1a487. All three shapes raised before:

```
content = [{"name": "x"}]  ->  AttributeError: 'list' object has no attribute 'get'
content = None             ->  AttributeError: 'NoneType' object has no attribute 'get'
content = "loss"           ->  AttributeError: 'str' object has no attribute 'get'
```

Worth noting list-shaped content is not hypothetical here: the image branch sets `destWidJson["content"] = [first_img]` when it starts a comparison, so the data model already contains panes whose content is a list.

Guarded in three places rather than one, since `.get` is called on content in three:

- the contributing pane, which covers both plot and image contributors
- the base plot pane, before `.get("data")`
- the base image pane at the point it is deep-copied and read for a caption

The base image pane is deliberately still allowed to become a list after that, because that is exactly what the image merge does — a blanket dict requirement on the base pane would have broken image comparisons.

**Verification** — `compare_envs.py` → **65 passed**, full `-m unit` → **123 passed** (excluding the three logger modules whose optional ML deps aren't installed locally). Mutation-tested: removing the three guards fails exactly the seven new tests and nothing else.

The new tests are parametrised over `list`, `None` and `str` content for both the base and contributor sides, plus one for the image branch specifically.

Jayantparashar10 (commented):
@thisisanubhav Please resolve conflicts

Manik-Khajuria-5 (changes requested):
@thisisanubhav I think the content check should happen before marking the pane as initialized

thisisanubhav:
Both points handled.

**@Jayantparashar10 — conflicts resolved** (merge commit c5a9ef7).

The conflict was a small one: `dev` added the `warmed` parameter and docstring to `compare_envs`, while this branch had added the `_is_named_trace` helper immediately above it. Both kept, nothing dropped. I re-ran the suite after resolving to confirm the merge didn't quietly undo any of the guards on this branch — all of them are still in place.

**@Manik-Khajuria-5 — reordered in e9e842c.**

You're right that the order was wrong. The dict check sat after `seen_dest_wids.add(destWid)` and the `has_compare`/`contentID` assignment, so a pane we then bailed on had already been recorded as initialised — and later envs consult `seen_dest_wids` to decide whether the base is ready to be appended to. The check now runs first, so we never mark a base we haven't actually converted to a list.

One thing worth being straight about: **I could not construct a case where the old ordering produced a wrong result.** I tried the malformed base image with two and three envs, and in both the pane is dropped correctly and nothing raises. I also mutation-tested it — moving the check back after the marker leaves all 66 tests passing. So this closes a latent inconsistency rather than a reproducible bug, and the test I added pins the invariant (no comparison produced, and the source env's pane left unmutated) rather than catching a regression. I'd rather say that than imply I found a failure I didn't.

**Verification** — `compare_envs.py` → **66 passed**; full `-m unit` on the merged tree → **1421 passed, 3 skipped** (skips are optional deps absent locally: xgboost, bs4/lxml).

---

On the scope question I raised earlier and haven't had an answer to: this branch is now nine commits. Two of the defects it fixes predate the PR (the `None`-trace crashes on both base and contributor sides). I'm happy to split those into a separate PR if that makes review easier — just say the word and I'll do it rather than keep growing this one.

Manik-Khajuria-5 (changes requested):
@thisisanubhav I was going through ur code and found one edge case where comparing environments can still fail if the stored plot data doesn't contain a layout. Could you please handle this case as well?

thisisanubhav:
@Manik-Khajuria-5 thanks for going through it — fixed in aa78371.

Confirmed first. `showlegend` was set through `destWidJson["content"]["layout"]` with no check that a layout existed, and every non-dict case fails the whole comparison:

```
layout missing      → KeyError: 'layout'
layout = None       → TypeError: 'NoneType' object does not support item assignment
layout = []         → TypeError: list indices must be integers or slices, not str
layout = "not a dict" → TypeError: 'str' object does not support item assignment
```

**One design choice to flag, since it differs from how the rest of this PR treats malformed data.** For traces I *skip* the pane, because a malformed trace list means there's nothing sensible to merge. For layout I went the other way and give the merged pane a fresh layout instead. Plotly treats `layout` as optional — a figure carrying only `data` is perfectly valid — so a stored plot without one is legitimate data rather than malformed, and dropping it from the comparison would be the wrong outcome. The comparison only needs `showlegend`.

The merged pane is a deep copy, so the stored env is left untouched; the new test asserts that explicitly.

If you'd rather skip these panes than repair them, tell me and I'll flip it — it's a one-line change.

I also checked the other two `layout` accesses in `compare_envs`, in the `show_all` block. Both are already guarded with `isinstance`, so this was the only exposed one.

**Verification** — `compare_envs.py` → **70 passed**, full `-m unit` → **1425 passed, 3 skipped** (skips are optional deps absent locally). Mutation-tested: removing the guard fails exactly the four new parametrised cases. No conflicts with current `dev`.

PR #1797, opened 2026-09-03
fix: compare wide integers instead of 500ing /experiments/search

adity1raut:
## Description

Fix numeric query filtering for very large integers.

Currently, `query.py` converts both sides of numeric comparisons to `float()` before comparing them. This causes two problems:

* Very large integers can raise `OverflowError`, which escapes the request handler and results in a `500`.
* Large integers can lose precision when converted to floats, causing incorrect comparisons.

For example, a query like:

```text
lr > <400-digit integer>
```

can return a `500` even when the value is only present in the query and nothing is stored.

The same issue occurs when a stored parameter contains a very large integer.

The fix removes the unnecessary `float()` conversion from `_to_number`, `_equals`, and `_order`. Python already compares `int` and `float` values correctly, so there is no need to normalize them through `float()`.

This also preserves the exact value of large integers instead of rounding them to floating-point precision.

## Why This Fix

The sorting path already handles this case safely. `store._is_number` catches `OverflowError` for integers that are too large to convert to floats and falls back to text ordering.

Filtering should behave consistently with sorting instead of raising an exception for the same data.

## Testing

Added unit and integration tests covering:

* Very large integer literals with `<`, `>`, `<=`, `>=`, `=`, and `!=`
* Very large stored integer values
* Large integers on both sides of a comparison
* Large integer comparisons without float-precision rounding
* Normal numeric comparisons
* String and `contains` operators
* Large float literals such as `1e999`

Integration tests also verify that `/experiments/search` returns `200` and filters large values correctly.

### Verification

```text
2040 passed, 4 skipped
```

`black py` reports no changes.

## Types of Changes

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Refactor / cleanup

## Additional Notes

No API, request/response, or authentication changes are required.

The only intentional behavior change is that very large integers are now compared exactly instead of being rounded through floating-point conversion.

### DIFF
--- a/py/tests/integration/experiment_search.py
+++ b/py/tests/integration/experiment_search.py
@@ -25,7 +25,7 @@ import tornado.testing
 
 from visdom import Visdom
 from visdom.data_model import JSONStore
-from visdom.experiments import ExperimentStore, QueryParseError
+from visdom.experiments import ExperimentStore, MAX_QUERY_LENGTH, QueryParseError
 from visdom.server.app import Application
 from visdom.server.handlers.web_handlers import ExperimentSearchHandler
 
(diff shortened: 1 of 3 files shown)


adity1raut:
@mariobehling PTAL

Manik-Khajuria-5 (approved):

adity1raut:
@vedansh-5 PTAL

Jayantparashar10 (approved):

rajnisht7 (approved):

PR #1801, opened 2026-09-05
fix(polling): catch event handler exceptions in setup_polling

### LINKED ISSUE #1636: setup_polling doesn't catch handler exceptions, unlike setup_socket it kills the listener thread silently
Visdom.setup_socket() wraps event handler execution in a try/except:

```python
for handler in list(self.event_handlers.get(message["target"], [])):
    try:
        handler(message)
    except Exception as e:
        logger.warn("Visdom failed to handle a handler for {}: {}".format(message, e))
        import traceback
        traceback.print_exc()
``` 

Visdom.setup_polling() has the identical loop, but with no exception handling at all:

```python
for handler in list(self.event_handlers.get(message["target"], [])):
    handler(message)
``` 

### Reproducing code
```python
import visdom

vis = visdom.Visdom(use_incoming_socket=False)  # forces polling mode

def buggy_handler(event):
    1 / 0  # simulate a bug in user code

vis.register_event_handler(buggy_handler, "some_window")
# Trigger an event from the browser targeting "some_window" the
# background polling thread dies silently, no further events are ever
# handled again for this session.
``` 
Expected Behaviour:
Something like `The handler exception should be caught, logged, and the listener threadshould keep running for future events — matching setup_socket's existing behavior.`

Actual behaviour:
`The exception propagates uncaught, killing the Visdom-Socket-Thread. No further events are processed for the rest of the session, with no clear indication of what happened.`

There's even a "# TODO merge with setup_socket?" comment right above setup_polling so this divergence looks that it was still a task todo

If a user's registered event handler throws an exception while using polling mode, the exception isn't caught anywhere. I believe It kills the background listener thread permanently. All future incoming events (not just from the one broken handler) silently stop working for the rest of the session, with no error message pointing to the actual cause.

### Client & Server Logs
these are not applicable here since its a bug in python client's own event-handling code (visdom/__init__.py)

### Additional Context
This only affects sessions using polling mode (use_incoming_socket=False), which is itself a fallback used when a real WebSocket connection isn't available or desired e.g. behind certain proxies or restrictive network
setups that don't support persistent WebSocket connections. So the impact is concentrated exactly on users who are already in a more constrained network environment, which may make it harder for them to notice/debug
since they might reasonably assume the silence is network-related, not a client-side bug.

sundeep8967:
## Description
Wraps event handler invocation in `setup_polling()` within a `try/except` block matching the existing behavior in `setup_socket()`.

## Motivation and Context
Fixes #1636

In polling mode (`use_polling=True` or `use_incoming_socket=False`), if a registered event handler raises an exception, the exception previously propagated uncaught and permanently terminated the background `Visdom-Socket-Thread`. This silently prevented any subsequent events from being processed.

With this change, exceptions from user event handlers in polling mode are caught, logged as warnings with tracebacks, and allow the listener thread to continue running for subsequent events.

## How Has This Been Tested?
Tested event handling in polling mode with a failing callback to confirm exceptions are caught and subsequent messages continue to be handled properly.

## Types of changes

## Checklist:

## Summary by Sourcery

Bug Fixes:
- Prevent exceptions from polling-mode event handlers from terminating the background listener, allowing subsequent events to continue being processed.

### DIFF
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -601,14 +601,10 @@ class Visdom(object):
                 for handler in list(self.event_handlers.get(message["target"], [])):
                     try:
                         handler(message)
-                    except Exception as e:
-                        logger.warning(
-                            "Visdom failed to handle a handler for {}: {}"
-                            "".format(message, e)
+                    except Exception:
+                        logger.exception(
+                            "Visdom failed to handle event for %s", message
                         )
-                        import traceback
-
-                        traceback.print_exc()
 
         def on_close(ws):
             self.socket_alive = False
--- a/py/visdom/__init__.py
+++ b/py/visdom/__init__.py
@@ -599,7 +599,16 @@ class Visdom(object):
                         )
             if "target" in message:
                 for handler in list(self.event_handlers.get(message["target"], [])):
-                    handler(message)
+                    try:
+                        handler(message)
+                    except Exception as e:
+                        logger.warning(
+                            "Visdom failed to handle a handler for {}: {}"
+                            "".format(message, e)
+                        )
+                        import traceback
+
+                        traceback.print_exc()
 
         def on_close(ws):
             self.socket_alive = False


sundeep8967:
Updated the exception logging to use `logger.exception("Visdom failed to handle event for %s", message)`. This avoids any issues with string-formatting custom exception instances and properly logs the traceback using standard library logging.

Manik-Khajuria-5 (changes requested):
@sundeep8967 the issue is already handled. I’d suggest closing this as already fixed.

PR #1802, opened 2026-09-06
fix: bump the pane version for every type, not just plots

adity1raut:
## Description

The incremental update protocol is gated on a version number. The server broadcasts `window_update` carrying the pane's new version, and `main.js` applies the patch only when that version is exactly one ahead of the pane it already holds:

```javascript
cmd.version == storeData.panes[cmd.win].version + 1
```

Anything else falls through to `sendEnvQuery()` and reloads the whole environment.

The version bump lived in `update_window()`, which only plot panes reach. `UpdateHandler.update()` dispatches on `p["type"]` and returns early for `text`, `image_history`, `plot_history`, and `table`, all before `update_window()` is called. Embeddings never go through `update()` at all, since `update_embeddings_packet()` builds their packet by hand and did not touch the version either.

As a result, five of the six updatable pane types remained at version `1` for their entire lifetime while the server continued broadcasting updates for them.

With both sides pinned at `1`, the frontend's check effectively becomes:

```text
1 == 2
```

which can never be true. Every update therefore took the full-environment reload branch.

The incremental update protocol was effectively disabled for the most commonly used pane types in the library, particularly `text` and `image_history`.

After four updates, the server-side versions were:

```text
text          version = 1
image_history version = 1
plot_history  version = 1
table         version = 1
embeddings    version = 1
scatter       version = 5   <- the only type that worked
```

A broadcast captured from a subscribed socket showed:

```text
command=window_update  version=1
command=window_update  version=1
command=window_update  version=1
```

This change moves the version increment out of `update_window()` and into `update_packet()`, which runs once per accepted update regardless of pane type.

The increment is performed before the `jsonpatch` diff so that the new version is included in the generated patch. If the version is bumped after creating the diff, the patch would not contain the new version and could still leave the client desynchronized.

`update_embeddings_packet()` now bumps the version and appends the `/version` operation itself, matching the behavior already used by the socket handlers for `update_comment` and `update_plot_layout`.

The version is updated using:

```python
p["version"] = p.get("version", 1) + 1
```

rather than:

```python
p["version"] += 1
```

This allows environments saved before panes carried a version field to be repaired instead of raising a `KeyError`.

`update_window()` has no other production caller, so no other code path loses the version bump. The plot path reaches `update_packet()` on exactly the same updates it did before, so its version sequence remains unchanged.

## Motivation and Context

The incremental update protocol depends on the client receiving a version exactly one greater than the version it currently holds.

Previously, only the plot path incremented the pane version. The other pane types continued broadcasting updates with version `1`, causing the frontend's sequential-version check to fail on every update.

This meant the server was broadcasting incremental updates, but the client could not apply them and instead reloaded the entire environment.

Moving the version bump into the common `update_packet()` path ensures that every accepted pane update advances the version consistently and that the new version is included in the broadcast patch.

The embeddings path is handled separately because `update_embeddings_packet()` constructs its packet manually.

## How Has This Been Tested?

Unit coverage verifies that:

* All six pane types increment their version once per update.
* Versions continue incrementing correctly across multiple updates.
* The new version is included in the generated patch.
* Panes without an existing `version` field are handled correctly.
* An unrecognised embeddings update type remains a no-op.
* No broadcast is produced for an unrecognised embeddings update type.

Integration coverage reads the versions from a real subscriber socket over an HTTP round trip. This verifies the version actually announced to the client, rather than only checking the server-side stored value, since the announced version is what the frontend uses for its incremental-update gate.

The expected announced sequence is:

```text
2, 3, 4
```

with no gaps.

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Code refactor or cleanup (changes to existing code for improved readability or performance)

## Checklist:

* [ ] I adapted the version number under `py/visdom/VERSION` according to [[Semantic Versioning](https://semver.org/)](https://semver.org/)
* [x] My code follows the code style of this project.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.

## Summary by Sourcery

Restore reliable incremental pane updates by advancing versions in the common accepted-update path and synchronizing the embeddings path.

Bug Fixes:
- Ensure every accepted pane update advances and broadcasts a consecutive version, restoring incremental client updates for text, image history, plot history, table, embeddings, and plot panes.
- Prevent rejected or unrecognized updates from changing versions or producing window-update broadcasts.

Enhancements:
- Handle panes from older saved environments that do not yet contain a version field.
- Keep embeddings patches synchronized by including their version changes explicitly.

Tests:
- Add unit and integration coverage for version progression, patch contents, legacy panes, rejected updates, and versions observed by subscribed clients.

## Summary by CodeRabbit

- **Bug Fixes**
  - Pane version numbers now increment consistently for accepted updates across supported pane types.
  - Browser update notifications now include the latest pane version without gaps.
  - Rejected or unchanged updates no longer trigger unnecessary notifications or version changes.
  - Embeddings updates correctly apply the announced version to the client.
  - Panes saved without a version now receive one automatically.
- **Tests**
  - Expanded coverage for version updates, rejected changes, and browser notifications across supported pane types.

### DIFF
--- a/py/tests/integration/pane_version_broadcast.py
+++ b/py/tests/integration/pane_version_broadcast.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python3
+
+# Copyright 2017-present, The Visdom Authors
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""What a subscribed browser actually receives when a pane is updated.
+
+``js/main.js`` applies an incremental ``window_update`` only when its
+``version`` is exactly one ahead of the pane the client already holds, and
+falls back to re-querying the entire environment otherwise. That makes the
+version sequence a wire contract, not an internal counter, so it is asserted
+here off a real subscriber socket rather than off the server's own state.
+
+The unit file ``py/tests/unit/pane_versioning.py`` pins the same rule at the
+handler; this one exists because the two halves of it -- the version the
+server keeps and the version it announces -- used to be able to disagree, and
+only the announced one is what the frontend gates on.
+"""
+
+import unittest
+
+import pytest
+
+from testutils import open_sub, sent
+from testutils.http import VisdomHTTPTestCase
+from testutils.payloads import content_args
+
+pytestmark = pytest.mark.integration
+
+
+class BroadcastTestCase(VisdomHTTPTestCase):
+    """Attaches a subscriber to ``main`` and reads the versions it is sent."""
+
+    def subscribe(self, eid="main"):
+        sub = open_sub(self._app)
+        sub.eid = eid
+        return sub
+
+    def update_versions(self, sub, win):
+        """Versions of the ``window_update`` packets broadcast for ``win``."""
+        return [
+            msg["version"]
+            for msg in sent(sub)
+            if isinstance(msg, dict)
+            and msg.get("command") == "window_update"
+            and msg.get("win") == win
+        ]
+
+    def assert_updates_are_consecutive(self, sub, win, count=3):
+        """The client starts at 1, so the run has to be 2, 3, ... with no gaps.
+
+        A repeated or stalled version is the failure this guards: it makes the
+        frontend drop the patch and reload the whole environment instead.
+        """
+        self.assertEqual(
+            self.update_versions(sub, win), list(range(2, 2 + count)), self.panes()
+        )
+
+
+class TestContentPaneBroadcasts(BroadcastTestCase):
+    """Panes that carry content rather than traces.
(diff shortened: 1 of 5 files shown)


Manik-Khajuria-5 (changes requested):
@adity1raut There are still version bump issues for rejected updates . Could you please address these?

Manik-Khajuria-5 (approved):

Jayantparashar10 (approved):

PR #1480, opened 2026-06-17
fix: Return HTTP 400 for malformed update requests

Debajeet-1411:
Fixes #1479
## Summary

This PR adds validation for required request fields in update-related handlers to prevent unhandled `KeyError` exceptions caused by malformed requests.

Previously, requests missing required fields such as `win` or `data` could trigger server-side exceptions and return HTTP 500 responses. With these changes, such requests are handled gracefully and return a descriptive HTTP 400 response instead.

Additionally, this PR removes unreachable code from `ErrorHandler.get()`.

### Request validation

Added validation checks for required fields before they are accessed:

* Validate `win` in `ExistsHandler.wrap_func()`
* Validate `win` in `UpdateHandler.wrap_func()`
* Validate `data` in `UpdateHandler.wrap_func()`
* Added a defensive `data` check in `UpdateHandler.update()` to avoid unexpected crashes if the method is called without prior validation

Malformed requests now short-circuit with:

```text
HTTP 400 Bad Request
```

and a descriptive message such as:

```text
missing required field: win
```

or

```text
missing required field: data
```

instead of raising an unhandled `KeyError`.

### Dead Code Removal

Removed unreachable code from `ErrorHandler.get()`.

The method previously contained statements after an unconditional:

```python
raise tornado.web.HTTPError(status_code)
```

which could never be executed.

## Reproduction

A malformed update request can be sent without the required `data` field:

```json
{
  "win": "window_id",
  "eid": "main"
}
```

Prior to this change, the request reached `UpdateHandler.update()` and attempted to access:

```python
args["data"][0]["content"]
```

which resulted in:

```text
HTTP 500 Internal Server Error
KeyError: 'data'
```

After this change, required fields are validated before processing and the request is rejected with:

```text
HTTP 400 Bad Request
missing required field: data
```

A detailed reproduction and traceback are included in the associated issue.

## Impact

This improves server robustness and input validation by ensuring malformed requests receive appropriate client-error responses instead of generating server-side exceptions.

Issue #1479, opened 2026-06-17
Missing request field validation in /update causes unhandled KeyError and HTTP 500


Debajeet-1411:
# Missing request field validation in `/update` causes unhandled `KeyError` and HTTP 500

## Summary

`UpdateHandler.update()` assumes required request fields are always present and directly accesses them using dictionary indexing.
When a malformed but valid JSON request omits the `data` field, the server raises an unhandled `KeyError` and returns HTTP 500 instead of a client-facing HTTP 400 validation error.
This results in an unnecessary server exception for invalid user input.

---

## Environment

* OS: Windows 11
* Python: 3.11
* Visdom: Current development version tested locally

---

### 1. Start Visdom

```bash
python -m visdom.server
```

### 2. Create a text window

```python
from visdom import Visdom

viz = Visdom()

win = viz.text("hello world")
print(win)
```

Example output:

```text
window_3f4c926d7478a0
```

### 3. Send an update request without the required `data` field

```powershell
$body = @{
    win = "window_3f4c926d7478a0"
    eid = "main"
} | ConvertTo-Json

Invoke-WebRequest `
    -Uri "http://localhost:8097/update" `
    -Method POST `
    -ContentType "application/json" `
    -Body $body
```

---

## Actual Behaviour

The server returns:

```text
500 Internal Server Error
```

and raises:

```python
KeyError: 'data'
```

Traceback:

```python
File "...web_handlers.py", line 122, in update
p["content"] += "<br>" + args["data"][0]["content"]
                         ~~~~^^^^^^^^
KeyError: 'data'
```

---

## Expected Behaviour

The request should be rejected with:

```text
400 Bad Request
```

Example response:

```json
{
  "error": "missing required field: data"
}
```

The server should validate required fields before attempting to access them.

---

## Root Cause

The implementation directly accesses request fields using dictionary indexing without checking whether the keys exist first. This pattern appears in three places:

- `UpdateHandler.update()` — `args["data"]` accessed directly on lines 122, 131, 144, and 158 for `text`, `embeddings`, `image_history`, and `plot_history` window types
- `UpdateHandler.wrap_func()` — `args["win"]` accessed directly on line 331 before reaching `update()`
- `ExistsHandler.wrap_func()` — `args["win"]` accessed directly on line 85, same pattern

In all three cases, a malformed request missing the expected key triggers an unhandled `KeyError` instead of a validation error.

The issue was reproduced locally and resulted in:

```text
HTTP 500 Internal Server Error
[...]

PR #1803, opened 2026-09-06
fix: answer opts-only /update requests instead of raising KeyError

tarone-saloni:
## Description

`update_window_opts()` returns a 500 on text, image_history, plot_history and
embeddings panes. Plot panes are the only ones it works on.

The payload that method sends is `{win, eid, layout, opts}`, with no `data` keyat all. `wrap_func` is happy with that, it only rejects a request when `data`,`layout` and `opts` are all missing (the check from #1480). The trouble is
`update()`, which switches on the pane type before it reaches the opts/layout merge, and the text, image_history and plot_history branches read `args["data"][0]` immediately. Embeddings does the same with `args["data"]["update_type"]`. That's a `KeyError`, and `wrap_func` only catches `TypeError` and `ValueError`, so it goes all the way up as a 500:

```
File "py/visdom/server/handlers/web_handlers.py", line 192, in update
  p["content"] += "<br>" + args["data"][0]["content"]
KeyError: 'data'
```

What this changes:

`update()` hands a data-less request to `update_window()` unless it's a plot pane, which is what plot panes were already doing anyway.
`update_window()` needed some work to cope with being called for these types.Text content is a plain string and image/plot history content is a list, so `content["layout"]` and `p["content"]["data"]` can't be read blindly any more.
A pane with no traces in it now just skips the legend rename.

While in there I noticed the pane type check right above has the same problem. It reads `p["content"]["data"]` to work out whether the pane is a plot, so an `/update` aimed at an image, properties or network pane blows up with a
`KeyError` before it can even write out the "win is not scatter..." rejection. That's fixed too, it now reports the type it turned down ("was image") instead of crashing. Plot panes still say "was bar" like before.

Two small ones that fell out of the same code: an empty `data` list no longerraises `IndexError` on a text pane, and a named update with no data comes back as a 400 instead of a `TypeError` out of `len(None)`.

## Motivation and Context

`update_window_opts()` is the only way to change a window's title or size after creating it, and it didn't work on four of the six pane types `/update` accepts. Two of those, image_history and plot_history, are plot windows, which is what the method's docstring says it's for in the first place.

Follow-up to #1479. #1480 fixed the case where the request carries no useful fields at all, but the `layout`/`opts`-only path still lands on the same `KeyError`. It's easier to hit than the original report as well, that one needed a hand-written HTTP request while this is one line of ordinary API use:

```python
viz = visdom.Visdom()
win = viz.text("hello world")
viz.update_window_opts(win, {"title": "New title"})   # 500 on dev
```

## How Has This Been Tested?

Ran two servers side by side, one on `dev` and one on this branch, and sent them the same request:

```
dev      viz.update_window_opts(win, {'title': 'after'})  ->  HTTP 500
PR       viz.update_window_opts(win, {'title': 'after'})  ->  HTTP 200  my_text_pane

--- dev server log ---
KeyError: 'data'
ERROR:tornado.access:500 POST /update (::1) 2.91ms
```

On this branch the saved pane afterwards reads `title = after` and
`version = 2`, with `content = hello world` left alone.

New tests live in `py/tests/unit/update_opts_only.py`. They push the exact payload `update_window_opts` sends through `wrap_func` for all six pane types that accept `/update`, and cover the rejection path and the two edge cases as well. 18 of the 24 fail on `dev`, all 24 pass here.

Rest of the suite on Python 3.12 / macOS, nothing else moved: `pytest -m unit` 1378 passed, `pytest py/tests/integration` 584 passed, `black` reports nochanges. I skipped `keras_logger`, `sklearn_logger` and `test_xgboost_logger` locally because xgboost can't load libomp on this machine, unrelated to any of this.

## Types of changes

## Checklist:

## Summary by Sourcery

Fix opts-only window updates so pane options and layouts are applied safely without requiring trace data.

Bug Fixes:
- Fix `/update` requests containing only layout or option changes so they succeed across supported pane types instead of raising `KeyError` and returning HTTP 500.
- Return clear client errors for invalid named updates and safely handle empty data updates.
- Prevent unsupported pane types and panes without traces from crashing during update validation.

Enhancements:
- Preserve non-plot pane content while applying window options and layouts, and support missing or invalid plot layouts.
- Apply legend updates only when trace data is present.

Tests:
- Add unit coverage for opts-only updates across pane types, rejected updates, content preservation, layout handling, legend behavior, and edge cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed options- and layout-only updates without replacing existing window content.
  * Preserved non-plot content during options-only updates.
  * Improved layout and legend handling across supported chart types.
  * Prevented malformed plot content, layouts, and trace data from causing errors.
  * Rejected invalid named trace updates before changing window state.
  * Corrected update handling for embeddings and unsupported window types.
  * Ensured delete requests with empty data still clear plot traces.

### DIFF
--- a/py/tests/unit/update_opts_only.py
+++ b/py/tests/unit/update_opts_only.py
@@ -1,0 +1,184 @@
+#!/usr/bin/env python3
+
+# Copyright 2017-present, The Visdom Authors
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""``/update`` requests that carry ``opts``/``layout`` but no ``data``.
+
+This is the shape ``Visdom.update_window_opts()`` sends. Only plot panes ever
+handled it; text, image_history, plot_history, table and embeddings panes
+indexed ``args["data"]`` unconditionally and answered HTTP 500.
+"""
+
+import copy
+
+import pytest
+import tornado.web
+
+from visdom.server.handlers.web_handlers import UpdateHandler
+from visdom.utils.shared_utils import get_rand_id
+
+from testutils.fakes import FakeHandler
+
+pytestmark = pytest.mark.unit
+
+
+def _pane(ptype, **extra):
+    pane = {
+        "command": "window",
+        "version": 1,
+        "id": "win_{}".format(ptype),
+        "title": ptype,
+        "contentID": get_rand_id(),
+        "type": ptype,
+    }
+    pane.update(extra)
+    return pane
+
+
+def _plot_pane():
+    return _pane(
+        "plot",
+        content={
+            "data": [{"type": "scatter", "x": [1], "y": [1], "name": "a"}],
+            "layout": {},
+        },
+    )
+
+
+def _text_pane():
+    return _pane("text", content="line0")
+
+
+def _image_history_pane():
+    return _pane("image_history", content=[{"src": "img0"}], selected=0)
+
+
+def _plot_history_pane():
+    return _pane("plot_history", content=[{"data": [], "layout": {}}], selected=0)
+
+
+def _table_pane():
+    return _pane("table", content=[["a", "b"]], editable=True)
+
+
+def _embeddings_pane():
+    return _pane(
+        "embeddings",
+        content={"data": [[1, 2]], "selected": None, "has_previous": False},
+        old_content=[],
+    )
+
+
+def _image_pane():
+    return _pane("image", content={"src": "img0", "caption": None})
+
+
+PANE_CASES = [
+    ("plot", _plot_pane),
+    ("text", _text_pane),
+    ("image_history", _image_history_pane),
+    ("plot_history", _plot_history_pane),
+    ("table", _table_pane),
+    ("embeddings", _embeddings_pane),
+]
+
+CONTENT_PRESERVING_CASES = [case for case in PANE_CASES if case[0] != "plot"]
+
+OPTS = {"title": "renamed", "width": 400}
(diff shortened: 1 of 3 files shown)


Manik-Khajuria-5 (changes requested):
@tarone-saloni Could you handle missing layout safely and support other valid pane types like bar, boxplot, and pie?

tarone-saloni:
> @tarone-saloni Could you handle missing layout safely and support other valid pane types like bar, boxplot, and pie?

I’d prefer to keep this PR focused on opts-only updates. I’ll fix the missing layout handling and add support for bar, boxplot, pie, and histogram for opts-only updates. We can handle their data-update logic separately in a follow-up PR.

Manik-Khajuria-5 (commented):
@tarone-saloni Please resolve soccery comments

Jayantparashar10 (approved):

Manik-Khajuria-5 (changes requested):
@tarone-saloni please resolve code rabbit comments

tarone-saloni:
> @tarone-saloni please resolve code rabbit comments

@Manik-Khajuria-5 Both CodeRabbit comments are addressed in e1aed0b.

- Named updates now require `data` to be a list with exactly one entry, so `{"name": "a", "data": 1}` gets a 400 instead of a 500.
- `layout` that isn't an object, `opts` that isn't an object, and `opts.legend` that isn't a list are all rejected with a 400 before the pane is touched, so a bad request no longer leaves a half-applied update behind.

I put the checks at the top of `update()` rather than inside `update_window()`, since `update()` is the only caller and the 400 belongs with the other request validation there. Added tests for each case, and each one checks the pane is unchanged after the rejection.

PR #1818, opened 2026-09-10
fix: keep numbers ahead of other types when ranking experiments descending

piyush182004:
## Description

ISSUE:- #1819

`_rank_key` returned `_order_key(value)` as the leading element of the sort key and `_rank` applied `reverse=descending` to the whole key. `_order_key`'s first element is a type tag — `0` for numbers, `1` for everything else — so reversing the comparison reversed the grouping too, ranking a run whose value is text above the run holding the highest number.

The tag is now negated for a descending sort, the same technique the function already used for the `seq` tie-breaker. Fixed in `_rank_key` rather than `_rank` because `_rank` has two paths — `sorted(reverse=)` and `heapq.nlargest` — and both were affected.

## Motivation and Context

A field can be numeric on some runs and text on others — a crashed run logged as `"failed"`, a pending one as `"n/a"`. "Show my best runs" then puts a run with no score at the top:

```python
store.log_experiment("baseline", params={"accuracy": 0.72})
store.log_experiment("tuned",    params={"accuracy": 0.91})
store.log_experiment("crashed",  params={"accuracy": "failed"})

store.search(sort_by="accuracy", descending=True)
# actual:   ['crashed', 'tuned', 'baseline']
# expected: ['tuned', 'baseline', 'crashed']
```

With paging it is worse: the heap selects on the same key and evicts real numeric results.

```python
store.search_page(sort_by="accuracy", descending=True, limit=2)
# → ['crashed', 'tuned']   — 'baseline' dropped for a run with no score
```

Ascending is correct; only descending is affected, and nothing warns.

## How Has This Been Tested?

- 28 ranking scenarios: numbers only, mixed number/text, several strings, ties in both directions, booleans, empty and single-entry lists, negatives and floats, `keep` larger than the list, and end-to-end through `search()`
- Confirmed both `_rank` paths were affected before the fix and both are correct after
- Tie stability preserved in both directions
- Two regression tests added to `py/tests/unit/experiment_store.py`, verified to fail against the pre-fix code (`'run_str' != 'run_ten'`) and pass with it
- `pytest -m "not server"` and `black --check` clean. The remaining failures on my machine are pre-existing Windows `MAX_PATH` ones in `data_model.py` and friends, which also fail on a clean `dev` checkout here and pass on the Linux runners.

## Types of changes

## Checklist:

## Summary by Sourcery

Preserve numeric-first ranking semantics for descending experiment searches and pagination.

Bug Fixes:
- Keep numeric experiment values ranked ahead of non-numeric values when sorting in descending order, including bounded search pages.

Tests:
- Add regression coverage for mixed-type rankings, bounded pages, and stable ordering of equal values in both sort directions.

## Bug Fixes

- Corrected experiment sorting by accuracy so numeric values consistently rank ahead of string values in ascending and descending results.
- Fixed paginated experiment results so non-numeric values do not displace higher-ranked numeric results.
- Preserved scan order when experiments have equal accuracy values, ensuring stable results in both sorting directions.

### DIFF
--- a/py/tests/unit/experiment_store.py
+++ b/py/tests/unit/experiment_store.py
@@ -499,6 +499,26 @@ def test_a_string_never_outranks_the_highest_number(experiments):
     assert _ranked(experiments, descending=False) == ["run_five", "run_ten", "run_str"]
 
 
+def test_a_bounded_page_is_not_filled_by_a_string(experiments):
+    """The heap path ranks like the sort path, so a page keeps the numbers.
+
+    ``search_page`` selects through ``heapq`` rather than sorting, on the same
+    key. Ranking a string first there does not merely misorder the page: it
+    takes a slot, evicting a run that belongs on it.
+    """
+    experiments.log_experiment("run_five", params={"acc": 5})
+    experiments.log_experiment("run_ten", params={"acc": 10})
+    experiments.log_experiment("run_str", params={"acc": "pending"})
+
+    page, total = experiments.search_page(sort_by="acc", descending=True, limit=2)
+    assert [experiment.env_id for experiment in page] == ["run_ten", "run_five"]
+    assert total == 3
+
+    page, total = experiments.search_page(sort_by="acc", descending=False, limit=2)
+    assert [experiment.env_id for experiment in page] == ["run_five", "run_ten"]
+    assert total == 3
+
+
 def test_equal_values_keep_their_scan_order_in_both_directions(experiments):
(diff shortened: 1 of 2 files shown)


Manik-Khajuria-5 (approved):

Jayantparashar10 (approved):

Jayantparashar10:
@piyush182004 Update the branch as the CI is failing please update the pr branch with the base branch.

PR #1820, opened 2026-09-11
fix: do not cache a failed window id in the keras and pytorch loggers

Jayantparashar10:
## Description
When a send fails, `viz.line()` returns `False` instead of raising, and both loggers stored that as the window id. The id is only ever written on the create path, so the metric stayed broken after the server recovered. Both loggers now register a window only when one actually came back.

## Motivation and Context
A brief connection drop mid-training permanently broke plotting for every metric that happened to be created during it: subsequent calls sent `win=False`, which the server cannot match to a stored pane, so it re-created the window on every append - one junk pane holding only the latest point instead of the per-metric charts.

## How Has This Been Tested?
Six unit tests added across `py/tests/unit/keras_logger.py` and `py/tests/unit/pytorch_logger.py`: each drives `viz.line` to return `False`, asserts the metric is not registered, then lets it succeed and asserts a fresh window is created rather than appending to `win=False`.

## Types of changes

## Checklist:

## Summary by Sourcery

Ensure failed metric sends do not corrupt logger state and are retried safely after Visdom recovers.

Bug Fixes:
- Prevent failed Visdom window creation from being cached as a window ID in the Keras and PyTorch loggers, allowing metrics to recover after connection failures.
- Preserve unsent metric values so failed sends can be retried when the server becomes available.

Tests:
- Add unit coverage for failed window creation, recovery, pending-value retention, and throttling behavior in the Keras and PyTorch loggers.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved metric plotting reliability when visualization requests fail or the server is unavailable.
  - Failed window creation is no longer treated as a valid plotting window.
  - Metrics create and track a fresh window after the visualization server recovers.
  - Failed metric values remain pending and can be retried during later logging or final flushes.
  - Metric throttling remains effective when a window has not yet been created, preventing premature send attempts.


Manik-Khajuria-5 (commented):
@Jayantparashar10 Since failed requests aren't cached in wins, which means log every throttling can be bypassed when the server is unavailable. Can u please check through it once

Jayantparashar10:
> @Jayantparashar10 Since failed requests aren't cached in wins, which means log every throttling can be bypassed when the server is unavailable. Can u please check through it once

@Manik-Khajuria-5 I have applied the fix please take a look.

Jayantparashar10:
@sourcery-ai  review

rajnisht7 (approved):

Manik-Khajuria-5 (approved):
@Jayantparashar10 LGTM!

vedansh-5 (approved):

